PR #88, opened 2026-06-19
feat: extract reusable build utils and replace CmdTasks in the build

totollygeek:
The project's own build had grown a set of free helper functions and a
handful of targets that shelled out through the generic CmdTasks fallback.
Promote the helpers to reusable core utilities and replace the CmdTasks
calls with typed tool wrappers, so other builds can reuse the same pieces
and Zuke keeps dogfooding its own typed APIs.

Core gains a small filesystem helper that removes a path while tolerating a
missing target, manifest helpers that read and validate a deno.json version,
and a kill-after execution timeout on the shell command and the tool
settings base that aborts a process past a deadline and raises a dedicated
timeout error.

The deno wrapper gains install and publish subcommands. The jsr wrapper
gains read-only registry helpers that report which versions of a package are
already published. A new release-please wrapper package wraps the
release-pr and github-release subcommands.

The build now uses these throughout: clean removes the dist directory via
the core helper, coverage runs through the deno coverage and run wrappers,
spell installs cspell on demand and lints through the cspell wrapper,
release installs and drives release-please, and the JSR publish loop reads
versions and registry state through the new helpers while bounding each
publish with the kill-after timeout.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_018M3JEtwN7jaUco8K5Xctth


zuke.ts
```diff
@@ -164,9 +181,15 @@ class ZukeBuild extends Build {
...
     .executes(async () => {
-      const argv = ["run", "--allow-read", "--allow-env", "--allow-sys"];
-      argv.push("npm:cspell@9", "lint", "--no-progress", "**");
-      await CmdTasks.exec(Deno.execPath(), (s) => s.args(...argv));
+      const cspell = await installCli(
+        "npm:cspell@9",
+        "cspell",
+        (s) => s.allow("read").allow("env").allow("sys"),
+      );
+      await CspellTasks.lint((s) =>
```
totollygeek, line 180: That method should be named "check", not lint. We should change it in the package

zuke.ts
```diff
@@ -55,10 +62,51 @@ const PACKAGES = [
...
+}
+
+/**
+ * `PATH` with the running Deno's directory prepended. The launcher that
+ * `deno install` writes for an npm package runs `exec deno …` by name, so an
+ * installed CLI only works when `deno` is resolvable on `PATH`. The ./zuke
+ * bootstrap may invoke Deno by absolute path without exporting it, so make it
+ * discoverable for the installed shims here.
+ */
+function pathWithDeno(): string {
```
totollygeek: That function should not be needed. When we install Deno with the install tasks, we should get that as a return and use it in the build as AbsolutePath object

zuke.ts
```diff
@@ -55,10 +62,51 @@ const PACKAGES = [
...
+  "release-please",
   "security",
 ];
 
-/** The `version` field of a package's `deno.json`, validated as a string. */
+/**
+ * Where build-time CLIs are installed on demand. Gitignored (`/.zuke/`), so the
+ * install is a transient, per-run artifact.
+ */
+const TOOLS_ROOT = ".zuke/tools";
```
totollygeek: That should be AbsolutePath object

zuke.ts
totollygeek: Done — `installCli` now returns the launcher as an `AbsolutePath` (joined onto `TOOLS_ROOT`, itself a `repoRoot(...)` `AbsolutePath`), and `pathWithDeno` is gone.

One note on why the helper existed: the launcher that `deno install` writes for an npm package is a shell script that runs `exec deno run npm:cspell …` — invoking `deno` **by name**. So the install location being an `AbsolutePath` doesn't by itself let the shim find Deno; it still needs `deno` resolvable on `PATH`. The right place to fix that is the `./zuke` bootstrap (and `zuke.ps1`), which is what locates Deno in the first place — both now prepend Deno's directory to `PATH`. That removes the workaround from the build entirely.

---

PR #151, opened 2026-07-13
feat(core): verified, cached toolchain provisioning

totollygeek:
## What & why

First of two PRs for **hermeticity** ("clone and run, nothing preinstalled"). Today every `*Tasks` wrapper assumes its binary is already on `PATH`, and `installRelease` fetches a binary but never verifies it and re-downloads every run. This PR makes a build **fully describe the tools it needs** — pinned, verified, and cached — so `works on my machine` becomes the build's problem.

### 1. Checksummed, cached `installRelease`

`installRelease` gains one optional field, **`checksum`** (the SHA-256 release pages already publish — the `.tar.gz`'s hash for an archive, the binary's for a `"raw"` download). When set it does double duty:

- **Verification** — the download is hashed and compared before anything is installed. A mismatch throws (`checksum mismatch for "…": expected …, got …`) and **nothing is left on disk** (the unverified `raw` binary is removed).
- **Caching** — the verified checksum is recorded next to the binary (`<tool>.sha256`), so a later run whose pin matches is a **cache hit that skips the download entirely**. Bumping the checksum (a new version) is a natural miss that re-fetches.

Fully backward-compatible: with no `checksum`, the tool is fetched every run and unverified, exactly as before.

```ts
const bin = await installRelease({
  name: "helm",
  destDir: ".zuke/bin",
  archive: "tar.gz",
  binaryPath: `${Deno.build.os}-amd64/helm`,
  checksum: "f43e1c3…",           // verify + cache
  url: ({ arch }) => `https://get.helm.sh/helm-v3.15.2-linux-${arch === "aarch64" ? "arm64" : "amd64"}.tar.gz`,
});
```

### 2. `toolchain()` — declare a build's whole toolset in one place

When a build needs several tools, `toolchain()` bundles them so the build file describes its environment instead of a README's "install these first". Add tools with `.tool(spec)` (each an `installRelease` spec), then `install()` fetches them **concurrently** — pinned, verified, cached — and returns a `Map<name, AbsolutePath>` for each wrapper's `.toolPath(...)`. Tools default to `.zuke/tools`, overridable per call (`destDir`) or per tool.

```ts
class Deploy extends Build {
  tools = toolchain((t) =>
    t.tool({ name: "helm", archive: "tar.gz", checksum: "…", url: helmUrl })
     .tool({ name: "kubectl", checksum: "…", url: kubectlUrl })
  );

  deploy = target().executes(async () => {
    const bin = await this.tools.install();
    await HelmTasks.version((s) => s.toolPath(bin.get("helm")));
  });
}
```

There's **no built-in tool registry** (you supply the URL + checksum), which keeps this dependency-free and avoids an aqua/mise-style maintenance surface — while still giving the declarative, pinned, verified experience.

### New public API

- `installRelease` — new optional `checksum` on `InstallReleaseOptions` (no behavior change when omitted).
- `toolchain()`, `Toolchain`, `type ToolSpec`, `type ToolchainInstallOptions`, `DEFAULT_TOOLS_DIR`.

### Testing & coverage

- New `toolchain_test.ts`; extended `install_test.ts` (checksum verify pass/mismatch for raw and `tar.gz`, cache-hit skips the download, a changed pin re-downloads). All seams (download, cwd) are injectable/temp-dir'd, so the suite stays hermetic.
- Full `deno task ci` green — **1154 tests**, aggregate coverage **99.0% lines / 96.6% branches** (above the 95% gate). `toolchain.ts` is 100%.

### Docs

- `docs/authoring.md` — the **Tool install** section documents `checksum` (verify + cache) and gains a **Pinned toolchains — `toolchain()`** subsection.
- Regenerated `llms.txt` / `llms-full.txt` / `packages/core/README.md` with `./zuke apiDocs`.

## Related issues

Part of the "hermeticity" brainstorm section. A follow-up PR adds **secret sources + guaranteed redaction** (the other half of that section).

## Checklist

- [x] Docs updated in the same PR (`README.md`, JSDoc, `docs/`).
- [x] Public API changes were regenerated with `./zuke apiDocs` (`llms.txt`, `llms-full.txt`, package README `## API`).
- [x] No `any`, no `as` casts or `!` non-null assertions in `src/` (narrowed with type guards instead).
- [ ] A new package was wired into all five places — n/a (no new package; all changes are within `@zuke/core`).

https://claude.ai/code/session_01PFVtbsGNx9piPuknJXqgrH

---
_Generated by [Claude Code](https://claude.ai/code/session_01PFVtbsGNx9piPuknJXqgrH)_


totollygeek:
Addressed the security finding (`o3t5vk` — checksums normalized but not validated) in `4769222`.

**Assessment:** not a real vulnerability. Verification is `sha256(bytes) === checksum`, and `sha256()` always yields a 64-char lowercase hex string, so a malformed checksum simply **can't match** → the download is rejected. It fails *closed* (rejects), never *open* (accepts a tampered binary).

**Still worth a small fix, though**, for robustness — and one case is a footgun in the per-platform resolver added later in this PR: a resolver like `({os,arch}) => sums[…]` returns `undefined` for an unmapped platform, so the old code threw a bare `TypeError`, and a typo'd/truncated hash gave a confusing `mismatch: expected <garbage>`. `resolveChecksum` now validates the value is a 64-char hex SHA-256 **before any download**, throwing an actionable error that names the platform (`invalid checksum for linux/x86_64: expected a 64-character hex SHA-256, got …`).

Covered by two new tests (malformed literal rejected before the download runs; resolver-with-no-entry gives the clear error). Full `deno task ci` green.

---

PR #164, opened 2026-07-17
feat(core): pass a typed TargetContext to target bodies

totollygeek:
## What

M0 of the orchestration plan (`plans/orchestrationplan.md`): the **execution-context seam** every cross-run feature (durable state, waits, locks, cancellation, fan-out) will build on.

- `TargetFn` now receives a `TargetContext` — the run id, the target name, a cancellation `AbortSignal`, and the dry-run flag. **Backwards compatible:** an existing zero-argument `executes` body still type-checks and runs unchanged (a zero-arg function is assignable to the one-parameter type).
- The executor mints one run id and one `AbortController` per run. A new `ExecuteOptions.signal` cancels the run; the run signal is mirrored on every target's context and installed as the shell's ambient default, so an in-flight `$` command is terminated (SIGTERM) on cancellation.
- `Command` gains a `.signal()` method plus an internal ambient-signal seam (`src/ambient_signal.ts`, not a published entrypoint) so a plain `$` in a body is cancelled without threading the signal by hand.

## Deferred (by design, to their milestones)

- Durable `state` and external-signal `signals` on the context → M1 / M3 (their backing types don't exist yet).
- OS `SIGINT`/`SIGTERM` wiring and per-target-timeout child abort → M6 (owns cancellation UX; avoids global signal-handler side effects and a retry-vs-signal race here).
- Printing the run id in the run header / job summary → M1 (would change the `Renderer` interface `@zuke/console` implements, and shifts ~20 output-pinned test assertions).

## Acceptance (met)

- Existing suite passes untouched.
- A body observes a `ctx` with a stable `runId` across targets, its own name, a live signal, and `dryRun=false`.
- Cancelling a run terminates its in-flight child process (the kill test settles in ~54ms against a 30s sleep).

## Verification

`deno task ci` green — fmt, lint, spell, `deno check`, and coverage (lines 98.9%, branches 96.5%, both above the 95% gate). `deno doc --lint` clean over all core entrypoints; `apiDocsCheck` shows no drift (`llms-full.txt` + core `README.md` regenerated).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@
 
 # Specs
 docs/superpowers/
+plans/
 
 # OMC session state
 .omc/
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -1464,6 +1464,12 @@ interface ExecuteOptions
     Renderer for the per-target banners and the end-of-build summary. Defaults
     to Zuke's built-in {@link defaultRenderer}; `@zuke/console` exports an
     alternative a build can inject to restyle its output.
+  signal?: AbortSignal
+    Cancel the run when this signal aborts. Every target body's
+    {@link "./target.ts".TargetContext} `signal` mirrors it, and it is applied
+    as the shell's ambient default so an in-flight `$` command is terminated
+    (SIGTERM) on cancellation. A body that ignores its signal still runs to completion —
+    graph-level cancellation/compensation is a later milestone.
 
 interface FanOutOptions
   Options for {@link fanOutPipeline}: how a build's targets become parallel CI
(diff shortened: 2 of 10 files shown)


totollygeek:
Thanks — addressed all three findings from the security + generic reviews (all `low`):

- **Global mutable ambient signal / concurrent-run race** (`le94ar`, `1eoj1pr`): the ambient signal now lives in an `AsyncLocalStorage` (`src/ambient_signal.ts`) instead of a module-global. The plan runs inside `withAmbientSignal(runController.signal, runPlan)`, so each run's signal is bound to its own async context — concurrent in-process runs are isolated. Verified by a new parallel-cancellation test that kills in-flight commands in every branch (proving the binding also propagates through the scheduled runner's callback pump).
- **Stale ambient signal if setup throws synchronously** (`gh9kct`): folding into `withAmbientSignal` removes the manual set/restore entirely — the `AsyncLocalStorage` binding unwinds on its own when the plan returns *or throws*, so there's no window to leak.

`node:async_hooks` is a Deno built-in (no npm/network, JSR-publishable), and it's the only way to get async-context scoping in Deno. `deno task ci` green (coverage 98.9% lines / 96.5% branches), `apiDocsCheck` no drift, `deno doc --lint` clean.

totollygeek:
On the re-review: the **generic** review is now clean (0 findings), and the one remaining **security** finding (`10be98t`, low) — "run cancellation can now terminate arbitrary in-flight shell commands" — is working-as-designed, not a defect. Leaving the code as-is, with the reasoning on the record:

- **Threat model in M0:** the only way to abort a run is via `ExecuteOptions.signal`, held by whoever calls `execute()` — the trusted orchestrator/CLI process itself. There is no remote or untrusted abort path in this PR, so the finding's precondition ("*if* untrusted code can abort the run") doesn't hold yet.
- **The behavior is the feature:** cancellation terminates the run's *own* child processes (SIGTERM) for graceful shutdown. It can't reach processes outside the run. Guarding against a caller cancelling its own build would defeat the purpose.
- **Where the gate belongs:** the untrusted/remote cancel path (`zuke cancel`, MCP `cancel_run`) arrives in **M5/M6**, and M5 explicitly adds the authz tiers + audit log that decide *who* may cancel. That's the correct layer for this concern, not the in-run plumbing here.

No code change; flagging as a non-issue for M0.

PR #166, opened 2026-07-17
feat(core): durable run records with a pluggable StateStore

totollygeek:
## What

**M1, PR 1 of 2** of the orchestration plan (`plans/orchestrationplan.md`): a **pluggable `StateStore`** that persists a run's status, graph shape, resolved non-secret parameters, and per-target progress — so a run can be reconstructed after the process exits, and a target can leave metadata for a later run to read. **Opt-in and zero-overhead when unused.**

### Store
- `StateStore` interface with whole-record **compare-and-swap** (`putRun(record, expectedVersion)`), `getRun`, and `listRuns({ status?, target?, since? })`.
- **`FileSystemStateStore`** — one `.zuke/runs/<id>.json` per run; atomic write-temp-then-rename under an `O_EXCL` lock; content-hash version. Single host (fine for dev).
- **`HttpStateStore`** — a hosted service using `ETag` / `If-Match` for optimistic concurrency (the production path). Same `{ url, token?, fetch? }` shape as `HttpCacheStore`.
- Selection by precedence: explicit option → `Build.stateStore()` override → `ZUKE_STATE_URL`/`ZUKE_STATE_DIR` → a `.zuke/runs` default **only when the run opts in** (`--state`).

### `ctx.state` (fills in the M0 seam)
- `ctx.state.set(patch)` merges a JSON patch, persisted through a serialized best-effort writer; `ctx.state.get()` reads it back. In-memory no-op when no store is configured.
- **Secrets stay out of state:** secret parameters are excluded from the record, and every `ctx.state` value is run through the run's redactor before persisting.

### Executor + CLI
- The executor records each target's start/finish and the run's terminal status, and drains the writer before returning. State writes are **best-effort** — a store hiccup is reported, never fatal.
- `--state` opts a plain run into the filesystem store; `--actor <name>` attributes it.

### Docs & skills
- New `docs/state.md`, `docs/state-api.md` (the REST contract for hosting a backend), and `docs/run-context.md` (the M0 context + cancellation, whose docs were deferred). `zuke-write-build` skill + cheatsheet updated for both M0 and M1.

## Acceptance (met)
- A build run with the FS store is reconstructed from disk alone by a fresh store (status, per-target status/`startedAt`, `ctx.state` metadata, secret excluded from `params`) — integration test.
- Two concurrent `putRun` writers at the same version → exactly one wins, the loser gets a typed conflict — tested against both the fake host and the real `O_EXCL` lock.

## Verification
`deno task ci` green — fmt, lint, spell, `deno check`, coverage (lines **98.8%**, branches **96.1%**, above the 95% gate). `deno doc --lint` clean over all core entrypoints; `apiDocsCheck` shows no drift.

## Not in this PR
`zuke runs list/show` (reading state from the command line) is **PR 2**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,8 +284,13 @@ Full documentation lives in [`docs/`](./docs/):
   vars (`parameter()`, `this.x.value`).
 - [Authoring API](./docs/authoring.md) — `target()`, `Build`, `run()`,
   code-first CI generation (`cicd()`), and gotchas.
+- [Run context & cancellation](./docs/run-context.md) — the `TargetContext` a
+  body receives (`runId`, `signal`, `state`) and cancelling a run.
 - [Caching](./docs/caching.md) — the incremental build cache
   (`.inputs()`/`.outputs()`) and the AI response cache (`aiCache`).
+- [Durable run state](./docs/state.md) — persist a run's status and per-target
+  metadata to a pluggable store (`StateStore`, `ctx.state`), with an
+  [HTTP API](./docs/state-api.md) for hosting a production backend.
 - [Shell wrapper (`$`)](./docs/shell.md) — ergonomic, injection-safe process
   execution.
 - [Paths (`absolutePath`)](./docs/paths.md) — the fluent path type.
--- a/docs/README.md
+++ b/docs/README.md
@@ -5,6 +5,8 @@
 - [Core concepts](./concepts.md) — the build/target/graph model and execution
   semantics.
 - [Authoring API](./authoring.md) — `target()`, `Build`, `run()`, and gotchas.
+- [Run context & cancellation](./run-context.md) — the `TargetContext` a body
+  receives (`runId`, `signal`, `state`), and cancelling a run.
 - [Parameters](./parameters.md) — typed build inputs from flags, env, or
   defaults.
 - [Secrets](./secrets.md) — source secret values from a manager with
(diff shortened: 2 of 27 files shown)


totollygeek:
Addressed the security finding (`16n7efw`, medium — "path construction depends on every caller validating run IDs"):

`assertSafeId` now runs **inside** the path constructors `#file(id)` and `#lock(id)` — the two places every run-id-derived path is built — instead of only at the `getRun`/`putRun` boundary. A traversal can no longer slip through a caller that forgets to validate; it's defence in depth at the point of use. `listRuns` builds paths from `Deno.readDir` base names (never attacker-controlled), so it was never a vector.

Extended the test to assert both `getRun` and `putRun` reject an unsafe id (`../escape`). `deno task ci` green (coverage lines 98.8% / branches 96.1%), `apiDocsCheck` no drift.

totollygeek:
On the re-review's two new findings (`gle71i`, `pc6532`, both medium) — both flag **operator-controlled configuration** as an attack surface, which is a deliberate, documented posture, not a defect:

- **`HttpStateStore` URL / SSRF (`gle71i`)** and **`FileSystemStateStore` dir / traversal (`pc6532`)**: the `url`, `token`, and `dir` are *trusted configuration* — the endpoint/location **you** choose to store run state (via `ZUKE_STATE_URL` / `ZUKE_STATE_DIR`, `--state`, or `Build.stateStore()`). This is exactly the posture of the existing `HttpCacheStore.url` and `FileSystemCacheStore` dir that this layer mirrors, and it's documented as such. An actor who can set those env vars / build config already controls the build environment; there is no lower-privileged input flowing into them.
- The genuinely **untrusted** value — the run **id** reaching a filesystem path — is validated at every point a path is built (the previous finding's fix).

Made the posture explicit in code: `HttpStateStore` already carried a "trusted configuration" security note; I added the matching note to `FileSystemStateStore`, and `docs/state.md` / `docs/state-api.md` state it too. No runtime validation of operator-chosen config is added, to stay consistent with the remote-cache backends. Docs-only change; `apiDocsCheck` clean, `deno doc --lint` clean.

PR #168, opened 2026-07-17
feat(core): suspendable targets with waitsFor and external signals

totollygeek:
## What

**M3, PR 1 of 2** of the orchestration plan — the **suspend half** of suspend/resume. A target can pause a run until an external event occurs; the run's state is saved so it can be resumed later, in a different process.

```ts
awaitTesting = target()
  .dependsOn(this.deploy)
  .waitsFor((s) =>
    s.on(externalSignal("testing-approved"))
      .timeout("72h")
      .onTimeout(() => this.rollback));
```

### Authoring
- `.waitsFor((s) => …)` — a fluent settings lambda (per the new AGENTS.md guideline 9) — makes a target a **gate** with no body. Triggers: **`externalSignal(name)`** (satisfied when a named signal is delivered; payload exposed via `ctx.signals`) and **`resumeWhen(predicate)`** (async predicate, re-checked on resume). `WaitTrigger` is exported so wrappers can add their own (e.g. M9's "wait for a GitHub workflow").

### Semantics
- **Unsatisfied → suspend:** the target is recorded `waiting` (with its trigger, deadline, and `onTimeout` disposition), the run is recorded `suspended`, targets behind the gate stay `pending`, **independent branches finish**, and the process **exits 0**.
- **Satisfied → pass through:** the gate is `succeeded` and dependents run in the same process.
- A wait **requires a state store** (a build with `.waitsFor()` enables the `.zuke/runs` store by default); declaring one with state disabled fails with a friendly error.

### Record & surfaces
- `RunRecord` gains a `signals` map and a per-target `waitingFor`; parsing defaults missing `signals` to `{}` (backwards compatible with M1/M2 records). `ctx.signals` exposes received payloads to bodies.
- `TargetStatus` and the summary gain a `waiting` state; `BuildResult` gains `suspended` (a suspended run is `ok: true`, exits 0).

### Docs & skills
- New `docs/orchestration.md`; `zuke-write-build` SKILL + cheatsheet updated.

## Acceptance (this PR's subset)
- Unsatisfied wait suspends: reconstructed from disk the run is `suspended`, the gate `waiting` with its `waitingFor` (trigger + deadline + `{target: "rollback"}`), the prior target `succeeded`, the blocked dependent left `pending`, and an independent branch ran — integration test.
- Satisfied wait passes through; no-store and no-trigger both fail with friendly errors.

## Verification
`deno task ci` green — fmt, lint, spell, `deno check`, coverage (lines **98.7%**, branches **96.0%**). `deno doc --lint` clean; `apiDocsCheck` no drift.

## Not in this PR
**PR 2** — `zuke resume <id> --signal … --data …` with exactly-once CAS, run continuation (re-instantiate the build, seed `done` from the record, graph-drift check), `AlreadyResumedError`, and lazy timeout enforcement. The full three-process acceptance scenario lands there. (Also still pending from earlier: **M1 PR2** `zuke runs list/show`.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,6 +293,8 @@ Full documentation lives in [`docs/`](./docs/):
   [HTTP API](./docs/state-api.md) for hosting a production backend.
 - [Cross-run locks](./docs/locks.md) — `.lock()` claims an exclusive resource
   across runs and machines, with a TTL backstop and typed `LockConflictError`s.
+- [Orchestration: waits](./docs/orchestration.md) — `.waitsFor()` suspends a run
+  until an external signal or predicate, saving its state to be resumed later.
 - [Shell wrapper (`$`)](./docs/shell.md) — ergonomic, injection-safe process
   execution.
 - [Paths (`absolutePath`)](./docs/paths.md) — the fluent path type.
--- a/docs/README.md
+++ b/docs/README.md
@@ -19,6 +19,8 @@
   metadata to a pluggable store, and read it back after the process exits.
 - [Cross-run locks](./locks.md) — `.lock()` claims an exclusive resource across
   runs and machines, with a TTL backstop and typed conflicts.
+- [Orchestration: waits & suspend/resume](./orchestration.md) — `.waitsFor()`
+  suspends a run until an external event, to be resumed later.
 - [State HTTP API](./state-api.md) — the REST contract for hosting a production
   state backend.
 - [Shell wrapper (`$`)](./shell.md) — ergonomic, injection-safe process
--- a/docs/orchestration.md
+++ b/docs/orchestration.md
@@ -1,0 +1,89 @@
+# Orchestration: waits & suspend/resume
+
+A deployment pipeline often has to **pause for something outside the build** —
+manual QA sign-off, a soak period, another system's callback — and continue
+later, possibly days later and in a different process. Zuke models this as a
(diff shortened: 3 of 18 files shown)


totollygeek:
On the one finding (`fjc6xr`, low — "unauthorized run resumption via external signals lacks a capability check"): valid to think about, but not actionable in this PR, and it's owned by a later milestone.

- **This PR has no signal-delivery or resume path.** It's the *suspend* half only: a gate evaluates its trigger against the run's signals (always empty here) and, if unsatisfied, records `waiting`/`suspended` and exits. There is no `zuke resume`, no `signal_run`, no way to *deliver* a signal yet — so there is nothing here an unauthorized actor can invoke to resume a run.
- **Signal delivery + resumption are PR 2**, and their **authorization is M5** (MCP hardening): the plan gates the mutating `signal_run` / `resume_check` tools behind the allow-list, `--protect` operator tokens, and the audit log. That's the right layer for the capability check — the in-run gate here isn't an auth boundary.
- Resumption also already has an *exactly-once* CAS guard coming in PR 2 (`suspended → running`), so a delivered signal can't double-resume regardless.

No code change; flagging as a non-issue for the suspend half, to be enforced where delivery/resume actually land. All checks green; `deno task ci` clean (coverage 98.7% / 96.0%).

PR #170, opened 2026-07-17
feat(core): zuke resume with exactly-once run resumption

totollygeek:
## What

**M3, PR 2 of 2** — the **resume half** of external-event waits. A run suspended at a `.waitsFor()` gate (PR 1) is continued in a fresh process, possibly days later, with **exactly-once** semantics.

```sh
zuke resume <run-id> --signal testing-approved --data '{"by":"qa"}'
zuke resume --check [<run-id>]   # predicate waits + timeouts (cron/webhook)
```

### Resumption
- `resumeRun(build, { runId, signal?, data? })`: delivers the signal and transitions `suspended → running` under a **compare-and-swap so exactly one resumer wins** — the losers get a clean `AlreadyResumedError`. It re-resolves params from the record (CLI overrides non-secret; secrets from env), **verifies the graph still matches** (drift → hard error unless `--force-graph`), seeds the already-succeeded targets as done, and continues via `execute` — re-running only what hadn't finished. The gate re-evaluates its trigger (now satisfied) and the run proceeds, possibly suspending again at a later wait.
- `resumeCheck(build, { runId? })`: sweeps suspended runs, re-evaluating `resumeWhen` predicates and enforcing timeouts.
- **Timeouts:** a wait past its deadline fails the target + run; the `onTimeout` disposition is preserved for the cancellation milestone (M6) to act on.

### Plumbing
- `execute()` gains a resume seam (`ResumeState`); `RunStateWriter.adopt()` continues an existing record; a settled gate clears `waitingFor`.
- CLI: `resume` command + `--signal` / `--data` / `--force-graph` (`--check` reused).

### Docs & skills
- `docs/orchestration.md` gains the resume lifecycle; CLI reference, skill, and cheatsheet updated.

## Acceptance (the M3 scenario, end to end)
- **Three processes:** A runs deploy and suspends at the gate; B resumes with the signal → promote runs **exactly once** (deploy is *not* re-run, `ctx.state`/`ctx.signals` carry across); a concurrent C is rejected. Plus: `AlreadyResumedError` (deterministic), graph-drift rejection + `--force-graph`, timeout → fail, and a `resumeCheck` sweep that advances a flipped predicate.

## Verification
`deno task ci` green — fmt, lint, spell, `deno check`, coverage (lines **98.3%**, branches **95.8%**). `deno doc --lint` clean; `apiDocsCheck` no drift.

## Side note
You asked why the CLI is hand-rolled rather than using a Deno built-in — answered in-thread: Deno has no built-in command framework, `@std/cli` is a third-party dependency (and only a flag tokenizer), and this repo is deliberately dependency-free. Happy to adopt `@std/cli` for the tokenizer on a separate branch if you'd like, but it's not a real substitute for the command/help/completions machinery. Also still pending from earlier: **M1 PR2** (`zuke runs list/show`).

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -16,6 +16,8 @@
 | `zuke completions print <shell>`    | Print a shell-completion script (`bash`, `zsh`, or `fish`).    |
 | `zuke completions install <shell>`  | Write the script and wire it into the shell's startup.         |
 | `zuke mcp [--allow-run]`            | Run an MCP server over the build for AI agents ([details](./mcp.md)). |
+| `zuke resume <id> [--signal <n>] [--data <json>]` | Resume a suspended run, optionally delivering a signal ([details](./orchestration.md)). |
+| `zuke resume --check [<id>]`        | Re-check suspended runs (predicate waits, timeouts).           |
 | `zuke --help` / `-h`                | Usage.                                                         |
 | `zuke` (no target)                  | Run the `default` target if defined, else print `--list`.      |
 
@@ -237,3 +239,14 @@ build's `stateStore()` override. A plain run with none of these writes nothing.
 See [Durable run state](./state.md) for the record shape, `ctx.state`, the
 pluggable backends, and the [HTTP API](./state-api.md) for hosting a production
 store.
+
+## Resuming suspended runs
+
+A run parked at a [`.waitsFor()`](./orchestration.md) gate is continued with
+`zuke resume`. `--signal <name>` delivers a named external signal (with an
+optional `--data <json>` payload); `--check [<run-id>]` re-checks predicate waits
+and enforces timeouts across suspended runs (the cron/webhook entry point).
+Resumption is **exactly-once** — concurrent resumers race a compare-and-swap and
+all but one get `AlreadyResumedError` — and re-runs only the targets that hadn't
+yet succeeded. `--force-graph` continues even if the build graph changed since
+the run was suspended. See [Orchestration](./orchestration.md).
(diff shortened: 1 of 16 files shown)


totollygeek:
Addressed the finding (`164hr0i`, low — "resume `--data` accepts arbitrary JSON without size/shape validation"):

- **Size:** added a **64 KB cap** on the `--data` payload in `parseSignalData` — a webhook forwarding a large body can no longer bloat the persisted run record. Over-limit input fails with a friendly error (tested via `main(... "resume" ... --data <huge>)` → exit 1).
- **Shape:** the payload is **free-form by design** — a signal payload is build-specific, exactly like a parameter value, and the build reads `ctx.signals.get(name).data` and validates what it needs. Zuke can't know the shape. JSON.parse stores it as inert data (walked via `Object.entries`), so there's no prototype-pollution/injection path.

`deno task ci` green (coverage 98.4% / 95.7%), `apiDocsCheck` no drift.

totollygeek:
On the re-review's new low finding (`--data` echoed back in the parse-error message): fixed — the malformed-JSON error no longer echoes the payload (`"resume: --data is not valid JSON."`). Combined with the 64 KB size cap, arbitrary `--data` input is neither reflected nor able to bloat the record. `deno task ci` green.

totollygeek:
Acknowledged, but this one is by design — not changing it.

`AlreadyResumedError` reports the run id, the actor that already resumed it (`by`), and when (`at`). The `by` value is the `--actor` name: an operator/CI identity chosen by whoever ran the resume, not a secret, credential, or PII. Surfacing it is the whole purpose of the error — when two operators or CI jobs race to resume the same suspended run, the loser needs to know *who* won so it doesn't blindly retry. This mirrors `LockConflictError`, which likewise names the lock holder.

The review notes "no high-confidence exploitable security regressions are evident," and this is operational attribution working as intended, so I'm leaving the message as-is. Suppressing `1xwg7am`.

PR #171, opened 2026-07-17
feat(core): add zuke runs list/show

totollygeek:
## What

Closes **M1 PR2** — the read side of durable run state. Runs already
persisted to the `StateStore`, but nothing surfaced them from the CLI, so
M1's acceptance ("kill the process, then reconstruct full per-target status
from disk alone") wasn't actually reachable. This adds the `runs` command.

- `zuke runs list` — one row per run (id, status, target, actor, created),
  newest first. Filters: `--status`, `--target`, `--since` (compose). `--json`
  emits the summary array.
- `zuke runs show <id>` — one run in full: header, non-secret parameters,
  per-target status with duration / error / pending-wait, and any signals.
  `--json` emits the whole record.

## How

- New `packages/core/src/runs.ts`: store resolution (same precedence as a run —
  `ZUKE_STATE_URL`/`ZUKE_STATE_DIR` → `stateStore()` → default `.zuke/runs`) plus
  pure `formatRunList` / `formatRunDetail`.
- `cli_spec.ts` gains the `runs` command and the `--status`/`--target`/`--since`
  flags, so `--help`, completions, and `describeCli` stay in step (guarded by
  `cli_test.ts`). `--json` now defers to a command that consumes it, so
  `runs … --json` emits run data rather than the build surface. An invalid
  `--status` fails with the valid set listed.
- Docs: `docs/cli.md` (command table + "Inspecting runs"), `docs/state.md`
  (leads with `zuke runs`). Agent skill + cheatsheet updated. `llms.txt`
  regenerated.

## Verification

`deno task ci` green — coverage 98.4% lines / 95.7% branches. `apiDocsCheck`
clean, `deno doc --lint` clean over the four core entrypoints. New
`runs_test.ts` (13 cases) plus parser + `main()` dispatch tests in `cli_test.ts`.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -1,25 +1,27 @@
 # CLI reference
 
-| Command                             | Behaviour                                                      |
-| ----------------------------------- | -------------------------------------------------------------- |
-| `zuke <target>`                     | Run the target and all its transitive dependencies, in order.  |
-| `zuke <target> --skip <dep>`        | Run the target but skip the named dependency (repeatable).     |
-| `zuke <target> --parallel`          | Run independent targets concurrently (`--parallel=N` caps it). |
-| `zuke <target> --no-cache`          | Ignore the incremental cache; re-run every target.             |
-| `zuke <target> --affected[=<base>]` | Run only targets affected by files changed since a git base.   |
-| `zuke <target> --dry-run`           | Print the plan without executing any target body.              |
-| `zuke <target> --state`             | Persist [durable run state](./state.md) under `.zuke/runs`.    |
-| `zuke <target> --actor <name>`      | Attribute the run to `<name>` in its state record.            |
-| `zuke --list` / `-l`                | List all targets with descriptions and dependencies.           |
-| `zuke graph`                        | Print the dependency graph (`target → deps`).                  |
-| `zuke graph --output=html`          | Render an interactive HTML graph into `.zuke/` and open it.    |
-| `zuke completions print <shell>`    | Print a shell-completion script (`bash`, `zsh`, or `fish`).    |
-| `zuke completions install <shell>`  | Write the script and wire it into the shell's startup.         |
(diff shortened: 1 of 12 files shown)


totollygeek:
Acknowledged, and by design — not adding a gate here.

`zuke runs list/show` is a **local, read-only CLI command** that reads a store the operator has already configured and can access directly. The authorization boundary is the layer that already owns access: filesystem permissions on `.zuke/runs` for the FS backend, or the bearer token on the HTTP backend. A local inspect command can't add authorization on top of files its invoker can already `cat` — and shouldn't pretend to.

Two further points:

- **The record is non-secret by contract.** Secret parameters are structurally excluded and `ctx.state`/signal values pass through the redactor before persistence (see `docs/state.md`). `runs` only ever surfaces that already-non-secret data.
- **Agent/network authz is M5.** Exposing runs to a *remote* caller is the MCP surface (`list_runs`/`show_run`), which lands in M5 with the authz tiers, operator tokens, and audit log. This PR's own docs and body call that out. The CLI command is not a network-exposed surface.

The review notes this is "expected." Leaving the behaviour as-is and suppressing `3f7a0g`; the "treat run metadata as sensitive in shared environments" guidance is already documented in `docs/state.md`.

totollygeek:
This is the same finding as above (`3f7a0g`), re-fingerprinted — suppressing the line-54 fingerprint just re-surfaced the identical concept at `runs.ts:82` under a new id (`16mb980`), now worded around `--json`. My earlier rebuttal stands unchanged: `zuke runs` is a local, read-only inspect of an operator-owned store of non-secret records; the FS/HTTP layer already owns access, and remote/agent authz is the M5 MCP surface.

I'm not going to keep adding suppressions for it — each suppression commit re-triggers the review and re-anchors the same finding at a new line/id, so chasing the fingerprint is a treadmill. It's a low, non-gating finding (score 2 ≪ the score-8 gate; all checks green) and the "treat run metadata as sensitive in shared environments" guidance is already in `docs/state.md`.

PR #173, opened 2026-07-17
feat(core): forEach fan-out targets with per-item isolation and reporting

totollygeek:
## What

Closes **M4** — dynamic fan-out over a runtime list (P0 #4). A build can now run
the same pipeline over N items (repos, tenants, regions) with per-item failure
isolation and bounded concurrency, without a hand-written target per item.

```ts
repos = parameter("services").array().required();

deployBatch = target().forEach(
  () => this.repos.value,
  (repo) => ({
    checks: target().executes(() => checkDeployable(repo)),
    deploy: target().executes((ctx) => applyToSit(repo, ctx)),
  }),
  (s) => s.concurrency(3).continueOnItemFailure(),
);
```

## How

- **`.forEach(items, factory, configure?)`** on `TargetBuilder`. `items` is a
  thunk (read after params resolve); `factory` returns an ordered record of
  sub-targets, each stage implicitly depending on the previous. Options are a
  **fluent settings lambda** (AGENTS.md guideline 9): `.concurrency(n)`,
  `.continueOnItemFailure()`. The item type is captured in a closure, so the
  stored spec stays non-generic with no `as`.
- **Execution reuses the scheduler**: `runForEachTarget` materialises sub-targets
  (`parent[item].stage`), builds a sub-plan (per-item chains), and drives it with
  the existing `runScheduled` — items concurrent up to the cap, stages sequential
  per item (the pipeline model, no barrier between items). `continueOnItemFailure`
  maps to lenient per-sub-target failure.
- **First-class sub-targets**: each gets a summary row (via a new
  `TargetOutcome.children`) and a run-record entry (dynamic keys the writer
  already seeds), so `zuke runs show` reports per-item verdicts. `--list`/`graph`
  annotate the one node `[fan-out]`.
- **Parameter polish** (finishes #4's "end of JSON-in-a-text-field"): `.array()`
  now validates each element through the scalar parser, so
  `.options(...).array()` checks every choice and `.number().array()` → `number[]`.

## Verification

`deno task ci` green — coverage 98.4% lines / 95.6% branches. `apiDocsCheck` and
`deno doc --lint` (four core entrypoints) clean. New `foreach_test.ts` (13 cases:
ordering, isolation, halt-on-failure, concurrency caps, runtime list, empty list,
per-item record entries, the `[fan-out]` annotation, and the array-param polish).
Docs: `orchestration.md` (fan-out section), `parameters.md`; skill + cheatsheet
updated.

### DIFF
--- a/docs/orchestration.md
+++ b/docs/orchestration.md
@@ -168,8 +168,13 @@ class CD extends Build {
 Sub-targets are materialised at run time, named `deployBatch[<item>].<stage>`,
 and are **first-class**: each gets its own row in the summary and its own entry
 in the [run record](./state.md) (so `zuke runs show` reports per-item verdicts).
-Static views — `zuke --list`, `zuke graph` — show only the one fan-out node
-(annotated `[fan-out]`), since the sub-targets exist only while the build runs.
+The item key is a scalar item's own value (else its index), so it becomes part
+of the target name in output and the record — **fan out over non-secret
+identifiers** (repo names, tenant ids), never over `.secret()` values, which are
+meant for parameters (they are excluded from the record) rather than as loop
+keys. Static views — `zuke --list`, `zuke graph` — show only the one fan-out
+node (annotated `[fan-out]`), since the sub-targets exist only while the build
+runs.
 
 Runtime lists pair naturally with [array parameters](./parameters.md#lists):
 `.options("api", "web").array()` validates each element, and `.number().array()`
--- a/docs/orchestration.md
+++ b/docs/orchestration.md
@@ -29,7 +29,8 @@ class Deploy extends Build {
     .waitsFor((s) =>
       s.on(externalSignal("testing-approved"))
         .timeout("72h")
-        .onTimeout(() => this.rollback));
+        .onTimeout(() => this.rollback)
+    );
 
   promoteToProd = target()
     .dependsOn(this.awaitTesting)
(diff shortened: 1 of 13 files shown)


totollygeek:
Fair kernel, handled by documentation rather than code — here's the reasoning.

A fan-out item key becomes part of the sub-target name (`deployBatch[<item>].<stage>`), which appears in output and as a key in the run record. But:

- **Item keys are identifiers, by design.** They exist to correlate an item's stages across the summary, the record, and `zuke runs show`. Redacting the key itself would collapse every item to the same name and break that correlation — so the key must stay a readable label, exactly like a target name or a parameter name (none of which are secret channels).
- **Console output is already redacted.** All reporter output passes through the run's `Redactor`, so a secret value that reached a name would be masked in printed logs. The one residual is the record's JSON key, which can't be redacted without breaking correlation (above).
- **Secrets have a different home.** `.secret()` parameters are structurally excluded from the record and redacted from output. Fanning out *over* secret values is an anti-pattern — you iterate over repo names / tenant ids, not credentials.

So the fix is to state the contract, not to add code that breaks the feature: `docs/orchestration.md` now says to fan out over non-secret identifiers, never over `.secret()` values. Suppressing `z2fmcx`. `deno task ci`, `apiDocsCheck`, and `deno doc --lint` remain green.

totollygeek:
Same finding as above, re-fingerprinted — suppressing `z2fmcx` re-surfaced the identical concern (fan-out keys in sub-target names/records) at `target.ts:731` under a new id. My prior comment stands: keys are author-chosen identifiers used for per-item correlation, console output is redacted, and secrets belong in excluded `.secret()` params — now documented in `docs/orchestration.md`.

Not adding further suppressions: each one re-triggers the review and re-anchors the same finding at a new line/id. It's low and non-gating (score 2 ≪ the score-8 gate; all checks green).

PR #174, opened 2026-07-18
feat(core): MCP streamable-HTTP transport

totollygeek:
## What

First half of **M5** — a non-stdio transport for `zuke mcp`. Adds
`zuke mcp --http <host:port>`, so a client can reach the MCP server over the
network rather than launching the process. (Authz tiers, the audit log, and the
store-backed run tools — `list_runs`/`signal_run` — are the second M5 PR.)

```sh
zuke mcp --http 7777              # bind 127.0.0.1:7777 (local only)
zuke mcp --http 0.0.0.0:7777      # all interfaces — requires ZUKE_MCP_TOKEN
```

## How

- **New `mcp/http.ts`** — MCP streamable-HTTP over `Deno.serve`, dependency-free,
  feeding the same `McpServer.handleMessage` as stdio. A `POST` body is one
  JSON-RPC message; the reply is its JSON-RPC response. Notification → `202`;
  bad JSON → `400` parse error; non-`POST` → `405` (Zuke never initiates
  server→client messages, so the optional SSE stream is omitted — clients fall
  back to POST-only, which is spec-compliant).
- **Serial handling** — messages are processed one at a time via a promise
  chain, because `execute` resolves parameters into shared per-run state, so
  concurrent handling of two POSTs would race. Mirrors the stdio invariant.
- **Security defaults (bridge, not gateway)** — loopback bind by default; a
  non-loopback bind *requires* a bearer token (`ZUKE_MCP_TOKEN`) or the server
  refuses to start; a configured token is checked on every request with a
  constant-time compare (`401` on missing/wrong, `WWW-Authenticate: Bearer`).
- **Wiring** — `serveMcp` gains an HTTP mode (`command.ts`); the CLI adds
  `--http` with address parsing (`<port>` or `<host:port>`), surfaced in
  `--help`/completions via `cli_spec.ts`.

## Verification

`deno task ci` green — coverage 98.4% lines / 95.5% branches. `apiDocsCheck` and
`deno doc --lint` clean. New `mcp_http_test.ts` (7 cases: POST round-trip, 405,
400, 202, the token matrix, `serveMcp` HTTP, and the non-loopback refusal) plus
CLI parser/dispatch tests. Smoke-tested live: `ping` over HTTP returns
`result: {}`, `GET` → `405`. Docs (`mcp.md` HTTP + trust-model, `cli.md`), skill,
and cheatsheet updated.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -15,7 +15,7 @@
 | `zuke graph --output=html`                          | Render an interactive HTML graph into `.zuke/` and open it.                             |
 | `zuke completions print <shell>`                    | Print a shell-completion script (`bash`, `zsh`, or `fish`).                             |
 | `zuke completions install <shell>`                  | Write the script and wire it into the shell's startup.                                  |
-| `zuke mcp [--allow-run]`                            | Run an MCP server over the build for AI agents ([details](./mcp.md)).                   |
+| `zuke mcp [--allow-run] [--http <host:port>]`       | Run an MCP server over the build for AI agents, on stdio or HTTP ([details](./mcp.md)). |
 | `zuke resume <id> [--signal <n>] [--data <json>]`   | Resume a suspended run, optionally delivering a signal ([details](./orchestration.md)). |
 | `zuke resume --check [<id>]`                        | Re-check suspended runs (predicate waits, timeouts).                                    |
 | `zuke runs list [--status/-target/-since] [--json]` | List persisted run records, newest first ([details](./state.md)).                       |
@@ -111,7 +111,10 @@ build on stdio, so an AI agent can operate the pipeline through typed tool calls
 instead of guessing shell commands. It exposes read tools (`list_targets`,
 `describe_build`, `graph`) and — only with `--allow-run` — one `run:<target>`
 tool per target, whose input schema is derived from the build's parameters.
-`mcp` is a reserved command name. See the full guide: [MCP server](./mcp.md).
+`--http <host:port>` serves the streamable-HTTP transport instead of stdio
+(loopback by default; a non-loopback bind requires a `ZUKE_MCP_TOKEN` bearer
+token). `mcp` is a reserved command name. See the full guide:
+[MCP server](./mcp.md).
 
 ## Parallel execution
 
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -39,6 +39,40 @@ Any client that speaks stdio MCP works the same way — give it the command
 `deno run -A zuke.ts mcp` (add `--allow-run` when you want the agent to execute
 targets).
 
+## HTTP transport
+
+For a client that connects over the network rather than launching the process,
+`--http` serves MCP's **streamable-HTTP** transport instead of stdio:
+
+```sh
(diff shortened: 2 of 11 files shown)


totollygeek:
Good catch — fixed in the latest push (not a rebuttal).

The auth check now parses the `Authorization` header rather than comparing the whole raw string: it splits off the scheme, matches `Bearer` **case-insensitively** (per RFC 7235's case-insensitive auth-scheme), tolerates surrounding/extra whitespace, then **constant-time compares the token itself** against `ZUKE_MCP_TOKEN`. So trivial client formatting differences (`bearer`, extra spaces) no longer fail auth, while the token value still must match exactly — the security property is unchanged.

Parsed by hand rather than with a regex, to keep it linear on adversarial input (no `\s*$`-style backtracking). A new test asserts `"bearer   swordfish"` authenticates while a wrong token still gets 401. `deno task ci` green.

totollygeek:
Tightened in the latest push. (The prior version wasn't exploitable — a credential with trailing content was compared *exactly*, so `Bearer <token> <junk>` yielded `"<token> <junk>"` ≠ `<token>` → 401 — but you're right that it should reject a multi-word credential outright per RFC 6750's `token68`.)

The parser now splits the header on whitespace and requires **exactly two parts** — the scheme and a single token — matching the scheme case-insensitively. `Bearer <token> <anything>` → `undefined` → 401; `bearer   <token>` still authenticates. `split(/\s+/)` is a linear scan, so no backtracking risk. Added a test asserting the trailing-content case is rejected. `deno task ci` green.

totollygeek:
Both of these are the documented design, not defects — I'm holding the code here. (I did take the two actionable token-parsing findings earlier: case-insensitive scheme, single-token `token68`.)

**`1ptdbbb` — "unauthenticated access on loopback."** By design, and documented. A `127.0.0.1` bind is reachable only from the same host; anyone who can reach it can already run `deno run -A zuke.ts <target>` directly, so there is no remote party to authenticate — exactly the trust model of the default stdio transport (also unauthenticated). Requiring a token to talk to your own loopback would be security theatre. This is the milestone's stated default ("binds 127.0.0.1 by default; non-loopback binds require a bearer token"), and a token *is* still enforced on loopback if `ZUKE_MCP_TOKEN` is set.

**`t5xtih` — "brute force if misconfigured behind a reachable proxy."** This is a hypothetical about a misconfiguration the docs explicitly warn against. The token is checked **before** the body is read or any build logic runs (only a `405`/`401` is reachable unauthenticated), with a constant-time compare, so there's no timing oracle. Rate-limiting / lockout and TLS are the job of the fronting gateway — `docs/mcp.md` says in as many words: "put real TLS and authentication in front of it … Zuke provides the transport, not an internet gateway." Building an auth gateway into a dependency-free bridge is out of scope for this milestone (per the plan: mTLS/OIDC belong in the consumer's hosting layer).

Net: loopback-default + mandatory-token-off-loopback + fail-closed refusal + constant-time compare is the intended surface. Leaving both as-is (score 4/10, non-gating; all checks green). Not adding suppressions, since each push just re-anchors the same design discussion at a new line.

PR #176, opened 2026-07-18
feat(core): MCP authz tiers, audit log, and run-state tools

totollygeek:
## What

Completes **M5** — the MCP hardening half (after PR1's HTTP transport). Turns
`zuke mcp` into a store-driven, tiered-authorization surface with an audit
trail, so the M5 acceptance holds: session A starts work and disconnects, and a
separate session B queries the run and signals it — every mutating call
attributed.

## How

- **Store-backed run tools** — `list_runs`/`show_run` (read-only, exposed when a
  store resolves) and `signal_run`/`resume_check` (mutating, gated), thin
  wrappers over the same `StateStore` / `resumeRun` surfaces the CLI uses.
- **`.readOnly()`** — a target advertises MCP `readOnlyHint` instead of the
  default `destructiveHint`, and is exempt from confirmation.
- **Authorization tiers:**
  - `--allow-run=<globs>` — a per-target allow-list. An unlisted target is
    **invisible**, and its denial is indistinguishable from a nonexistent tool
    (leaks no existence).
  - `--protect <globs>` + `ZUKE_OPERATOR_TOKEN` — a constant-time-checked
    operator token, **fail-closed** (no token configured → always denied).
  - `--confirm-destructive` — a destructive call returns its plan until
    re-sent with `confirm:true`.
  - The **resume tools enforce the same gate** before running a target's code.
- **Audit log** — every mutating/denied call appends a redacted `RunEvent`
  (time, tool, resolved actor, outcome, args) to a store-level `mcp-audit`
  record; secrets are masked wherever they appear, the operator token is never
  persisted. `zuke runs show mcp-audit` prints it. `events[]` parses
  backwards-compatibly.
- **Structured JSON errors** for lock conflicts, `AlreadyResumedError`, and
  denials; a thrown framework error returns a structured result, never crashing
  the transport.

## Testing (new three-layer mandate)

- **Unit** — `mcp_authz_test.ts` (matcher + constant-time compare),
  `mcp_hardening_test.ts` (the allow-list / protect / confirm / audit / run-tool
  matrix, 18 cases).
- **Integration** — `tests/integration/mcp_audit_test.ts`: the audit trail
  rendered through `runs show` via the real `main()`.
- **E2E** — `tests/e2e/mcp_e2e.ts`: two real processes — a suspended run, then an
  MCP server session that queries and signals it over HTTP, asserting the audit
  attribution (wired into the `integration` build target).

## Adversarial review

Per the standing practice, I ran an adversarial security review of the authz/
audit surface **before** this PR. It found (and I fixed + regression-tested)
three real defects the green gate missed: `signal_run`/`resume_check` bypassing
the allow-list and operator token, an inert audit redactor that let a secret
echoed into a non-secret argument persist, and an unguarded `execute()` that
could crash the transport on a framework throw.

## Verification

`deno task ci` green — coverage 97.8% lines / 95.3% branches. `apiDocsCheck` and
`deno doc --lint` clean. `integration` target (e2e lane) passes. Docs, skill, and
cheatsheet updated. Design was chosen from a scored panel of independent
approaches (minimal/orthodox won, security-first ideas grafted).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -18,8 +18,8 @@ exact signatures are published — read them:
 - **One file with the whole typed surface of every package:**
   [`llms-full.txt`](./llms-full.txt) at the repo root. [`llms.txt`](./llms.txt)
   is the short index.
-- **A single wrapper on the command line:** `deno doc jsr:@zuke/<package>`
-  (e.g. `deno doc jsr:@zuke/deno`).
+- **A single wrapper on the command line:** `deno doc jsr:@zuke/<package>` (e.g.
+  `deno doc jsr:@zuke/deno`).
 - **On each package's JSR page / README:** a generated `## API` section.
 - **The CLI surface — commands, flags, and a build's actual targets:** run
   `zuke --help` (or `deno run -A zuke.ts --help`). It prints the usage grammar,
@@ -39,10 +39,10 @@ The mental model:
 - A build is a class that **extends `Build`**. Each **target is a class field**
   built with `target()`: `.description(...)`, `.dependsOn(...)`,
   `.executes(async () => { … })`.
-- **Dependencies are `this.<field>` references, not strings** — `dependsOn(this.lint)`,
-  never `dependsOn("lint")` — so renames and typos are compile-time errors. A
-  target may only depend on siblings **declared above it** (fields initialise
-  top-to-bottom).
+- **Dependencies are `this.<field>` references, not strings** —
+  `dependsOn(this.lint)`, never `dependsOn("lint")` — so renames and typos are
+  compile-time errors. A target may only depend on siblings **declared above
+  it** (fields initialise top-to-bottom).
 - Make the file runnable with **`await run(MyBuild)`** at the bottom — no
   `if (import.meta.main)` guard; `run` no-ops when the module is imported.
 - **Every external tool is a namespaced `*Tasks` object** (`DenoTasks`,
@@ -56,7 +56,9 @@ The mental model:
   class CI extends Build {
     lint = target().executes(() => DenoTasks.lint());
     test = target().dependsOn(this.lint)
-      .executes(() => DenoTasks.test((s) => s.allowAll().coverage("cov_profile")));
+      .executes(() =>
+        DenoTasks.test((s) => s.allowAll().coverage("cov_profile"))
+      );
   }
 
   await run(CI);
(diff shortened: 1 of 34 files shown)


totollygeek:
Both findings looked at hard (a HIGH token-bypass claim earns real scrutiny) — neither holds. Details:

**`1ownw8s` (high) — "protected signal_run/resume_check ignore the operator token" — false positive.** The gate *is* enforced. `signalRun` (runtools.ts:262-267) and `resumeCheckTool` (both the single-id and sweep paths) call `deps.authorize(record.rootTarget, args)` before resuming; the server wires `authorize` to `#authorizeTarget` → `#checkOperatorToken` (server.ts), the same constant-time operator-token check `run:` uses. Fail-closed included (no `ZUKE_OPERATOR_TOKEN` configured → always denied). The cited line 188 is the `structuredError` branch, not the gate. Two tests prove it: `resume tools require the operator token for a protected target` asserts `missing`/`invalid` denials for both `signal_run` and `resume_check` (I just added the `resume_check` case in this push). I suspect the reviewer tokenized the branch's two-commit diff (a WIP commit before the gate, then the fix) and scored the intermediate state; HEAD enforces it.

**`1eav335` (medium) — "list_runs/show_run exposed regardless of --allow-run" — by design.** These two tools are **read-only** and operate over **non-secret** run records (secret params are structurally excluded; `ctx.state`/args are redacted). M5 specifies them as "read-only, always exposed when a store resolves" — the *mutating* tools (`signal_run`/`resume_check`) are the ones gated behind `--allow-run` (and the allow-list/operator-token). Access to the endpoint is already gated by the transport (loopback stdio, or a bearer token off-loopback). This mirrors the CLI `zuke runs`, which any local user can already run.

Both are non-gating (score 7 < the score-8 threshold; all checks incl. E2E on three OSes are green). Suppressing both with the rationale in `zuke.ts`.

totollygeek:
These three are re-fingerprinted versions of concerns already addressed on this PR — I'm not adding further suppressions, since each push just re-anchors the same points at new ids.

- **`tcr3yz` — "run-state tools don't enforce the allow-list."** They do. `signalRun`/`resumeCheckTool` call `deps.authorize(record.rootTarget, args)` (runtools.ts:264, and both the single-id and sweep paths of resume_check), which runs the server's `#authorizeTarget` = allow-list (`#allowMatch`) **plus** operator token (`#checkOperatorToken`). This is the same `tcr3yz`/`1ownw8s` finding re-worded from "operator token" to "allow-list"; both are covered by `resume tools respect the allow-list` and `resume tools require the operator token` tests. Line 188 is the `structuredError` branch, not the gate.

- **`1cd5n44` — "audit shown via runs show without a redaction gate."** Redaction happens at **write** time: `#auditArgs` masks every secret value (even echoed into a non-secret field) and drops the operator token before the event is persisted (with a test). `runs show` renders already-redacted data — there is nothing unredacted to gate. Reading it is the same local FS-permission boundary as any `zuke runs` command.

- **`12pq4y0` — "clientInfo attribution on a shared HTTP server."** Known and documented (the `#initialize` comment and docs/mcp.md both say: on a shared endpoint clientInfo reflects the most recent client, so set `--actor` for authoritative attribution). clientInfo is the **lowest-priority** actor candidate and is **never an authorization input** — only a label. Authoritative per-client attribution needs MCP session tracking, which is out of scope for this bridge (loopback/tokened HTTP).

The real defects an adversarial pass found earlier this PR (a resume-path authz bypass, an inert audit redactor, an unguarded execute()) are fixed and regression-tested. Score 5/10, non-gating; all checks — including E2E on macOS/Ubuntu/Windows — are green.

PR #177, opened 2026-07-18
feat(core): compensation targets and zuke cancel

totollygeek:
## M6 — Cancellation as a graph concept

Implements milestone **M6** from `plans/orchestrationplan.md`: cancelling a run is always safe and complete — compensations run in reverse order, locks release, and the record settles `cancelled`.

### Authoring
- **`.onCancel(target | () => target)`** on `TargetBuilder` — a compensation that runs **iff its target succeeded**. On cancel, the compensations of the succeeded targets run in **reverse topological order**, each with a normal `TargetContext` whose `ctx.state` exposes **the original target's** persisted metadata (so a rollback reads exactly the slot the deploy recorded). Thunk form for forward references.

### Cancelling — three entry points, one walk
- **`zuke cancel <run-id>`** — cancel any persisted run (idempotent; a finished run is a friendly no-op).
- **Ctrl-C / SIGTERM** — wired in `run()`; the in-process run runs the same walk. A second Ctrl-C force-exits.
- **MCP `cancel_run`** — gated by the allow-list + operator token and audited, exactly like `signal_run`.
- A **live** run being cancelled by another process observes `cancelling` on its next state write and aborts (releasing its locks via the ambient signal); the canceller owns the compensation walk.
- A timed-out `.waitsFor()` routes its `cancel-run` / named-target `onTimeout` disposition through this machinery, closing the stuck-deploy → wait loop.

### New status
- `cancelling` — the transient state between `running`/`suspended` and `cancelled`.

### Robustness (from the adversarial review — see below)
Cleanup is **maximal and safe**: a throwing `onCancel` thunk or compensation body, a hung compensation (bounded by its own `.timeout()`), and a store hiccup mid-finalize are all recorded rather than fatal; a run stranded `cancelling` by an interrupted cancellation is **recoverable** by re-running the cancel; compensation failure messages are **redacted**; and `resume --check` isolates a per-run error instead of aborting the sweep.

### Tests (all three layers)
- **Unit** `packages/core/tests/cancel_test.ts` — reverse-order walk, meta read-back, idempotent no-op, in-process cancel, live observe-hook, recovery, redaction, timeout, throwing thunk, timeout dispositions.
- **Integration** `tests/integration/cancel_test.ts` — the full flow through `main()` (`zuke cancel` → `runs show`).
- **E2E** `tests/e2e/cancel_e2e.ts` — two real processes: one suspends, the other cancels.
- Plus MCP `cancel_run` coverage and the `resumeCheck` isolation regression.

### Adversarial review
An adversarial-review workflow (4 dimensions → per-finding verify) found **13 real defects** the green gate missed — an unguarded `onCancel` thunk that wedged the run, a secret leak through unredacted compensation errors, a double-compensation race, a missing in-process audit event, a silently-stranded `cancelling` run, a sweep-aborting `resume --check`, and a dropped-succeeded-settle. **All fixed with regression tests before this PR.**

### Docs
`docs/orchestration.md` (cancellation section + fan-out limitation), `cli.md`, `authoring.md`, `mcp.md`, `state.md`; skill + cheatsheet; `llms.txt`/`llms-full.txt`/README regenerated.

`deno task ci` green — 98.1% lines / 95.1% branches; `apiDocsCheck` and `deno doc --lint` clean.

### DIFF
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -186,6 +186,12 @@ deploy = target()
   target's own dependents are skipped.
 - **`.always()`** — run even after the build has already failed, for
   cleanup/teardown. It still waits for its own dependencies to complete.
+- **`.onCancel(target | () => target)`** — register a **compensation** that
+  undoes this target when the run is [cancelled](./orchestration.md#cancellation--compensation-oncancel).
+  It runs only if this target **succeeded**; on cancel, compensations run in
+  reverse order. The compensation body's `ctx.state` exposes _this_ target's
+  persisted metadata (so a rollback reads what the deploy recorded). Use the
+  thunk form to reference a compensation declared below. Needs a state store.
 - **`.unlisted()`** — hide a helper target from `--list`/`--help`; it can still
   be run by name or depended on.
 - **`.readOnly()`** — mark a target query-only for [MCP](./mcp.md): its run tool
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -20,6 +20,7 @@
 | `zuke resume --check [<id>]`                            | Re-check suspended runs (predicate waits, timeouts).                                    |
 | `zuke runs list [--status/-target/-since] [--json]`     | List persisted run records, newest first ([details](./state.md)).                       |
 | `zuke runs show <id> [--json]`                          | Show one run's full per-target status and metadata.                                     |
+| `zuke cancel <id> [--actor <name>]`                     | Cancel a run and run its compensations ([details](./orchestration.md#cancellation--compensation-oncancel)). |
 | `zuke --help` / `-h`                                    | Usage.                                                                                  |
 | `zuke` (no target)                                      | Run the `default` target if defined, else print `--list`.                               |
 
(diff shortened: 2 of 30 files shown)


totollygeek:
Thanks — the **generic review is clean (0 findings)**, and I've verified both **security** findings against the code. Both are by-design consistency observations rather than exploitable defects; responding to each per our review rule:

**`10gkmqr` (high) — cancel_run authorization uses the target's allow-list, not a separate cancel permission** — *intended and consistent.* `cancel_run` gates on `authorize(record.rootTarget, args)` (`runtools.ts:373`), the **same allow-list + operator-token gate** as `signal_run` and `resume_check` — the mutating run-state-tool model reviewed and accepted in M5 (PR #176). The rationale is identical: cancelling runs the target's **author-declared compensation code**, so it must ride the target's own run permission. It is also gated behind `--allow-run` (`server.ts` `#callRunStateTool`), so a read-only server never exposes it. A run whose root target isn't in the current allow-list can't be cancelled (`not_allowed`), and a protected target's runs require the operator token — conservative, not over-broad. A separate cancel-permission tier isn't in the M6 plan and would fragment the one-permission-per-target model; if finer control is wanted later it belongs in a follow-up, not here.

**`2ytv89` (medium) — cancellation via Ctrl-C has no confirmation barrier** — *intended UX.* The SIGINT/SIGTERM handler (`cli.ts` `run()`) is a **local operator action on their own terminal**; it runs the run's **compensations — cleanup/rollback, the safe direction — not a destructive escalation**, which is exactly M6's goal ("cancelling a run is always safe and complete"). A confirmation prompt on Ctrl-C would trap the operator and contradict normal CLI behavior; the design instead makes a **second Ctrl-C force-exit** (`cli.ts` `onSignal`) so a stuck build can never wedge them. `--confirm-destructive` is the belt for MCP *agents* that ignore `destructiveHint`, not for a human's interrupt.

For context: an adversarial-review pass on this PR before opening found and fixed **13 real defects** the gate missed (a run-wedging `onCancel` throw, a secret leak through unredacted compensation errors, a double-compensation race, a stranded `cancelling` run, a sweep-aborting `resume --check`), each with a regression test. Score 7/10, non-gating; all checks green (Tests + E2E on macOS/Ubuntu/Windows), 98.1% lines / 95.1% branches.

PR #179, opened 2026-07-18
feat(core): richer plugin lifecycle payloads

totollygeek:
## M7 PR1 — richer plugin lifecycle payloads

The first of two PRs for **M7 (OpenTelemetry export)**. This one enriches the core plugin lifecycle so an observer can group, time, and count a run — the foundation the `@zuke/otel` exporter (PR2) builds on. No new package; core stays dependency-free.

### Enriched `Plugin` hooks
- `onStart(run)` / `onTargetStart(target, run)` / `onFinish(result, run)` carry a **`RunInfo`** (`{ runId, dryRun }`).
- `onTargetEnd(target, status, timing)` carries a **`TargetTiming`** (`{ runId, durationMs }`).
- New **`onRunStateChange(record)`** fires on each run-level durable status change (`running` → … → terminal) with the full [`RunRecord`](docs/state.md) — per-target timings, waits, and the audit trail in one payload. Fires only when a state store is configured.
- The **run id is stable across a suspend/resume boundary**, so an exporter's spans join one trace; `resumeRun`/`resume --check` forward plugins so a resumed run is observed under the same identity.

### Backwards compatible
The extra arguments are **additive** — a plugin written against the old signatures (`onTargetEnd: (target, status) => …`) keeps working unchanged, since a function that ignores its trailing arguments stays assignable. Existing plugin tests pass untouched.

### Robustness (from the adversarial review)
A plugin is an **observer** (the contract is "report/time/notify, don't change the run), so a **throwing plugin hook is now caught and reported** rather than breaking the run — the build's own hooks still propagate. `onRunStateChange` delivers the full `running → cancelling → cancelled` sequence on an in-process cancel; the cross-process cancel case (the owning process observes through `cancelling`) is documented on the hook.

### Tests
- **Unit** (`executor_test.ts`): run id / dry-run / timing on every hook, `onRunStateChange` running→terminal (and the cancel sequence), no-store silence, throwing-hook isolation, and old-signature compatibility.
- **Integration** (`tests/integration/plugin_test.ts`): the enriched lifecycle through the real CLI (`--state`).
- **Resume continuity** (`resume_test.ts`): a resumed run observed under the original run id.

### Adversarial review
A focused adversarial-review workflow (2 lenses → verify) surfaced 5 low-severity findings — all in the `onRunStateChange`/plugin-isolation area — each addressed before opening: `cancelling` is now delivered, plugin hooks are isolated, and the cross-process terminal-status nuance is documented.

`deno task ci` green — 98.x% lines / 95%+ branches; `apiDocsCheck` + `deno doc --lint` clean. Next: PR2 `feat(otel): OTLP export plugin`.

### DIFF
--- a/docs/extending.md
+++ b/docs/extending.md
@@ -20,23 +20,36 @@ import { type Plugin, run } from "jsr:@zuke/core";
 
 const timing: Plugin = {
   name: "timing",
-  onTargetEnd: (target, status) => console.log(`${target}: ${status}`),
+  onTargetEnd: (target, status, { runId, durationMs }) =>
+    console.log(`${runId} ${target}: ${status} in ${durationMs}ms`),
   onFinish: (result) => console.log(`done: ${result.ok ? "ok" : "failed"}`),
 };
 
 await run(MyBuild, { plugins: [timing] });
 ```
 
-| Hook                       | When it runs                                          |
-| -------------------------- | ----------------------------------------------------- |
-| `onStart()`                | Once, before any target runs.                         |
-| `onTargetStart(target)`    | Before a target's body executes (not skipped/cached). |
-| `onTargetEnd(target, s)`   | After each target settles, with its status.           |
-| `onFinish(result)`         | Once, after the run (success or failure).             |
+| Hook                                  | When it runs                                                        |
+| ------------------------------------- | ------------------------------------------------------------------- |
+| `onStart(run)`                        | Once, before any target runs.                                       |
+| `onTargetStart(target, run)`          | Before a target's body executes (not skipped/cached).               |
+| `onTargetEnd(target, status, timing)` | After each target settles, with its status and duration.            |
+| `onFinish(result, run)`               | Once, after the run (success or failure).                           |
+| `onRunStateChange(record)`            | On each run-level status change; needs a [state store](./state.md). |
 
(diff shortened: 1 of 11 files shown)


totollygeek:
Thanks — on the one **low** finding (`New run-state callback exposes full durable run records to plugins`, plugin.ts):

Verified against the code, and this is by-design, not an exposure. The `RunRecord` handed to `onRunStateChange` is the **secret-free, redacted projection**, not raw run state:

- `secret()` parameters are **structurally omitted** from `params` (`buildRunRecord` in `state/record.ts` skips every `secret_` param).
- `ctx.state` metadata, target `error` messages, and audit-event arguments are all run through the run's `Redactor` **before** they land in the record (`RunStateWriter.stateHandle` / `markTargetSettled` / `appendEvent`).

So the record a plugin sees is exactly the same data already **persisted to the state store** and printed by `zuke runs show <id>` — no additional surface. And plugins are **author-registered trusted code** (`run(MyBuild, { plugins })`), not untrusted third parties; handing the exporter the full record from one source is the intended design (it derives spans, wait durations, and counters without re-reading the store).

I've added an explicit note to the `onRunStateChange` JSDoc stating the record is the secret-free projection and safe to export, so the guarantee is visible at the call site (commit 11d6ae5).

Score 1/10, non-gating; all other checks green.

PR #186, opened 2026-07-19
feat(core): pluggable BuildRegistry and zuke register

totollygeek:
## What

Milestone **M11 PR1** — a pluggable **build registry**: a catalog of the pipelines (builds) that exist and where they live. Where durable run state records *runs*, the registry records *builds*, so a later registry-backed `zuke mcp` server (PR2) can discover newly-registered pipelines without a redeploy.

Kept a **separate concern** from the run `StateStore` but configured the same way; over HTTP it rides the same `docs/state-api.md` contract with a `/builds` collection beside `/runs`, so one service can host both.

## Surface

- **`BuildRegistry`** interface — `getBuild` / `register` / `deregister` / `listBuilds`, with whole-record compare-and-swap.
- **`BuildDescriptor`** — a versioned, secret-free snapshot: id, name, launch `location` (module or command form), the `describeCli(build)` surface, actor, timestamps.
- Two dependency-free reference backends, mirroring the state layer:
  - `FileSystemBuildRegistry` — one `<id>.json` under `.zuke/builds` (a sibling of `.zuke/runs`, never colliding), `O_EXCL` marker + atomic rename CAS.
  - `HttpBuildRegistry` — `GET/PUT/DELETE /builds/:id`, `GET /builds`, `ETag`/`If-Match` CAS, `{ url, token?, fetch? }`.
- **`zuke register`** — idempotent CAS write of this build's descriptor; `--json` prints it. Resolution precedence: explicit → `Build.registry()` override → `ZUKE_REGISTRY_URL`/`_TOKEN` or `ZUKE_REGISTRY_DIR` → default `.zuke/builds`.
- **`Build.registry()`** override, mirroring `stateStore()`.

## Secrets

Descriptors are secret-free by construction — built from `describeCli(build)`, which emits only structural metadata; `register` resolves no parameter values. Two extra guards (added in the adversarial pass below): a secret parameter's declared `.options(...)` values are omitted, and credentials embedded in a remote module URL are stripped from `location.module`.

## Adversarial review

An adversarial pass attacked path-safety, secret leakage, CAS/concurrency, crash/parse hardening, and CLI wiring; each finding was verified against the real code. Three confirmed issues were fixed here, each with a regression test:

- secret parameter `.options(...)` values were flowing into `describeCli` output (and thus the descriptor, `--list --json`, and MCP) — now omitted at the source;
- a credentialed remote module URL was persisted verbatim — userinfo is now stripped;
- `HttpBuildRegistry.listBuilds` parsed an untrusted body with an unguarded `JSON.parse` — now a friendly error.

Two low-severity findings were assessed and deliberately not changed: a defensive "no registry configured" guard reachable only via the `registry: false` API (tested, correct), and `register --list` being shadowed by the global `--list` (pre-existing and consistent with `runs`/`cancel`/`resume`; the root-cause fix is cross-cutting and out of scope).

## Tests & docs

- Unit (descriptor round-trip + hardened parse, FS CAS/deregister/listBuilds, HTTP verbs, resolve precedence, register command), integration (`zuke register` through `main()`), and a cross-process e2e (two real processes race a register — no torn write).
- `docs/registry.md` (new), the `/builds` section of `docs/state-api.md`, the skill cheatsheet, and regenerated `llms.txt` / `llms-full.txt` / README.

Full `deno task ci` green (coverage 95.2% branches), `apiDocsCheck` and `generate-ci --check` clean, `deno doc --lint` clean.

Dynamic MCP discovery over this registry is **PR2**.

### DIFF
--- a/docs/registry.md
+++ b/docs/registry.md
@@ -1,0 +1,125 @@
+# The build registry
+
+The **build registry** is a catalog of the pipelines (builds) that exist and
+where they live. Where [durable run state](./state.md) records *runs*, the
+registry records *builds*: each build registers a small **descriptor** — its id,
+its CLI surface (targets, parameters, commands), and how to launch it — into a
+pluggable store.
+
+Its purpose is agentic discovery driven by the store: an MCP server pointed at
+the registry can expose newly-registered pipelines as tools **without a redeploy
+or new instructions**. A build that registers itself for the first time becomes
+discoverable through an already-running server. (That dynamic `zuke mcp`
+integration is the next milestone; this page covers the registry and
+`zuke register` that back it.)
+
+The registry is a **separate concern** from the run [`StateStore`](./state.md) —
+a run history and a build catalog are different things — but it is configured the
+same way and, over HTTP, rides the same [REST contract](./state-api.md#build-catalog-builds)
+with a `/builds` collection beside `/runs`, so one service can host both.
+
+## Quick start
+
+```sh
+# Record this build in the registry (writes .zuke/builds/<id>.json by default).
+deno run -A zuke.ts register
+
+# Print the descriptor that was written.
+deno run -A zuke.ts register --json
+```
+
+Registration is **idempotent**: re-running `register` refreshes the descriptor
+(and its `updatedAt`) while preserving the original `createdAt`. Concurrent
+registrations converge on one record via compare-and-swap.
+
+## The build descriptor
+
+A descriptor is a versioned JSON snapshot of one build. It carries only static,
+structural metadata — never parameter *values* — so, like a run record, it
+excludes secrets by construction.
+
+| Field       | What it is                                                        |
+| ----------- | ---------------------------------------------------------------- |
+| `id`        | Stable build id (the build class name, unless overridden).       |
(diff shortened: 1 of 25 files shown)


totollygeek:
**Re: AI security review — `11f120x` (low): arbitrary registry URL can exfiltrate build metadata (`http_registry.ts:39`).**

Assessed against the code — this is by-design, not a defect, so no code change:

- The registry `url` is **trusted operator configuration** (`ZUKE_REGISTRY_URL` / a `Build.registry()` override), never attacker-controllable input. There is no lower-privilege path that can inject it, so there's no exfiltration primitive an attacker gains here.
- It is the **same posture as the already-shipped `HttpStateStore` and `HttpCacheStore`**, which send run records / cache tarballs to their configured host on exactly the same basis.
- It is **already documented** in the class's `**Security.**` JSDoc (`http_registry.ts:41–44`): "The \`url\` and \`token\` are *trusted configuration* … point it only at a service you control."

Constraining the URL (e.g. to loopback/an allow-list) would defeat the pluggable-backend design this milestone exists to provide — consumers host their own catalog service. Treating as a known, documented residual.

PR #188, opened 2026-07-19
feat(core): registry-backed dynamic MCP tool discovery

totollygeek:
## What

Milestone **M11 PR2** (the second half of M11) — **registry-backed dynamic MCP discovery**. `zuke mcp --registry` serves the whole [build registry](./docs/registry.md) instead of the single build its process was launched with, and **re-reads it on every `tools/list` and `tools/call`**. A pipeline registered by another process (`zuke register`, PR1) appears as a tool in an already-running server with **no restart** — the M11 acceptance.

## How

A registered build has no live instance in the server — only a descriptor — so a run tool cannot `execute()` in-process. It **spawns the descriptor's launch location** (`deno run <module> <target>`, or an explicit command) through an injectable runner and returns the captured output. This is the plan's resolved crux: the MCP host runs the location/command itself.

- **Discovery:** `list_builds` (catalog) and `describe_build {build}` read tools; each registered target becomes a `run:<buildId>:<target>` tool.
- **Execution** is off unless `--allow-run` and rides the **M5 authz tiers**, keyed on the qualified `<buildId>:<target>` name — allow-list (`--allow-run=Api:*`), operator token (`--protect=Api:deploy`), confirmation (`--confirm-destructive`). Every mutating or denied call is **audited**.
- **Scope:** a run tool takes only `dryRun`/`confirm`/`operatorToken` — the spawned build resolves its own parameters from the server environment. Per-parameter passthrough (which needs a secret-safe cross-spawn contract) is a documented follow-up. The single-build `McpServer` is **untouched** (zero regression).

## Adversarial review

A security pass (5 dimensions, each finding verified against the code) confirmed and fixed three issues, each with a regression test:

- **DoS:** `tools/list` did live registry I/O unguarded, so a transient HTTP-backend fault or one malformed descriptor would reject out of the stdio read loop and kill the server for all clients. Now guarded like `tools/call` — a registry hiccup returns a JSON-RPC error and the server keeps serving.
- **Confirmation bypass:** `dryRun:true` skipped `--confirm-destructive`. Sound for the in-process server (which enforces dry-run structurally), but the registry server only *appends* `--dry-run` and spawns — a command-wrapper that ignores it would execute for real. Confirmation is now required before **any** spawn, dry run included.
- **Secret leak:** a spawned build inherited the server's full environment, including the operator and HTTP bearer tokens. Those authorization secrets are now stripped from the child environment (the rest is preserved so the build still resolves its params).

Two candidates were refuted: argv ordering is safe (the target follows the module, so it's a script argument, never a `deno` runtime flag), and run output is already redacted by the spawned build's own reporter.

## Tests & docs

Unit (`registry_server_test.ts`: live discovery, spawn via a fake runner, the authz tiers, audit, the three fixes, the real `defaultRegistryRunner`), and a cross-process **e2e** that starts the server over an empty registry, registers a build from another process, and runs it — asserting the tool appears live and its spawned output comes back. `docs/mcp.md` gains a Registry-mode section; `docs/registry.md` and the cheatsheet updated.

Full `deno task ci` green (coverage 95.1% branches), `apiDocsCheck` and `generate-ci --check` clean.

Completes **M11**. Next: **M8** (timezone-aware schedules in `cicd()`).

### DIFF
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -162,6 +162,40 @@ The trail lives in a store-level record; read it with `zuke runs show mcp-audit`
 never influences authorization. On a shared HTTP endpoint it reflects the most
 recent client to connect, so set `--actor` for authoritative attribution there.
 
+## Registry mode (dynamic discovery)
+
+By default `zuke mcp` serves the single build its process was launched with. With
+`--registry` it instead serves the [build registry](./registry.md) — the catalog
+`zuke register` writes to — and **re-reads it on every `tools/list` and
+`tools/call`**. So a pipeline registered by another process appears as a tool in
+an already-running server with **no restart**:
+
+```sh
+# Serve every registered pipeline, execution enabled:
+zuke mcp --registry --allow-run
+```
+
+- **Discovery.** `list_builds` returns the catalog; `describe_build` (with a
+  `build` id) returns one build's surface. Each registered target is exposed as a
+  `run:<buildId>:<target>` tool, re-read live.
+- **Execution is a spawn.** A registered build has no live instance in the
+  server, so a run tool **spawns the build's registered launch location** (the
+  `deno run <module> <target>` `zuke register` recorded, or an explicit command)
+  and returns its captured output. This is code execution, so it is off unless
+  `--allow-run`, and it honours the same [authorization](#authorization) tiers —
+  the allow-list and `--protect` globs match the **qualified** `<buildId>:<target>`
+  name (e.g. `--allow-run=Api:*`, `--protect=Api:deploy`). Every mutating or
+  denied call is [audited](#audit-log).
+- **Scope.** A run tool takes no per-parameter inputs yet — only `dryRun`,
+  `confirm`, and `operatorToken`; the spawned build resolves its own parameters
+  from the server's environment. Passing parameters across the spawn boundary
+  (which needs a secret-safe contract) is a follow-up. Because a descriptor does
+  not record whether a target is read-only, every registry run tool is treated as
+  destructive.
+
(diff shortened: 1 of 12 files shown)


totollygeek:
**Re: AI security review (2 findings).** Both assessed against the code; one doc improvement pushed (`a3f44ca`), no behaviour change needed.

**`1pls2i3` — registry mode exposes arbitrary registered commands as tools (medium).** This is the feature by design, at the documented trust level: a descriptor's launch `command`/`module` is **trusted configuration** — only a trusted `zuke register` (or a consumer-hosted registry) can write it, and anyone who can write the registry can already run code. It rides entirely on the M5 gates: execution is **off unless `--allow-run`**, tools are hidden unless allow-listed, protected targets need an operator token, `--confirm-destructive` gates every spawn, and every mutating/denied call is audited. Documented in `registry_server.ts` (module JSDoc) and `docs/mcp.md` (Registry mode). Treating as a known, intentional, gated code-execution surface.

**`1gd8j4` — spawned targets inherit the server env, may leak non-auth secrets (medium).** The two MCP **authorization** secrets — `ZUKE_OPERATOR_TOKEN` and `ZUKE_MCP_TOKEN` — are already stripped from the child env (this PR's adversarial fix). The build *does* still inherit the rest of the environment, because that is how it resolves its own parameters (`docs/mcp.md`: "the spawned build resolves its own parameters from the server's environment") — fully isolating it would break that. Narrowing further isn't feasible generically: the descriptor doesn't carry the build's env→param mapping, so the server can't know which vars are "the build's". The right control is operational: run a registry-backed server with only the environment the registered pipelines should see — which I've now made explicit in `docs/mcp.md` (new **Environment** note under Registry mode). Residual accepted at the documented trust level.

PR #190, opened 2026-07-19
feat(core): timezone-aware schedules in cicd()

totollygeek:
## What

Milestone **M8** — timezone-aware cron schedules in `cicd()`, so a pipeline's schedule lives in code and you can delete the external Cloud-Scheduler-plus-webhook dance.

```ts
ci = cicd({
  provider: "github",
  pipeline: { triggers: { schedule: [{ cron: "30 9,13,15 * * 1-4", tz: "Europe/Sofia" }] } },
});
```

## How

GitHub and Azure cron are **UTC-only**, so a `tz` entry is compiled to the UTC cron(s) that fire at the intended *local* wall-clock time:

- A zone that changes offset across the year contributes **one UTC cron per distinct offset**. On GitHub, a generated **guard job** (`zuke-schedule-guard`) that every other job `needs`/`if`s on lets through only the firing whose wall-clock in the zone matches — so the "wrong" offset's firing is skipped. A fixed-offset zone (or plain UTC) is a single cron, no guard.
- Grammar is a numeric subset (single values, comma lists, `a-b` ranges, slash steps) with whole-hour offsets; named fields, fractional-hour zones, and day-constrained schedules that would cross midnight raise a friendly error telling you to write the UTC cron directly.
- Offsets are sampled across a **pinned reference year**, so `generate-ci --check` output never churns with the calendar.
- **Azure** emits native `schedules:` for UTC/fixed-offset zones (a DST zone errors — the guard is GitHub-only). **GitLab/Bitbucket** configure schedules in their UI, so a `schedule` trigger is ignored for them.

## Adversarial review

A correctness pass (4 dimensions, each finding verified against the code) confirmed and fixed five issues, each with a regression test:

- an open-ended slash step (`9/4`) collapsed to a single value → now expands to the field max, as in Vixie cron;
- two-sample (Jan/Jul) DST detection missed a sub-seasonal offset window (Ramadan zones like `Africa/Casablanca`) → the whole reference year is now sampled, and the cron-per-offset + wall-clock guard generalizes to any offset count;
- the guard **AND**ed day-of-month and day-of-week, but cron **OR**s them when both are restricted → fixed;
- a job colliding with the generated guard id was silently overwritten → friendly error;
- the once-a-year DST fall-back double-fire is documented as a known ceiling (make schedules idempotent).

## Tests & docs

Unit (`ci_schedule_test.ts`: passthrough, fixed/DST shifting, ranges/steps, every error boundary, the guard shell, determinism, and each fix), provider-render tests in `ci_test.ts`, and an integration test driving `generate-ci` end-to-end (proving determinism via `--check`). Docs: an authoring "Scheduled runs" section and a cheatsheet note.

Full `deno task ci` green, `apiDocsCheck` and `generate-ci --check` clean.

Completes the P1 Track-A items. Next: Track B (T1–T4).


totollygeek:
**Re: AI security review — shell-injection surface in the guard step from an unescaped timezone (`ci_schedule.ts`, low).**

Fixed in `266fe15` (defense-in-depth) with a regression test.

Assessment: the `tz` is interpolated into `TZ='<tz>'` in the generated guard. It was **not reachable in production** — `github()` builds `on.schedule` via `flatMap(utcCronsFor)`, which validates *every* entry's `tz` through `Intl.DateTimeFormat` (throws `unknown timezone` on anything non-IANA) **before** the guard is built, and a canonical IANA name (`America/Port-au-Prince`, `Etc/GMT+5`, …) contains only `[A-Za-z0-9_+/-]`, which is inert inside single quotes. So a quote-carrying string fails generation, never reaching the guard.

But `guardShell` is exported and shouldn't depend on its caller, so it now validates each `tz` itself (rejecting anything Intl won't accept) before interpolating — making the exported function self-defending. The new test asserts `{ tz: "Europe/Sofia'; rm -rf / #" }` throws `unknown timezone` rather than emitting an injectable script.

PR #194, opened 2026-07-19
feat(core): registry run-tools accept build parameters

totollygeek:
## What

Round 2 / M12 (P0 #1). A registry MCP `run:<build>:<target>` tool now exposes the registered build's **declared parameters** as its input schema and **forwards** supplied values to the spawned build — the single hardest blocker for the shared control plane, since the requester's deploy build is parameterized (`repos`, `sit`, `skipE2e`).

Before: a registry run tool took only `dryRun`/`confirm`/`operatorToken`; the spawned build could only resolve parameters from the server's environment.

## Design

The in-process `McpServer` already did per-parameter schemas; the registry path lacked it only because the descriptor dropped the metadata. So:

- **Descriptor** (`CliParameterInfo`) gains `name` (property name), `kind`, and `default`. Parsing is tolerant of a pre-M12 descriptor: absent `kind` derives from the `boolean` flag, absent `name` falls back to `flag`.
- **Input schema** is keyed by the parameter's **property name** (`skipE2e`, matching the in-process server and the requirement's example), with kind, description, enum, and default.
- **Validation before spawn**: array-vs-scalar, number, boolean, and enum are checked against the descriptor; a bare string for an array, a non-numeric number, an out-of-set enum, or an unknown parameter is a clean tool error — the subprocess never starts.
- **Forwarding**: validated values become `--flag=value` child arguments (the CLI already parses `--flag=false` for booleans, so no `cli.ts` change).
- **Secrets**: `.secret()` parameters are omitted from the descriptor entirely (`describeCli(build, { omitSecrets: true })`, written by `zuke register`), so a secret can neither be requested nor forwarded — it's rejected as an unknown parameter. The child resolves it from its own env / `.from()` source.
- **Reserved names**: `discoverParameters` rejects a parameter named `dryRun`/`confirm`/`operatorToken` — which also fixes a latent **silent collision** in the in-process server (it would overwrite a same-named parameter with the control key).

## Tests

- **Unit** — `registry_server_test.ts` (schema exposure, forwarding, type/enum/number/unknown rejection before spawn, boolean `false`, control-key coexistence), `describe_test.ts` (name/kind/default, `omitSecrets`), `params_test.ts` (reserved-name guard, `default_`), `registry_test.ts` (tolerant descriptor parse + round-trip).
- **Integration** — `tests/integration/registry_params_test.ts`: the real `register` → `FileSystemBuildRegistry` (JSON round-trip) → `RegistryMcpServer` pipeline; asserts the run tool exposes params, forwards them, and excludes the secret.
- **E2E** — `registry_mcp_e2e.ts` + fixture: a real cross-process spawn binds `repos`/`skipE2e` into the child (which echoes them), and a type mismatch is rejected without spawning.

## Adversarial review

Fanned out attackers across secret-boundary, argv-injection, validation-bypass, schema-evolution, and the discovery guard; verified each finding refute-by-default. Two confirmed defects fixed here with regressions:

- **Audit-log leak (medium):** a value mistakenly supplied under a secret parameter's name (unknown, since secrets are absent from the descriptor) was recorded verbatim in the durable audit trail. Now only recognized (non-secret) argument values are recorded; unknown keys keep their name but their value is elided.
- **Malformed schema from an untrusted descriptor (low):** a `kind:number` parameter with a non-numeric `default` or a string `enum` produced an invalid JSON Schema. Both are now dropped so the advertised schema stays well-formed.

Refuted / out of scope: the inherent `.array()` comma-delimiter contract (identical on the CLI), and a self-contradictory `.options("a,b").array()` declaration that fails cleanly.

## Docs

`docs/mcp.md` (registry run-tool parameters, secret exclusion, audit elision); `llms-full.txt`/README regenerated; `deno doc --lint` clean; `deno task ci` green.

### DIFF
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -187,19 +187,31 @@ zuke mcp --registry --allow-run
   the allow-list and `--protect` globs match the **qualified**
   `<buildId>:<target>` name (e.g. `--allow-run=Api:*`, `--protect=Api:deploy`).
   Every mutating or denied call is [audited](#audit-log).
-- **Scope.** A run tool takes no per-parameter inputs yet — only `dryRun`,
-  `confirm`, and `operatorToken`; the spawned build resolves its own parameters
-  from the server's environment. Passing parameters across the spawn boundary
-  (which needs a secret-safe contract) is a follow-up. Because a descriptor does
-  not record whether a target is read-only, every registry run tool is treated
-  as destructive.
+- **Parameters.** A run tool exposes the registered build's declared parameters
+  as its input schema — keyed by the parameter's property name (e.g. `skipE2e`),
+  with the kind, description, enum, and default from the descriptor. Supplied
+  values are validated against their kinds **before** the build spawns (a type
+  mismatch — a bare string where an array is required, a non-numeric number, an
+  out-of-set enum, or an unknown parameter — is a clean tool error, not a failed
+  subprocess), then forwarded to the child as `--flag=value` arguments alongside
+  the target. A value still set in the server's environment applies when the
+  call omits it. Because a descriptor does not record whether a target is
+  read-only, every registry run tool is treated as destructive.
+- **Secrets never cross the boundary.** `.secret()` parameters are omitted from
+  the descriptor entirely (`zuke register` writes the secret-free surface), so a
+  secret can neither be requested nor forwarded — it is rejected as an unknown
+  parameter if a client tries. The spawned build resolves a secret from its own
+  environment / `.from()` source instead. In the [audit log](#audit-log) only a
+  recognised parameter's value is recorded; any unknown argument keeps its name
+  but its value is elided, so a value mistakenly supplied under a secret's name
+  is never written to the durable trail.
 - **Environment.** A spawned build inherits the server's environment (that is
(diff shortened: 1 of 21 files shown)


totollygeek:
Re: security review finding `e1b71m` (descriptor defaults surfaced into MCP schemas) — assessed as by-design, non-exploitable, no change:

- A descriptor `default` is **advisory schema metadata only**. It is never forwarded to the spawned build: `#run` builds the child argv from `validateParamArgs`, which processes only *supplied* arguments; the child resolves its own defaults from its own build definition, not from the descriptor.
- The registry-run surface already treats the descriptor's **launch location** as trusted code execution (gated by `--allow-run` + allow-list + operator token). A misleading advisory default is negligible next to arbitrary-code execution that is already the accepted trust boundary.
- This PR also hardens the schema against a malformed/untrusted descriptor: `defaultToJson`/`schemaForParam` drop a `default`/`enum` that doesn't match the declared kind, so the advertised schema stays well-formed (regression test: "a malformed default/enum from an untrusted descriptor is dropped from the schema").

Leaving as-is; the finding is correctly rated low/non-exploitable.

PR #198, opened 2026-07-19
feat(core): state/registry conformance kit and protocol version

totollygeek:
## What

Round 2 / M16 (P1 #5 + P2 #9). A hosted state-api backend must implement the same CAS, listing, and TTL-lock semantics the filesystem backend does — the exactly-once resume, lock takeover, and one-writer-wins guarantees the core relies on ride on them. The requester is building a Deno+Postgres backend and chose option (b): a conformance kit over a reference server.

## Design

- **Conformance kit** — a new `@zuke/core/conformance` entrypoint with store-agnostic scenarios: CAS stale-write rejection, concurrent-writers-one-wins, `listRuns`/`listBuilds` filter+sort, run-event round-trip, and the full lock lifecycle (acquire, holder-named conflict, TTL takeover, concurrent-one-wins). Zuke's own test lane runs them against the FS store (no coverage loss); a backend author runs them against a live URL:

  ```sh
  deno run -A jsr:@zuke/core/conformance --url http://localhost:8080 [--token …]
  ```

  Exposed as `checkStateStore(factory)` / `checkBuildRegistry(factory)` (programmatic) and `runConformanceCli(args, deps)` (the CLI, injectable for tests). Every scenario uses fresh `crypto.randomUUID()` ids, so it is safe against a shared, persistent service.
- **Protocol version** — every HTTP client request stamps `x-zuke-state-protocol: 1` (via a shared `state/protocol.ts`, wired through both clients' `#headers()`/a new `#request()` wrapper). A response that declares a *different* version fails loudly rather than mis-parse; a server that omits the header is treated as compatible, so adoption is optional.
- **Docs** — `docs/state-api.md` documents the protocol version + the kit; `docs/state.md` adds an API-stability policy (the state surface is stable-with-deprecation-cycle; the `RunRecord` schema-evolution tolerance is a stated guarantee).

## Tests

`conformance_test.ts`: the kit passes against the FS state store and FS registry, fails loudly against a CAS-violating fake (CAS scenario flagged specifically), catches an order-violating registry, and the CLI runner exits 0/1 correctly. The protocol handshake rejects a mismatched-version server and tolerates an omitting one. The two HTTP-client header deep-equal asserts were updated for the new request header.

## Adversarial review

A 5-dimension workflow (false-pass, false-fail/flake, protocol, isolation, CLI/types) with refute-by-default verification found **4 confirmed**, all fixed with regressions:

- **Two HIGH false-flakes** (verified with a 400-iteration repro; the committed test already flaked 1/20): the CAS-stale and concurrent-writer scenarios false-failed a *content-hash* backend (the FS store!) when create and update landed in the same millisecond — identical content yielded an identical version, so the "stale" write correctly succeeded. Fixed: each write now carries distinct content, so the version genuinely advances on any versioning scheme.
- **listBuilds ordering was uncertified** (medium) and the **`since`/`name` filters were never exercised** (low) — a filter/order-violating backend passed. Fixed: the registry listing scenario now checks newest-first ordering and both filters; a regression test proves the kit catches an order-violating registry. The live-lock scenarios were also decoupled from the short takeover TTL (a 1-hour TTL) so a slow backend never races the expiry.

## Gates

`deno task ci` green (5× flake-check on the kit clean); `apiDocsCheck`, `generate-ci --check`, and `deno doc --lint` over all 5 entrypoints clean. Skill cheatsheet updated with the kit + protocol note.

### DIFF
--- a/docs/state-api.md
+++ b/docs/state-api.md
@@ -22,6 +22,13 @@ The client is dependency-free and talks plain HTTP, so any stack can serve it.
   returns it as an `ETag` response header and honours it in `If-Match` /
   `If-None-Match` request headers. Any stable per-write token works (a row
   version, a content hash, a monotonic counter).
+- **Protocol version.** This document describes wire protocol **`1`**. Every
+  client request carries `x-zuke-state-protocol: 1`. A service **may** echo its
+  own version in that response header; when it does and the value differs, the
+  client fails loudly rather than risk a silent mis-parse. A service that omits
+  the header is treated as compatible, so adopting it is optional — but echoing
+  it lets a version bump be detected immediately. A breaking change to this
+  contract bumps the number.
 
 ## Endpoints
 
@@ -30,10 +37,10 @@ The client is dependency-free and talks plain HTTP, so any stack can serve it.
 Fetch one run.
 
 | Response | Meaning                                                         |
-| -------- | -------------------------------------------------------------- |
-| `200`    | Body is the run record; **`ETag` header is required**.         |
+| -------- | --------------------------------------------------------------- |
+| `200`    | Body is the run record; **`ETag` header is required**.          |
 | `404`    | No such run (the client treats this as "not found", not error). |
-| other    | The client raises an error.                                    |
+| other    | The client raises an error.                                     |
 
 The client **requires** the `ETag` header on a `200`; omitting it is an error.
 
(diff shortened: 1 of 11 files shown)


totollygeek:
Re: security finding (conformance CLI accepts arbitrary URLs without scheme/host validation) — assessed as by-design, non-exploitable, no change:

- The conformance kit is a **developer testing tool**: the operator runs it against **their own** backend, and `--url` is the thing they are deliberately pointing it at (from their own command line). There is no untrusted input crossing a trust boundary and no privilege escalation — it is not an SSRF surface.
- The `--url` has exactly the same trust posture as `HttpStateStore({ url })` / `HttpBuildRegistry({ url })`, which `docs/state-api.md` already documents as *trusted configuration* ("point it only at a service you control").
- Restricting the scheme/host would reject the `http://localhost:8080` dev usage the milestone's acceptance explicitly calls for.

Leaving as-is; the finding is correctly rated low/non-exploitable.

PR #199, opened 2026-07-19
feat(core): fan-out items run their own onCancel compensation

totollygeek:
## M15 — Fan-out per-item compensation on cancel (Round 2, P1 #4)

On cancel, a `.forEach()` fan-out target's per-item sub-targets now run **their
own** `.onCancel()` compensation — the requester's most common compensation
scenario (their deploys are fan-out-first, each item leasing per-repo
resources).

### What changed

- **Per-item compensation.** The cancel walk re-materialises each fan-out parent
  (`spec.materialize()`), matches each item's stage sub-targets by name
  (`parent[item].stage`) against the run record, and runs the compensation of
  every item that had **succeeded or was still in-flight**, each with its own
  item-scoped `TargetContext` and persisted `ctx.state`. Reverse order (later
  items/stages unwind first); a throwing item compensation is recorded and the
  walk continues (cleanup stays maximal).
- **Per-item audit events.** Each compensation lands its own `compensate`
  `RunEvent` naming the runtime sub-target (e.g. `deployBatch[repo-a].deploy`),
  alongside the existing summary `cancel` event, on both `zuke cancel` and the
  in-process (Ctrl-C) path.
- **Nested fan-out** is handled by recursion; the unmatched-item scan runs once
  against the full descendant-name set, so nested items don't false-warn.
- Retires the previous "sub-target `.onCancel()` is not run" limitation from
  `docs/orchestration.md`, the JSDoc, and the skill references.

### Acceptance

Batch of 3, cancel after item 1 finished and item 2 mid-flight → items 1 and 2
compensate with their own contexts, item 3 never does; the event log shows
per-item `compensate` entries; a throwing compensation on item 1 doesn't stop
item 2's. Covered by unit tests (`cancel_test.ts`) and an end-to-end CLI
integration test (`tests/integration/cancel_test.ts`).

### Adversarial review

A fan-out adversarial pass (5 attack dimensions → refute-by-default verify)
surfaced **4 confirmed** findings; each verified against the real code:

- **[high] Unguarded `spec.materialize()` in `runForEachTarget`** — a throwing
  item-list thunk/factory escaped as an uncaught rejection, crashing the run and
  leaving the record non-terminal. **Fixed:** wrapped so it fails the target
  cleanly on both the sequential and parallel dispatch paths; regression tests
  assert a failed result and a `failed` record.
- **[low] Nested fan-out** dropped grandchildren compensations and emitted a
  spurious "no matching re-materialised item" warning. **Fixed:** recurse with a
  shared name set and a single top-level unmatched-item scan; regression test
  asserts the grandchild compensates with no false warning.
- **[low] Resolution failures** (a throwing `.onCancel()` thunk / a throwing
  item list) inflated the summary's failed count but produced no per-item
  `compensate` event. **Fixed:** they are now recorded as attempts, so per-item
  events agree with the summary; regression test asserts the error event.
- **[medium] Out-of-process in-flight window** — `zuke cancel` compensates a
  `running` item from the record snapshot, and the owning process aborts a live
  body only on its next state write, so the two can briefly overlap. This is
  inherent to the requirement's "compensate in-flight items" design (the
  acceptance has no live process), and fully closing it needs prompt
  abort-propagation in the executor — a cancellation-hardening follow-up beyond
  this milestone. **Dispositioned:** documented with the `ctx.state.set(...)` /
  `.lock()` mitigations that close the window, plus a ceiling comment naming the
  upgrade path. In-process cancel has no such window (bodies settle before the
  walk).

### Gate

clean; skill references synced.

### DIFF
--- a/docs/orchestration.md
+++ b/docs/orchestration.md
@@ -239,11 +239,36 @@ to whoever cancelled it, so `zuke runs show <id>` shows what was unwound.
 Cancellation needs a state store, so a build that uses `.onCancel()` turns on
 the `.zuke/runs` filesystem store by default (like `.lock()` and `.waitsFor()`).
 
-> **Limitation:** the cancel walk considers only the static plan, so an
-> `.onCancel()` on a [`.forEach()`](#fan-out-over-a-list--foreach)
-> **sub-target** is not run (sub-targets are materialised at run time and don't
-> exist at cancel time). Put a compensation on an ordinary target, or on the
-> parent fan-out target itself.
+An `.onCancel()` on a [`.forEach()`](#fan-out-over-a-list--foreach) **sub-target**
+runs too. Cancel re-materialises the fan-out, matches each item's sub-targets by
+name against the record, and runs the compensation of every item that had
+**succeeded — or was still in-flight when the cancel landed** (a mid-deploy item
+has partial work to undo). Items unwind before the parent's own compensation, in
+reverse order; a throwing item compensation is recorded and the walk continues,
+exactly as for an ordinary target. Each item's cleanup reads that item's own
+persisted state (`ctx.state`), and lands its own `compensate` entry in the
+[audit trail](./state.md) (naming the runtime sub-target, e.g.
+`deployBatch[repo-a].deploy`) alongside the summary.
+
+Nested fan-out works too: a `.forEach()` stage that is itself a `.forEach()` has
+its inner items' `.onCancel()` run, matched by the same nested
+`parent[item].stage[inner].stage` names.
+
+> **Caveat:** for per-item compensation to find its items, the `.forEach()`
+> item list must be **deterministic** — cancel re-evaluates it (from the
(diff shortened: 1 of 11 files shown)


totollygeek:
Re: the AI-review low finding (audit event duplication on replay, `cancel.ts:766`) — verified as a false positive. `finalizeCancelled` re-reads a **fresh** record via `store.getRun(id)` at the top of every retry iteration, clones it, appends the events once, and `putRun`s with that read's version. On a CAS conflict the events never landed (the put failed), so the next iteration appends to a fresh copy that doesn't contain them; on success it returns immediately. Net result: the compensate + cancel events are appended exactly once. The early `status === "cancelled"` guard also stops a second canceller from re-appending. This is the same append pattern the pre-existing summary `cancelEvent` already used — M15 only adds the per-item `compensate` events alongside it, with no change to the CAS retry structure. No change needed.

PR #201, opened 2026-07-19
feat(core): run-record retention and runs prune

totollygeek:
## M17 — Run-record retention (Round 2, P1 #6)

Answers **who owns run-record retention** — then ships the CLI that answer
demands. The requester needs the ownership answer before freezing their backend
schema.

### What changed

- **`StateStore.deleteRun(id)`** — added to the interface. The filesystem backend
  removes the run's file (under its lock; a missing run is a no-op); the HTTP
  backend maps it to `DELETE /runs/:id` (`404` is not an error), a
  **documented-optional** endpoint a hosted store implements only if it wants the
  CLI to prune it.
- **`RunQuery.limit`** — return at most the newest N. Honoured server-side by both
  backends (FS slices after the newest-first sort; HTTP sends `?limit=`), and
  threaded through `zuke runs list --limit` and the `list_runs` MCP tool.
- **`zuke runs prune --keep <age> --keep-last <n> [--dry-run]`** — deletes only
  **terminal** runs (`succeeded`/`failed`/`cancelled`) matching **neither** rule:
  older than the keep window **and** beyond the newest keep-last. A **non-terminal**
  run (`suspended`/`running`/`cancelling`) is **never** pruned — a run suspended
  for days awaiting a human is the point of the system. At least one rule is
  required, so a bare `prune` can't wipe the store; `--dry-run` previews.
- **The ownership answer** (`docs/state-api.md`, `docs/state.md`): the FS store
  owns pruning via the CLI; for the HTTP backend retention is the **server's** job
  (a TTL/scheduled sweep), with `limit` and the optional `DELETE` the only
  contract additions. The conformance kit now certifies the `limit` semantic.

### Acceptance

Prune on a mixed store removes old terminal records, keeps the newest N and
everything non-terminal; `RunQuery.limit` honoured by both backends; the ownership
answer documented. Covered by unit (`runs_test.ts`, `state_test.ts` for both
backends + the conformance limit scenario) and end-to-end CLI integration
(`tests/integration/state_test.ts`).

### Adversarial review

Five attack dimensions (prune-safety, delete-race, limit, cli-parse,
contract-ripple) → refute-by-default verify. **1 confirmed** (the other four
dimensions found nothing):

- **[low, high-confidence] Empty space-form `--keep ""` silently dropped.** From
  `--keep "$KEEP"` with `$KEEP` unset in CI, the `if (value)` guard treated `""`
  as absent, so `prune --keep "" --keep-last 5` proceeded with **only** keep-last
  — deleting recent terminal runs the intended age window would have kept, exit 0.
  Inconsistent with the `--keep=` form, which already errors on empty. **Fixed:**
  the three new flags capture an explicit empty value (`if (value !== undefined)`)
  so it is validated and rejected (exit 1), matching the inline form; regression
  tests on both the parse and the `main` error path.

### Gate

clean; skill references synced.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -18,8 +18,9 @@
 | `zuke mcp [--allow-run[=<globs>]] [--http <host:port>]` | Run an MCP server over the build for AI agents, on stdio or HTTP ([details](./mcp.md)). |
 | `zuke resume <id> [--signal <n>] [--data <json>]`       | Resume a suspended run, optionally delivering a signal ([details](./orchestration.md)). |
 | `zuke resume --check [<id>]`                            | Re-check suspended runs (predicate waits, timeouts).                                    |
-| `zuke runs list [--status/-target/-since] [--json]`     | List persisted run records, newest first ([details](./state.md)).                       |
+| `zuke runs list [--status/-target/-since/-limit] [--json]` | List persisted run records, newest first ([details](./state.md)).                    |
 | `zuke runs show <id> [--json]`                          | Show one run's full per-target status and metadata.                                     |
+| `zuke runs prune [--keep <age>] [--keep-last <n>] [--dry-run]` | Delete old terminal run records; never touches non-terminal runs.                |
 | `zuke cancel <id> [--actor <name>]`                     | Cancel a run and run its compensations ([details](./orchestration.md#cancellation--compensation-oncancel)). |
 | `zuke --help` / `-h`                                    | Usage.                                                                                  |
 | `zuke` (no target)                                      | Run the `default` target if defined, else print `--list`.                               |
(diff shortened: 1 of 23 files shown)


totollygeek:
Both AI-review findings verified as false positives / by-design:

**1. HTTP prune authorization (medium, `http_store.ts:159`).** `deleteRun` sends the request through `this.#headers()` (line 164) — the *same* bearer-token auth as every other operation (`getRun`, `putRun`, `listRuns`, the lock ops). The client never performs authorization; it presents a credential, and the **server** terminates authn/authz. That is the documented trust posture of the entire HTTP store: the `url`/`token` are trusted configuration and `docs/state-api.md` states "terminate TLS and enforce authn/authz in front of the service." `DELETE /runs/:id` is no different from the already-mutating `PUT /runs/:id` — no new authorization gap is introduced, and it mirrors the existing `DELETE /locks/:key`.

**2. Unbounded --keep-last/--keep (low, `cli.ts:292`).** Both are validated (positive integer / valid duration; empty and non-positive are rejected with exit 1). A *large* value is the **conservative** direction — it keeps *more* and deletes *fewer* runs, so it cannot force a larger sweep; `selectRunsToPrune` stays O(n) regardless (a huge keep-last just makes every terminal run satisfy `index < keepLast`). "No-op retention" (keeping everything) is safe by construction, and prune is an explicit operator action against their own store, not an attacker-reachable surface. No change needed.

PR #202, opened 2026-07-19
feat(gh): created-window workflow correlation and marker fast-fail

totollygeek:
## M18 — `githubWorkflow` correlation without the marker (Round 2, P1 #7)

Unmodified target workflows (the ~100-repo long tail you don't own) can't echo
the correlation marker into their `run-name:`. This adds a **best-effort
fallback** and makes both modes **fail fast** instead of eating the whole
`.timeout()`.

### What changed

- **`.correlate("created-window")`** — snapshots the runs that already exist at
  dispatch, then claims the *new* `workflow_dispatch` run on the dispatch ref
  created just after (excluding pre-existing runs; **loud** on two fresh
  candidates in the window). Marker correlation stays the exact default.
- **`.discoveryTimeout(...)` fast-fail** — if no run is identified within a short
  window (default one minute), the gate fails with guidance
  (`does "<workflow>" echo "<marker>" into run-name:?`) rather than hanging until
  the wait's `.timeout()`. Anchored to the **persisted dispatch time**, so it
  holds across a suspend/resume, and it applies even when the correlation call
  keeps erroring (bad token / renamed workflow fails fast).
- `WorkflowRun` gains `createdAt`/`headBranch` (parsed from `created_at`/
  `head_branch`); `WorkflowCorrelationError` + `CorrelateMode` are exported.

### Acceptance

An unmodified workflow + the fallback resolves the single-dispatch case and
errors distinctly on two dispatches; marker mode against a non-echoing workflow
fails within ~a minute with the guidance message. Covered by unit tests
(`workflow_test.ts`) and an end-to-end CLI fast-fail across a resume
(`tests/integration/gh_workflow_test.ts`).

### Adversarial review

Five attack dimensions → refute-by-default verify. **3 confirmed** (all
high-confidence), each **fixed with a regression test**:

- **[medium] created-window could claim a foreign run.** A `workflow_dispatch`
  run someone else started on the same branch just before ours — visible before
  ours propagates — was claimed as ours (the 2+ ambiguity guard misses it when
  only one is visible). **Fixed:** snapshot the existing run ids at dispatch and
  exclude them. The narrow residual (a foreign dispatch landing after the
  snapshot yet visible before ours) is documented as the best-effort ceiling.
- **[low] Pre-M18 records re-stamped the discovery clock.** A record predating
  `dispatchedAt`, resumed under M18, recomputed the anchor every poll (the
  `?? now()` fallback wasn't persisted on the not-found path), so the window
  never elapsed. **Fixed:** backfill and persist the anchor once on first
  observation.
- **[medium] Fast-fail bypassed when the correlation API threw.** A permanent
  correlation error (missing `actions:read`, a renamed workflow, a long outage)
  jumped to the catch and returned "suspended", never reaching the deadline —
  so it ate the full `.timeout()`. **Fixed:** the discovery deadline is enforced
  on the error path too, while the run is still uncorrelated.

### Gate

clean; gh unit + integration + e2e green; skills synced.

### DIFF
--- a/docs/orchestration.md
+++ b/docs/orchestration.md
@@ -92,10 +92,19 @@ class Release extends Build {
   correlation marker in the gate's durable state, and suspends. Each
   `resume --check` polls the run; a resume in a **different process** never
   re-dispatches, because the marker persisted with the run.
-- **Correlation.** `workflow_dispatch` returns no run id, so the trigger passes a
-  marker input (default `zuke_marker`) and matches it against the run's display
-  title — the dispatched workflow must echo it into `run-name`:
-  `run-name: ${{ inputs.zuke_marker }}`.
+- **Correlation.** `workflow_dispatch` returns no run id, so by default the
+  trigger passes a marker input (default `zuke_marker`) and matches it against
+  the run's display title — the dispatched workflow must echo it into `run-name`:
+  `run-name: ${{ inputs.zuke_marker }}`. A workflow you can't modify (the long
+  tail of repos you don't own) correlates **best-effort** with
+  `.correlate("created-window")`: the trigger claims the `workflow_dispatch` run
+  on the dispatch ref created just after dispatch, and fails loudly if two
+  candidates share the window.
+- **Fast-fail.** If no run is identified within a short discovery window
+  (`.discoveryTimeout(...)`, default one minute), the gate fails with guidance
+  instead of eating the whole `.timeout()` — so a workflow that silently never
+  echoes the marker surfaces in ~a minute. The deadline is measured from the
+  persisted dispatch time, so it holds across a suspend/resume.
 - **Result.** On completion the per-job conclusions (`{ passed, jobs: [{ name,
   conclusion, url }] }`) are written to the gate target's state; a dependent
   reads them with `readWorkflowResult(ctx.stateOf("<gate>"))` and branches on a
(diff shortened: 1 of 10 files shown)


totollygeek:
Re: the AI-review low finding (fixed 30s created-window skew, `workflow.ts:387`) — dispositioned as a by-design best-effort tradeoff, no change needed:

- **The failure mode is loud, not silent.** If a legitimate run is created more than 30s *before* our dispatch timestamp (a host clock skewed >30s vs GitHub), the run is excluded → no correlation → the **discovery fast-fail** fires within the window with actionable guidance. The operator sees a clear error (and a >30s host clock skew is a misconfiguration they'd want surfaced), not a hang and not a wrong claim.
- **30s deliberately errs toward false-negative-that-fails-loud over false-positive-that-ships-wrong.** A *larger* skew widens the window in which a foreign `workflow_dispatch` run on the same ref could be claimed as ours — the more dangerous direction for a release gate. The plan specifies "a small skew allowance" for exactly this reason, and the new dispatch-time baseline already excludes pre-existing runs.
- **The exact alternative is one call away.** `created-window` is documented as best-effort; when correlation exactness matters, the default marker mode (echo `zuke_marker` into `run-name:`) has no skew dependency at all. Well-synced hosts (NTP) skew <1s, so 30s is already generous.

Making the skew configurable is a reasonable future enhancement, but not warranted for a low finding whose worst case is a loud, guided fast-fail.

PR #203, opened 2026-07-19
feat(core): lazy edge provider for batch ordering

totollygeek:
## M19 (P2 #10) — `build.orderWith()` lazy edge provider

The requester keeps ownership of their cross-repo `dependency-graph.json`; Zuke
gains only the **binding seam** to feed it into ordering at run time.

### What changed

- **`Build.orderWith(targets)`** — a **lazy, per-run** provider of soft ordering
  edges (`[before, after]`), merged with the synchronous `extraEdges`. Unlike
  `extraEdges` (declaration-time), it may be **`async`** and is resolved when a
  run plans, so it can *load* the ordering — read `dependency-graph.json`, hit an
  API. Shares `extraEdges`' rules (endpoints outside the run's set ignored,
  cycles reported). Defaults to `[]`, so every existing build is unchanged.
- Honoured by a run and by `zuke cancel`'s reverse-order compensation walk (via a
  new `resolveOrderingEdges` helper wired into both planning points), but — being
  resolved only when a run plans — **not** by the static `graph`/`--list` views
  nor by `cicd()`-generated CI (whose `needs:` mirrors hard `dependsOn`).

### Acceptance

An async `orderWith` imposes ordering through a real run; it merges with
`extraEdges` into one edge set; `zuke cancel` unwinds compensations in the
`orderWith` execution order. Covered by unit (`graph_test.ts`) + end-to-end CLI
integration (`tests/integration/orderwith_test.ts`).

### Adversarial review

Four attack dimensions → refute-by-default verify. **1 confirmed** (high
confidence, medium severity), **fixed with a regression test**:

- **A failing `orderWith` during cancel stranded the run.** Because the provider
  can be async and reach an external graph service, an *unreachable* service
  during `zuke cancel` made `resolveOrderingEdges` reject **after** the run was
  CAS'd to `cancelling` — so the compensation walk and the settle never ran, and
  a retry `cancel` took the "recover" path that **skips** compensations, losing
  the rollbacks permanently. **Fixed:** the cancel path now degrades a failed
  provider to the base topological order and still compensates and settles
  (the edges only affect *order*). On the run path — where ordering can be a
  correctness requirement — a failed provider now fails the build **cleanly**
  with a friendly message rather than crashing with an unhandled rejection or
  silently running in the base order; both paths have regression tests.

### Gate

`deno task ci` green (lines 98.2%, branches 95.1%); `build.ts` 100% covered;
`apiDocsCheck`, `generate-ci --check`, and `deno doc --lint` over all five core
entrypoints clean; skills synced.

*(M19 is a P2 batch — this is the `orderWith` slice. The gcloud/kubectl/GCS
wrapper slices remain, each its own small PR when demand warrants.)*

### DIFF
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -678,6 +678,25 @@ class Monorepo extends Build {
 }
 ```
 
+**Lazy ordering — `override orderWith(targets)`.** Same edges, but **async and
+resolved per run**: use it when the ordering must be *loaded* at run time — read
+the monorepo's `dependency-graph.json`, hit an API — rather than declared
+statically. Return the same `[before, after]` pairs; they merge with
+`extraEdges`, and share its rules (endpoints outside the run's set are ignored,
+cycles reported). Both are honoured by a run and by `zuke cancel`'s reverse-order
+compensation walk, but — being resolved only when a run plans — neither shows in
+the static `graph` / `--list` views nor in `cicd()`-generated CI.
+
+```ts
+override async orderWith(t: Map<string, Target>): Promise<OrderingEdge[]> {
+  const graph = await loadDependencyGraph(); // read it at run time
+  return graph.edges.flatMap(([before, after]) => {
+    const from = t.get(before), to = t.get(after);
+    return from && to ? [[from, to] as OrderingEdge] : [];
+  });
+}
+```
+
 A field literally named `default` is the **default target**, run when no target
 is named on the command line.
 
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -716,6 +716,30 @@ class Build
       }
     }
     ```
+  orderWith(_targets: Map<string, TargetBuilder>): OrderingEdge[] | Promise<OrderingEdge[]>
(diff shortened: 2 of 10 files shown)


totollygeek:
Re: the AI-review low finding (lazy hook can do runtime I/O, `build.ts:209`) — by-design, no change needed. `orderWith` is an **override on the build author's own `Build` subclass** — first-party code in the same trust context as `extraEdges`, `stateStore()`, `remoteCache()`, `mcpIdentity()`, and every `target().executes(...)` body, all of which already run arbitrary code and I/O. It is not attacker-controlled input and crosses no new trust boundary: anyone who can override `orderWith` can already run any code in a target body. Performing I/O (loading `dependency-graph.json`) is the *point* of the seam — that's why it's async. The failure paths are handled: the run path fails the build cleanly on a provider error, and cancel degrades to base order and still compensates. The reviewer concurs no high-confidence issue is introduced (2/10, low).

PR #206, opened 2026-07-19
feat(core): npm-package tool provisioning in toolchain()

totollygeek:
## Track B T2 — provision npm-package tools via `toolchain()`

Not every tool ships a release binary — many (`vitest`, `dprint`, `@nestjs/cli`,
…) live on the **npm registry**. This adds npm-package provisioning next to the
existing release-binary toolchain, so a build can pin and cache them without an
ambient `npm ci`.

### What's new

- **`installNpmTool(spec, options)`** (the npm counterpart of `installRelease`):
  runs a scoped, version-pinned `npm install --prefix <dir> --no-save
  <name>@<version>` into `<destDir>/npm/<name>@<version>` and returns the
  installed bin's `AbsolutePath` (`.cmd`-shimmed on Windows). A marker file
  records the pinned `{ name, version }`; a matching marker whose bin is still
  present **skips npm** on a later run.
- **`Toolchain.npm(spec)` / `ToolTasks.npm(spec, options?)`** — group npm tools
  in a toolchain (installed concurrently with release tools, keyed into the same
  `Map<name, AbsolutePath>`) or provision one directly. `bin` covers packages
  whose bin name differs from the package (`@nestjs/cli` → `nest`).
- **npm is the one ambient requirement** — it resolves and downloads the
  package; the subprocess goes through an injectable runner so tests are
  hermetic and network-free. An explicit `.toolPath(...)` pin still wins over
  `node_modules/.bin` resolution (composes with T1).

### Fails loudly, never poisons the cache

Mirroring `installRelease`, the install throws — **and records no marker** —
when the spec is malformed (an unsafe name, version, or bin), when npm fails, or
when npm exits 0 but the expected bin is absent (a typo'd `bin`, or a package
that ships no executable). A bad install errors once, here, instead of returning
a path to a nonexistent file that only breaks later at `.toolPath(...)` and
re-runs npm on every invocation forever.

### Tests

Unit (`packages/core/tests/npm_tool_test.ts`, + `tool_test.ts` for the
`Toolchain`/`ToolTasks` wiring) and integration
(`tests/integration/toolchain_npm_test.ts`, a real build provisioning a tool
end-to-end via the CLI) — all through a fake npm runner. `npm_tool.ts` is at
100% line/branch coverage.

### Adversarial review

Ran a 6-dimension adversarial pass (cache/marker correctness, path/argv safety,
concurrency, error handling, cross-platform, test adequacy) → refute-by-default
verify. **6 confirmed findings, all fixed with a regression test each:**

- **Returned a bin path without verifying npm produced it** — a typo'd `bin` or
  a bin-less package returned a phantom path and re-ran npm forever. Fixed with
  the post-install bin check above.
- **`..` in name/bin could escape `<destDir>`, and a leading-dash name could be
  parsed by npm as a flag** (both trusted-author-gated). Fixed with front-door
  name/version/bin validation that rejects the offender with a friendly error
  *before* npm runs (a valid scoped `@nestjs/cli` still passes).
- **Untested error paths** — a rejecting npm runner propagating without writing
  a marker, and the non-`NotFound` re-throw arms — now have regression tests
  (the coverage lift to 100%).

The one finding left as-is by design: a missing-`npm` spawn surfaces the raw
Deno error, exactly as `installRelease` leaves its download/copy errors
unwrapped — consistent sibling behavior, not a regression.

### Gate

clean; `docs/installing-tools.md` and the `zuke-write-build` skill updated.

### DIFF
--- a/docs/installing-tools.md
+++ b/docs/installing-tools.md
@@ -10,10 +10,11 @@ Tools are provisioned in Zuke's fluent settings-lambda style — the same
 `(s) => s.method(...)` shape the [tool wrappers](./tools.md) use — through two
 entry points in `@zuke/core`:
 
-| API                                                          | Installs   | Reach for it when                |
-| ------------------------------------------------------------ | ---------- | -------------------------------- |
-| [`ToolTasks.install((s) => …)`](#tooltasksinstall--one-tool) | one tool   | you need a single binary         |
-| [`toolchain()`](#toolchain--many-tools)                      | many tools | a build depends on several tools |
+| API                                                          | Installs      | Reach for it when                    |
+| ------------------------------------------------------------ | ------------- | ------------------------------------ |
+| [`ToolTasks.install((s) => …)`](#tooltasksinstall--one-tool) | one binary    | you need a single release binary     |
+| [`toolchain()`](#toolchain--many-tools)                      | many tools    | a build depends on several tools     |
+| [`ToolTasks.npm(...)` / `.npm(...)`](#npm-package-tools)      | an npm package | the tool ships on the npm registry   |
 
 Both return the installed binary's [`AbsolutePath`](./paths.md); hand it to a
 **[wrapper](./tools.md)** (`.toolPath(...)`), to `CmdTasks`, or to `defineTool`
@@ -190,7 +191,8 @@ class Deploy extends Build {
 - **Custom downloader.** `install({ download })` swaps the downloader for every
   tool — a test seam — and a per-tool `.download(...)` applies when the
   toolchain sets none.
-- **Introspection.** `chain.tools` returns the configured settings in order.
+- **Introspection.** `chain.tools` returns the configured release settings in
+  order; `chain.npmTools` the [npm-package tools](#npm-package-tools).
 - **Cheap to re-run.** Because a matching checksum is a cache hit, calling
   `install()` again (locally or on CI) is a no-op once the tools are present.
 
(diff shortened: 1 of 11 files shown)


totollygeek:
Thanks — dispositioning the three AI-review findings:

**1. `npm` resolved from PATH, not pinned** (security, low) — **by design.** npm is the one deliberate ambient requirement: it resolves and downloads the package, and a build cannot provision npm *with* npm. This is documented in the module JSDoc and `docs/installing-tools.md`, and matches the plan ("npm itself is the one ambient requirement; document that"). The subprocess goes through an injectable runner so it is still fully test-seamed.

**2. `installNpmTool` is a loose exported function** (generic, low) — **by design, for symmetry.** It is the npm counterpart of the already-exported `installRelease`; the two are a matched pair of low-level *install primitives* (the sanctioned exception to the task-shaped-API rule, alongside `Build`/`target`/`run`). The ergonomic, task-shaped surface is `ToolTasks.npm(...)` / `Toolchain.npm(...)`; `installNpmTool` is the primitive they wrap, exactly as `ToolTasks.install` wraps `installRelease`.

**3. Non-atomic cache promotion** (generic, medium) — **within the documented ceiling; invariant is sound.** The marker is written **only after** the post-install bin check confirms the bin is present, so a reader never observes a marker without its bin (no half-written cache hit). Concurrent installs of the *same* pin into the *same* directory only do redundant work — the accepted ceiling from the plan; a build resolves its toolchain once, and distinct pins use distinct directories. I added a JSDoc note making this invariant and ceiling explicit (latest commit).

Separately, a 6-dimension adversarial review before this PR found and fixed 6 confirmed issues with regression tests (notably: the install now fails loudly and writes no marker when npm produces no bin, instead of returning a phantom path; plus front-door name/version/bin validation). `npm_tool.ts` is at 100% line/branch coverage.

PR #209, opened 2026-07-19
feat(cli): node-repo ergonomics — npm workspaces, setup collision guard, zuke doc

totollygeek:
## Track B T4 — Node-repo ergonomics batch (FR-5, FR-6, FR-7)

Three small, independent DX fixes for driving Zuke inside a Node monorepo.

### FR-5 — `NpmTasks.run().workspaces()`

`.workspaces()` runs a script in **every** npm workspace (`--workspaces`),
composing with `.ifPresent()` to skip those missing the script. It's mutually
exclusive with `.workspace(name)` — setting both throws a friendly build error
instead of silently emitting a contradictory command. Pure argv change.

### FR-6 — `zuke setup` directory-collision guard + `--launcher-name`

`zuke setup` used to **silently skip** any existing path — including a `zuke/`
**directory** — producing a launcher-less setup (a real footgun in package
monorepos). Now setup does a pre-pass over every name it writes (`zuke.ts`, the
launchers, `zuke.json`, `deno.json`, `.gitignore`) and, if one is a directory,
**fails up front with an actionable error before writing anything**. The new
`--launcher-name <name>` writes the launcher scripts under a different base name
for repos that must keep the `zuke/` directory. A thrown setup error surfaces as
a clean message and a non-zero exit, not an uncaught stack trace. `zuke import`
(which reuses the scaffolder) inherits the guard.

### FR-7 — `zuke doc <package>`

`deno doc jsr:@zuke/deno` inside a Node repo drowns in `@types/node` resolution
warnings. `zuke doc <package>` runs `deno doc` from an **isolated throwaway
directory**, so type resolution starts clean. A bare name resolves to
`jsr:@zuke/<name>`, a scoped name to `jsr:<name>`, an explicit
`jsr:`/`npm:`/`https:` specifier passes through, and a relative path is pinned to
the caller's directory before the isolation kicks in. Extra flags pass through to
`deno doc`.

### Tests

Unit (npm argv, setup scaffolding, `resolveDocSpec`) and CLI-driven tests through
the real `main()` entry point (collision → friendly exit 1, `--launcher-name`
rename, `zuke doc` argv + a real `deno doc` subprocess end-to-end). All hermetic:
fake host / fake runner, no real npm, no network. npm 100% / setup.ts 100% line /
cli mod.ts 98% line.

### Adversarial review

Ran a 5-dimension pass (npm argv, setup collision, launcher-name escape, doc
spec/spawn, test adequacy) → refute-by-default verify. Confirmed findings fixed
with regression tests:

- **`--launcher-name` shadow check was case-sensitive** (medium) — on a
  case-insensitive filesystem (macOS/Windows) `--launcher-name Zuke.ts` slipped
  past the guard and, with `--force`, overwrote the `zuke.ts` build file. Now the
  check is case-insensitive.
- **`--launcher-name` allowed a colon** (low) — `C:evil` (Windows drive-relative)
  and `foo:bar` (NTFS alternate data stream) passed the separator guard. Now
  rejected.
- **`zuke doc ./relative.ts` was broken** by the temp-dir cwd change — a relative
  path resolved against the throwaway directory. Now pinned to the caller's cwd
  first.
- Test-adequacy gaps closed: the real doc runner's non-zero-exit path, the
  case-variant shadow, and the colon guard now have regression tests.

The one refuted finding: the setup directory-collision guard already covers
setup directory-collision guard already covers `deno.json`/`.gitignore` (a
finding claiming otherwise was refuted against the real code).

### Gate

`generate-ci --check`, and `deno doc --lint` over the core entrypoints clean;
`docs/tools.md`, `docs/getting-started.md`, and the `zuke-setup` skill updated.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -25,7 +25,15 @@ zuke setup                                  # in your project
 ```
 
 Without installing, the same wizard runs via `deno run -A jsr:@zuke/cli setup`
-(flags: `--dir <path>`, `--name <Class>`, `--force`, `--yes`).
+(flags: `--dir <path>`, `--name <Class>`, `--force`, `--yes`). If a `zuke/`
+directory already occupies the launcher's name, setup stops with an actionable
+error — pass `--launcher-name <name>` to write the launcher (and its `.ps1`)
+under a different name.
+
+Already have a `@zuke/*` package's API in hand? `zuke doc <package>` prints it
+(`zuke doc core`, `zuke doc @scope/pkg`), running `deno doc` in an isolated
+directory so a surrounding Node project's `@types/node` resolution doesn't drown
+the output.
 
 ## Migrate an existing project with `zuke import`
 
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -126,3 +126,24 @@ await helmUpgrade((s) => s.arg("api", "./chart").flag("install").cwd("infra"));
 `flag`/`option` add a `--` prefix unless the name already starts with a dash (so
 `flag("-v")` stays `-v`). Argv is a discrete array end-to-end, so a `defineTool`
 command is just as injection-free as the built-in wrappers.
+
+## Reading a wrapper's API — `zuke doc`
+
+`deno doc jsr:@zuke/deno` run inside a Node repo drowns the output in
+`@types/node` resolution warnings, because Deno discovers the surrounding
+project's `deno.json` / `node_modules`. `zuke doc <package>` runs `deno doc` from
+an **isolated throwaway directory**, so type resolution starts clean and the API
+prints readably in place:
+
+```sh
+zuke doc core            # jsr:@zuke/core
+zuke doc @scope/pkg      # a scoped package
+zuke doc deno --filter DenoTasks   # extra flags pass through to deno doc
+```
+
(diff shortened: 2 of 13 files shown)


totollygeek:
Thanks — one AI finding to disposition (the other review returned 0 findings):

**`zuke doc` forwards raw path/URL specifiers to `deno doc`** (security, low) — **by design.** `zuke doc` is an explicit, documented passthrough to `deno doc`; running it with isolated type-resolution *is* its whole purpose. The specifier is the user's own CLI argument, so no privilege boundary is crossed — it is equivalent to the user typing `deno doc <spec>` themselves, and `deno doc` fetching a `jsr:`/`npm:`/`https:` specifier they named is the intended behavior. The argv is a discrete array (`["doc", ...extra, spec]`) handed to `Deno.Command`, so there is no shell-injection surface. `resolveDocSpec` and the help text document that an explicit `jsr:`/`npm:`/`https:`/path specifier passes through as-is (a bare name resolves to `jsr:@zuke/<name>`).

Separately, a pre-PR 5-dimension adversarial review found and fixed three real issues with regression tests: a case-insensitive `--launcher-name` shadow that could overwrite the build file on macOS/Windows, a colon slipping past the launcher-name separator guard (Windows drive/ADS), and `zuke doc ./relative.ts` breaking under the temp-dir cwd. All checks are green across the three-OS matrix.

PR #215, opened 2026-07-20
feat(core): install multi-file runtime trees via installTree

totollygeek:
## What

`installRelease` extracts exactly one binary, so a multi-file runtime such as **Node.js** — which ships `bin/node`, `bin/npm`, `bin/npx`, and `lib/node_modules/**` in one tarball — could not be toolchain-provisioned (installing left only `.zuke/tools/node`; npm/npx were discarded). This adds **`installTree`**, which keeps the whole extracted archive tree and returns its root as a callable `AbsolutePath`:

- `root("bin", "node")` — a binary to invoke
- `root("bin")` — the directory to put on `PATH` (the FR-9 follow-up)

Surfaced through `ToolTasks.installTree((s) => …)`, `Toolchain.tree(...)`, and `ToolInstallSettings.strip()/.bins()`.

## How it works

- **Tar symlink support** — the tar reader (`untar`/`tar`/`TarEntry.linkname`) now preserves symbolic-link entries, because Node's `bin/npm`/`bin/npx` are symlinks into `lib/node_modules`. A new **`assertSafeLinkTarget`** rejects absolute or `..`-escaping targets (the second escape vector a symlink adds beyond the existing name check).
- **`strip` option** on `extractTarGzip`/`extractZip` drops the leading `tool-v1.2.3/` directory a release tarball wraps everything in.
- Declared `bins` are marked executable on POSIX (the tar reader does not preserve mode bits; `chmod` follows a symlinked bin to its real target).
- Cached by the pinned archive checksum plus a declared-bins-present check.

## Tests

- **Unit** — tar symlink round-trip, `strip`, `assertSafeLinkTarget` (accept/reject), duplicate-entry overwrite, and `installTree` (extract/strip/chmod/symlink-resolve, checksum verify + cache reuse, checksum mismatch, version-bump re-install, missing-bin/corrupt-marker re-install, zip tree, no-checksum path), plus `ToolTasks.installTree` / `Toolchain.tree` / `treeOptions_`.
- **Integration** — a build provisions a runtime tree end-to-end via the CLI, proving the symlinked bin resolves and the real bin is executable (POSIX) / the tree extracts (Windows).

## Adversarial review

Two independent reviewers attacked the change. **Confirmed & fixed:** a duplicate archive entry crashed extraction with a raw `AlreadyExists` (now last-one-wins, with a regression test); two branch-coverage gaps closed. **Refuted:** symlink escape (incl. directory-traversal), strip×safety ordering, and zip-slip are all bounded within the destination root. **Declined (documented ceiling):** the tree cache is keyed on the pinned download checksum rather than re-hashing the whole tree — the download checksum is the integrity boundary, and re-hashing an ~80MB runtime every build is the cost a single-binary install avoids.

All gates green: `deno task ci` (lines 98.3%, branches 95.3%), `apiDocsCheck`, `generate-ci --check`, and `deno doc --lint`.

### DIFF
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -88,6 +88,13 @@ function assertSafeEntryName(name: string): void
   absolute path or one with a `..` segment (a "zip slip"). A downloaded or
   poisoned archive must never place files outside where it is being unpacked.
 
+function assertSafeLinkTarget(entryName: string, target: string): void
+  Reject a symlink whose target would resolve outside the destination directory
+  — an absolute target, or a relative one that climbs (with `..`) above the
+  extraction root once resolved against the link's own directory. A file entry's
+  name is bounded by {@link assertSafeEntryName}; a symlink adds a second escape
+  vector (its target), so a poisoned tarball can't plant `bin/x -> ../../etc`.
+
 async function cancelRun(build: Build, options: CancelOptions): Promise<CancelResult>
   Cancel the run `options.runId` for `build`: transition it to `cancelling`
   (exactly one canceller drives the walk; a live owning process observes the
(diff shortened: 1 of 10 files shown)


totollygeek:

**generic review (0/10, 0 findings)** — no action.

**security review (1 low, `install.ts:374`) — tree cache checks bin presence, not content integrity.** Acknowledged and **intentionally declined** as a documented ceiling (see the `ponytail:` comment on `cachedTree`), for the same reasons the T3 zip-bomb bound was declined:

- The pinned **download checksum is the integrity boundary** — a tampered-in-transit archive is rejected before anything is unpacked, and a version bump (different checksum) always misses the cache and re-installs.
- `installRelease` re-hashes on cache hit because it guards **one small binary**; a runtime tree is hundreds of files / ~80 MB, and re-hashing it on every build (many builds resolve their toolchain each run) is precisely the cost this cache exists to avoid.
- `.zuke/tools/` is local build state under the same trust boundary as `node_modules/` — a local attacker who can rewrite files there can already rewrite the build script.
- To force a fresh verified install, bump/re-pin the `.checksum(...)`.

Re-hashing just the declared bins was considered and rejected: the bins are the largest files (node is ~80 MB), so it reintroduces the very cost the cache avoids for no gain over the download-checksum boundary.

All 9 checks green (Windows Tests fixed by passing an explicit symlink `type`). No code change for this finding.

PR #219, opened 2026-07-20
fix(core): executor async-safety and durable wait-timeout

totollygeek:
First PR of the remediation plan (`plans/REMEDIATION_PLAN.md`, PR 1 — the 🔴 critical async-safety items). Four defects that all undermined the framework's core promise of durable, resumable runs.

## Fixes

**1.1 — Wait `.timeout()` never fires (C1).** Every re-suspend recomputed `deadline = now + timeout`, so an hourly `resume --check` pushed the deadline forward forever. The deadline is now recorded once (first suspend) and preserved across re-suspends via `env.priorWaits` (seeded from the resumed record). The timeout finally arrives.

**1.2 — Unobserved rejection in the schedulers (C2/F2).** `onlyWhen`/`cacheKey` thunks and lifecycle hooks reject *outside* `runTarget`'s try/catch. The parallel scheduler's `.then` had no `.catch` (→ unhandled rejection + a wedged slot that hangs the run); the sequential path let the reject unwind out of `execute()`, stranding the record `running`. Both now route such a reject into the normal failure path so the run finalizes.

**1.3 — Lock-renewal heartbeat could crash the build (C2/F3).** `void store.renewLock(...)` in a `setInterval` didn't swallow rejections. Renewal and release are now best-effort (`.catch`) — a failed heartbeat just lets the lock lapse at its TTL (the documented backstop) instead of crashing from a background timer.

**1.4 — Raw secret emitted to a caller's reporter (F4).** The GitHub `::add-mask::<secret>` directive (raw value, deliberately un-redacted) was gated only on `style.github`. A caller-supplied reporter *is* the base reporter, so an embedded `execute()` was handed the plaintext. Now gated on `writesToConsole` too — the directive reaches only the real runner stdout.

## Adversarial review

Two independent reviewers attacked the change (the plan flags PR 1 for this). The **security** pass found nothing (the mask fix is complete). The **async-safety** pass found a defect my first cut introduced: a **throwing reporter** — a buggy custom sink, or **EPIPE when stdout is piped to a reader that closes** (`zuke build | head`) — makes `failTarget → reporter.error` throw, which still escaped the schedulers and stranded the record `running`. Fixed at the root: reporter output is now **best-effort** (`safeReporter`), so a broken sink can never break the build; the parallel failure handler also frees its slot *before* printing. The lenient-target failure arm got a dedicated test.

## Tests

Integration (`waiting_test.ts`): a re-suspend preserves the original deadline. Unit (`executor_test.ts`): throwing `onlyWhen` fails the target without crashing (sequential + parallel); a rejected lock renewal never crashes the build; a **throwing reporter** never strands the record (sequential) nor hangs the parallel scheduler; a lenient target's failure doesn't halt its siblings. Existing mask tests updated to assert the no-leak property + a real-console positive.

All green: `deno task ci` (lines/branches gate), `apiDocsCheck`, `generate-ci --check`, `deno doc --lint`. Commit signed. No public API change.

### DIFF
--- a/packages/core/src/executor.ts
+++ b/packages/core/src/executor.ts
@@ -102,6 +102,33 @@ function redactingReporter(inner: Reporter, redactor: Redactor): Reporter {
   };
 }
 
+/**
+ * Wrap a reporter so a failing write can never escape into the run. Output is a
+ * best-effort side effect: a sink that throws — a custom reporter with a bug, or
+ * the default console raising `BrokenPipe`/EPIPE when stdout is piped to a reader
+ * that closed early (`zuke build | head`) — must not turn into a rejection that
+ * unwinds the scheduler and strands the durable run record `running`. A thrown
+ * write is dropped; every other write still lands.
+ */
+function safeReporter(inner: Reporter): Reporter {
+  return {
+    info: (line) => {
+      try {
+        inner.info(line);
+      } catch {
+        // best-effort: a broken output sink must not break the build
+      }
+    },
+    error: (line) => {
+      try {
+        inner.error(line);
+      } catch {
+        // best-effort: a broken output sink must not break the build
+      }
+    },
+  };
+}
+
 /** Options for {@link execute}. */
 export interface ExecuteOptions {
   /** Suppress all banner/summary output (used by tests). */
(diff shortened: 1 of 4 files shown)


totollygeek:

**generic review (0/10, 0 findings)** — no action.

**security review (1 low, `executor.ts:108`, `safeReporter`) — swallowing reporter writes could hide secret-mask / audit delivery.** Acknowledged and **accepted as a deliberate trade-off**, because the paths that actually matter for secrecy and audit do not depend on the reporter:

- **Zuke's own output is redacted regardless.** `safeReporter` wraps `redactingReporter`, so redaction runs *before* the write — a swallowed write drops a line, it never emits an unmasked secret. Zuke never prints a secret in the clear even if a write fails.
- **The `::add-mask::` directive is only undeliverable when the sink is already broken.** It's a belt-and-suspenders for the *runner* to mask *other* processes' output. The one realistic failure is EPIPE — stdout piped to a reader that closed (`zuke build | head`) — in which case there is no log sink to leak into. It is also emitted during parameter resolution, before any run record exists, so it cannot strand a record.
- **The audit trail is durable in the state store, not the reporter.** `runs show mcp-audit` reads persisted records written through the writer/store, which this change does not touch. Swallowing reporter I/O has no effect on audit durability.
- **The alternative is worse.** Letting a write throw is exactly the confirmed C1/C2 defect this PR fixes: a broken sink (EPIPE, or a buggy custom reporter) escaping `failTarget` and stranding the durable record `running` — un-resumable and un-cleanable. A build must not die because its output pipe closed.

Net: output is best-effort by design; the security-relevant guarantees (redaction of Zuke's output, durable audit in the store) are preserved by other mechanisms. No code change.

All 9 checks green; commit signed.

PR #221, opened 2026-07-20
feat(deno): make the coverage gate trustworthy

totollygeek:
Remediation plan PR 2, part 2 (items 2.2 + 2.3) — split from the `deepEqual` fix (#220) because this is a `feat(deno)` (new API + `@zuke/deno` bump), whereas that was test-only.

## Fixes

**2.2 — the gate passed vacuously on an empty report (M1).** `pct` returned `100` for `0/0`, so a run that instrumented **no code at all** went green at "100%". `enforceCoverage` now fails a gated run whose report is empty (no lines *and* no branches measured) with a clear `no coverage data measured` message.

**2.3 — the threshold was aggregate-only (M2).** A wholly-untested file could dilute into a healthy repo-wide average. New **`.perFileThreshold(percent)`** (+ `CoverageThresholds.perFile`) fails the gate when any single **instrumented** file's line coverage is below the floor. Zuke wires a **50%** floor into its own gate — well below the current lowest src file (~82%), so it flags a genuinely neglected file without churning on the existing spread.

## Adversarial review

A reviewer confirmed two MEDIUM issues in my first cut, both fixed here:

1. **Branchless code spuriously failed.** `threshold(95)` sets both lines and branches, and my initial unconditional `branchesFound===0` failure would fail a real, 100%-line-covered but **branchless** module ("no branch data"). The true "nothing measured" signal is `linesFound===0`; branchless-but-measured code is now correctly a vacuous 100% branch pass (regression test added). Zuke's own gate was never affected (thousands of branches) — it was a downstream footgun.
2. **The per-file floor is blind to never-imported files.** `deno coverage` only reports *loaded* files, so a source file no test imports is invisible to any coverage metric. The JSDoc previously overclaimed ("a wholly-untested file or package can't hide"); it now states the tool's limit plainly. The floor still catches the common regression — an imported-but-under-tested file (reported at 0%). A src-tree glob for a stronger guarantee is noted as a possible follow-up.

Refuted: LCOV per-file segmentation (Windows `C:` paths preserved), malformed-LCOV crash, double-count, per-file false positives, test-file leakage — all clean.

## Tests

`parseLcovPerFile` segmentation (incl. colon paths); empty-report fail; branchless-pass; per-file floor flags a low file while the aggregate passes; per-file skips zero-line files; a per-file floor alone gates and forces `--lcov`; the getter/argv wiring. The old "zero-found as fully covered" test was inverted to assert the new fail-on-empty behaviour.

All green: `deno task ci` (the gate now runs with the new floor on this repo), `apiDocsCheck`, `generate-ci --check`. Pre-existing `deno doc --lint` errors are unchanged (that's plan PR 7.5); this PR adds none. Commit signed.

### DIFF
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -3335,6 +3335,11 @@ class DenoCoverageSettings extends DenoSettings
     Fail the gate if branch coverage is below `percent` (see {@link linesThreshold}).
   threshold(percent: number): this
     Fail the gate if either line or branch coverage is below `percent`.
+  perFileThreshold(percent: number): this
+    Fail the gate if any single instrumented file's line coverage is below
+    `percent` — a per-file floor, so an under-tested file can't hide inside a
+    healthy aggregate (see {@link CoverageThresholds.perFile}, which notes the
+    `deno coverage` limit for files no test loads).
   get thresholds(): CoverageThresholds
     The configured thresholds; read by {@link DenoTasks.coverage}.
   get outputPath(): string | undefined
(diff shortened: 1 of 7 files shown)


totollygeek:

**security review (0/10, 0 findings)** — no action.

**generic review (2 low, `coverage.ts:94`) — trailing `\r` in parsed file paths on CRLF reports.** Valid (if cosmetic): the per-file percentages were always correct (`Number` trims `\r`), but a Windows-authored LCOV left a trailing carriage return on the `SF:` path, which would surface in a per-file failure message as `file.ts\r`. **Fixed** — `parseLcovPerFile` (and `parseLcov`) now split on `\r?\n`, so paths are clean regardless of line endings. Added a CRLF regression test.

All 9 checks green; commit signed.

PR #223, opened 2026-07-20
fix(core): validate MCP Origin and redact URLs; fix ai diff parsing

totollygeek:
Remediation plan **PR 4** — MCP HTTP `Origin` validation (the other security-flagged item) plus the `@zuke/ai` diff-parsing correctness fixes. Touches `@zuke/core` (4.1/4.4/4.5) and `@zuke/ai` (4.2/4.3), so both take a patch bump.

## Fixes

- **4.1 (🟠) — No `Origin` validation on the MCP HTTP transport.** A browser drive-by / DNS-rebinding page could POST to a loopback-bound server. `serveHttp` now rejects (403) a request whose `Origin` isn't allowed: on a loopback bind only loopback origins pass; a client that sends no `Origin` (CLI/MCP) is unaffected; a non-loopback bind is the operator's responsibility (or set `allowedOrigins`). A Content-Type check was deliberately *not* added — Deno/browser `fetch` defaults a string body to `text/plain`, so requiring JSON would break legit clients, and `Origin` already covers the cross-origin threat.
- **4.2 (🟡) — Diff parser dropped removed lines starting `--`.** Removing `--verbose` (diff line `---verbose`) was mistaken for a `---` header and lost. Header detection is now gated to *before* the hunk (the `+++ b/` symmetric case too).
- **4.3 (🟡) — `filterDiff` failed open on spaced paths.** `\S+` stopped at the space, so an excluded file with a space in its name was reviewed anyway. It now reads the unambiguous `+++ b/` line; an unparseable *file* section is dropped when filters are active (fail-safe), preamble is kept.
- **4.4 (🟡) — `--data` byte cap measured UTF-16 code units.** Now `TextEncoder().encode(raw).length`, so a multibyte payload over the byte budget is rejected.
- **4.5 (🟡) — `HttpError` echoed the full URL.** It now redacts userinfo and credential query params before the URL reaches the message or `.url`.

## Adversarial review

Two reviewers attacked the change. **1 HIGH confirmed and fixed:** `isLoopbackHost`'s unanchored `/^127\./` accepted an attacker domain like `127.0.0.1.evil.com`, bypassing the loopback guard — now a fully anchored dotted-quad (regression test added). **1 MEDIUM confirmed and fixed:** `redactUrl`'s exact-name denylist missed OAuth params (`client_secret`, `refresh_token`, …) — switched to substring matching (test covers both). Documented ceilings (not fixed): a secret in the URL *path* (non-standard), and the greedy `diff --git` fallback for a rename whose path contains ` b/` (no reviewable content at stake). Everything else — Host-header need, no-Origin bypass, allow-list matching, 4.2/4.4 correctness — refuted.

## Tests

`originAllowed` branch matrix (loopback variants incl. the `127.x.evil` bypass, allow-list, non-loopback bind); a live cross-origin POST → 403; the `--` removal parse; spaced-path include/exclude + unparseable-section fail-safe; the multibyte `--data` rejection; and `HttpError` userinfo + OAuth-param redaction.

All green: `deno task ci`, `apiDocsCheck`, `generate-ci --check`, `deno doc --lint`. Commit signed.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -4,6 +4,7 @@
   "useGitignore": true,
   "words": [
     "ACCESSTOKEN",
+    "apikey",
     "COLLECTIONURI",
     "Dependabot",
     "Dockerfiles",
@@ -22,6 +23,7 @@
     "Svensson",
     "TEAMPROJECT",
     "USERPROFILE",
+    "userinfo",
     "Vixie",
     "actionlint",
     "asciinema",
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -70,6 +70,14 @@ below), so one long `run:` never head-of-line-blocks another client's read.
   bind** a non-loopback address rather than exposing an unauthenticated
   endpoint.
 - A token is also enforced on a loopback bind when `ZUKE_MCP_TOKEN` is set.
+- **Origin validation** guards against a browser drive-by / DNS-rebinding page:
+  on a loopback bind, a request that carries an `Origin` header is accepted only
+  when it is a loopback origin, and rejected `403` otherwise. A client that sends
+  no `Origin` (a CLI/MCP client — not a browser) is always allowed, so this is
+  invisible to normal use. Permit a specific extra origin with
+  `--allowed-origin <origin>` (repeatable); when set, a present `Origin` must
+  match one exactly. A non-loopback bind runs no default Origin check — front it
+  with your own policy.
 - This is a bridge for a trusted network segment: **put real TLS and
   authentication in front of it** (a reverse proxy, a service mesh) for anything
   production-facing. Zuke provides the transport, not an internet gateway.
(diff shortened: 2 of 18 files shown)


totollygeek:
Dispositioning the AI security-review finding (low, \`qbehjz\` — "overbroad acceptance of malformed Origin on non-loopback binds", http.ts:118):

**Accepted as designed — not a defect.** The `if (!isLoopbackHost(bindHost)) return true;` branch is intentional. The Origin guard exists specifically as the drive-by / DNS-rebinding defense for **loopback** binds (the MCP-spec threat model: a malicious web page reaching a localhost server). A non-loopback bind (`0.0.0.0` / a public address) is already an explicit "expose this beyond localhost" decision, and for those the operator opts in to Origin filtering via `--allowed-origin` (or `allowedOrigins`), which is then enforced exactly. Applying a default loopback-only Origin filter to a public bind would break the normal remote-client case, not harden it.

The behavior is spelled out in the function JSDoc ("a non-loopback bind runs no default check") and covered by the `originAllowed` branch-matrix tests, including the `127.0.0.1.evil.com` / `localhost.evil.com` rejection cases. No code change.

PR #226, opened 2026-07-20
fix(core): harden executor cancellation, graph validation, and state edge cases

totollygeek:
Remediation plan **PR 6** — five executor/cancel/state edge cases in `@zuke/core`. All internal; no public API change.

## Fixes

- **F6 — retries/remediations ignore cancellation.** On a cancelled run (Ctrl-C / SIGTERM / aborted signal), `runBody` kept retrying and `runBodyWithRecovery` still fired `recoverWith` remediations — including a **paid AI-fixer** the user had just cancelled. Both now stop the moment `ctx.signal.aborted` is true.
- **F7 — `cancelling` double-settle over a live cleanup.** A second `zuke cancel`, or one racing the in-process executor, could finalize a run `cancelled` with `EMPTY_OUTCOME` ("no compensations") while another process was mid-compensation. Introduced a **per-run cancel lock** (`state/cancel_lock.ts`) — a store lock with a heartbeat and a 30s TTL crash backstop, held by whoever drives the walk. `cancelRun` and the executor both acquire it before compensating; a contender is a friendly no-op. The `recover` path (settle a stranded `cancelling` record) now only runs **while holding the lock**, so no live canceller exists.
- **F8 — waiting target stranded in a failed record.** A run that both failed *and* parked at a `.waitsFor()` gate left the gate row `waiting` forever in a terminal `failed` record no resume sweep reaches. Waiting rows + blocked deps now settle `skipped`.
- **F10 — graph validation only on the CLI path.** A forward-referenced dependency crashed `execute()`/`resumeRun()`/`cancelRun()` with a raw `TypeError`. The shared planning walk (`executionSet`) now raises the friendly `GraphError` for every entry point.
- **F12 — non-idempotent CAS double-apply.** On a conflict where `getRun` returned `null` (run vanished/pruned), the writer re-applied the mutator to the already-mutated copy — double-pushing an audit event and resurrecting a deleted run. It now drops the best-effort write.
- **F11** — removed a stray `ponytail:` placeholder token from a cancel comment.

## Tests

New tests per fix, incl. the **cancel-vs-cancel** and **cancel-vs-live-body** races, the executor lock-loss deferral, the CAS-vanish drop, the forward-ref `GraphError`, and the cancel-lock heartbeat renewal.

## Adversarial review

Ran an attack→verify pass across concurrency/regression/coverage dimensions. **No correctness defects survived.** Two confirmed **test-coverage** findings were fixed: added the executor lock-loss test + made the cancel-lock TTL injectable to exercise the heartbeat; also removed a dead ternary arm the review flagged.

## Verification

- `deno task ci` green: fmt, lint, spell, type-check, coverage (**lines 98.3%, branches 95.3%**).
- `./zuke apiDocsCheck` and `generate-ci --check` green — no drift, no public API change.

### DIFF
--- a/packages/core/src/cancel.ts
+++ b/packages/core/src/cancel.ts
@@ -40,6 +40,7 @@ import { absolutePath } from "./path.ts";
 import { findConfigDir, pathExists } from "./config.ts";
 import { defaultStateHost, type StateStore } from "./state/store.ts";
 import { resolveStateStore } from "./state/resolve.ts";
+import { acquireCancelLock } from "./state/cancel_lock.ts";
 import { resolveActor } from "./state/record.ts";
 import type {
   RunEvent,
@@ -483,7 +484,7 @@ function collectForEachInto(
  * when the cancel landed — an item mid-flight may have partial side effects
  * (a started deploy) its `.onCancel(...)` needs to unwind.
  *
- * ponytail: an out-of-process `zuke cancel` compensates a `running` item from a
+ * An out-of-process `zuke cancel` compensates a `running` item from a
  * record snapshot, so if that item's body is still live in the owning process
  * (which aborts only on its next state write / lock heartbeat), the compensation
  * can overlap the body's tail. Bodies that checkpoint via `ctx.state.set(...)`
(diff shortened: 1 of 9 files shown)


totollygeek:
Dispositioning the AI security-review finding (low, `15kipzh` — "Best-effort cancel-lock release can leave a live lock until TTL expiry", cancel_lock.ts:58):

**Accepted as designed — not a defect.** The best-effort `release()` (swallow a failed `releaseLock`, rely on the TTL) is deliberate and matches the repo's established lock pattern exactly — it is byte-for-byte the same as the `.lock()` release in `executor.ts:584-589`. The trade-off is intentional:

- A failed release must **not** turn a completed cancellation into a failure, so the rejection is swallowed.
- The **30s TTL is the backstop**: if a release ever fails to land (a transient store error), the lock lapses on its own and a re-cancel proceeds — the same crash-recovery path a dead canceller uses. The heartbeat only renews *while the holder is alive and actively cancelling*, so a stalled/failed holder does not keep renewing.
- The worst case is a bounded ≤30s window where a re-cancel of the **same** run returns a friendly "being cancelled by another process" no-op — strictly safer than the alternative (throwing on release failure, which would fail an otherwise-successful cancellation), and self-healing.

This is the intended semantics for the F7 hardening (single active canceller, TTL crash backstop). No code change.

PR #229, opened 2026-07-20
fix: align kubectl/jest/git wrapper argv with the real CLIs

totollygeek:
Remediation plan **PR 7** — the correctness half (**7.1–7.4**). These are argv bugs a single snapshot test can't catch; each wrapper emitted a command the real CLI rejects, or dropped a positional. Every fix ships an **interaction** test.

## Fixes

**kubectl** (`packages/kubectl`)
- **7.1** `apply`: `-f` and `-k` are now mutually exclusive; `-k` also rejects `-R` (kubectl: *"only one of -f or -k"* and *"the -k flag can't be used with -f or -R"*).
- **7.3(a)** `describe`: requires a resource type — a bare selector (`describe -l …`) is rejected by kubectl.
- **7.3(b)** `annotate`/`label`: require a resource type; `--all`/`--selector` are modifiers on a type, not substitutes.
- **7.3(c)** `logs`: rejects a pod name together with a selector (kubectl allows only one).
- **7.3(d)** `top`: `--containers` and `-A`/`--all-namespaces` are `top pod`-only and rejected on `top node`; a selector stays valid on nodes.
- **7.3(e)** `scale --current-replicas`: **no change** — the understand pass verified it is *not* deprecated (it is a documented precondition flag), so the plan item was refuted.

**jest** (`packages/jest`) — **7.2**: array flags (`--reporters`, `--selectProjects`) now emit `--flag=value` form, so jest's yargs parser binds one value per flag instead of greedily swallowing the trailing positional test-path patterns.

**git** (`packages/git`) — **7.4**: `add` emits `--` before pathspecs, so a path beginning with `-` is treated as a path, not parsed as a flag.

## Scope note

**7.5** (house-wide `deno doc --lint` first-party `private-type-ref` cleanup + wiring the lint into the gate) is a large, mechanical, ~50-package change of a different character — it ships as a **separate follow-up** so this PR stays reviewable and release attribution stays clean (like the earlier PR 2a/2b split).

## Verification

- `deno task ci` green: fmt, lint, spell, type-check, coverage (**lines 98.3%, branches 95.4%**); `./zuke apiDocsCheck` clean (guard-only changes, no public API drift).
- **Adversarial review** (too-strict guards + regressions → verify against real kubectl v1.36.1): **0 too-strict / 0 regressions**; it surfaced one adjacent too-loose case (the `-k`/`-R` combo) which is now guarded + tested.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -100,6 +100,7 @@
     "oxlint",
     "oxlintrc",
     "pagelen",
+    "pathspec",
     "pathspecs",
     "patriksvensson",
     "pedantic",
--- a/packages/git/src/git.ts
+++ b/packages/git/src/git.ts
@@ -167,7 +167,9 @@ export class GitAddSettings extends GitSettings {
     const argv = ["add"];
     if (this.#all) argv.push("--all");
     if (this.#update) argv.push("--update");
-    argv.push(...this.#paths);
+    // `--` so a pathspec beginning with `-` (e.g. `-weird.txt`) is treated as a
+    // path, not parsed by git as a flag. `add` positionals are always pathspecs.
+    if (this.#paths.length > 0) argv.push("--", ...this.#paths);
     return argv;
   }
 }
--- a/packages/git/tests/git_test.ts
+++ b/packages/git/tests/git_test.ts
@@ -63,6 +63,7 @@ Deno.test("add and commit render their options", () => {
     "git",
     "add",
     "--all",
+    "--",
     "src",
     "mod.ts",
   ]);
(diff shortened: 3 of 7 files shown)


totollygeek:
Dispositioning the AI security-review finding (low, `13x9fgu` — "Prompt-injection text embedded in comments/tests", jest_test.ts:39):

**False positive — no change.** Line 39 is `"--selectProjects=api"`, an element of an expected-argv array in the "every option renders" test; the surrounding new lines are jest CLI-flag assertions (`--reporters=default`, `--reporters=jest-junit`, `src/`, `test/`) and a comment explaining that jest's yargs parser would otherwise greedily swallow the positional `src/`. None of it is prompt-injection content or untrusted input — it is a pure `buildArgs()` argv snapshot. The reviewer appears to have pattern-matched the CLI-flag/parser phrasing. No code change; the test correctly proves the `--flag=value` fix (positionals survive).

All other checks are green and the generic review reported no findings.

PR #239, opened 2026-07-20
chore: repo hygiene — license metadata, lock, doc entrypoints, notes

totollygeek:
## What

Repo hygiene (remediation PR 10), no consumer-facing behaviour change.

- **Per-package license** — added `"license": "MIT"` to all 55 `packages/*/deno.json` (was 0/55) so a published JSR artifact carries its own license rather than relying solely on root inference. A new `release_config_test` case asserts every package declares it.
- **`deno.lock` cleanup** — removed the orphaned `jsr:@std/yaml@1` entry (nothing imports it; the CI-YAML generator hand-rolls its output). Surgical removal only — verified `deno check --frozen` and a `--frozen` cspell run still pass, and the diff touches nothing else.
- **`llms-full.txt` / README completeness** — `collectPackageDocs` now documents every declared entrypoint (via the existing `packageEntrypoints`), not just `mod.ts`, so core's `./shell`, `./tooling`, `./render`, `./conformance` appear in the generated docs — matching the "whole typed surface" claim.
- **`cleanDoc` path-strip** — `deno doc`'s multi-file mode (needed for the above) emits bare `file://<abs-path>` section headers and `reference X: file://…` lines. `cleanDoc` now strips them too, so no machine/checkout path leaks into the committed docs and `apiDocsCheck` stays location-independent (deterministic on any runner). Regression test added.
- **Docs** — a `SECURITY.md` "Known trade-offs" note on the `quality` job's fork-safe `contents: write` + `persist-credentials`, and an accurate constant-time note on `timingSafeEqual` (explicitly flagging the length short-circuit as the one accepted leak).

## Skipped

The plan's `.gitignore` local-vs-project split (`plans/`, `.omc/`, `docs/superpowers/` → a local exclude) is **not done**: those entries are load-bearing. The `cspell '**'` gate (and deno's file discovery) read `.gitignore`, not `.git/info/exclude`, so moving them out makes cspell lint local scratch files. Leaving them in `.gitignore` is the correct hygiene.

## Versioning

`chore` (no bump). The license metadata lands in each package's next release; the `cleanDoc` change is behaviour-identical for existing single-entrypoint doc consumers (the new filters only fire in multi-file mode, newly enabled here).

## Gate

`deno task ci` green (coverage 98.3% lines / 95.4% branches); `apiDocsCheck` + `docLint` pass and are now checkout-independent. Adversarial review: 2 dimensions (lock/license validity; doc-gen + note accuracy) — it **caught a confirmed critical**: the multi-entrypoint change initially leaked absolute paths into the committed docs and would have failed `apiDocsCheck` on CI. Fixed at the root (`cleanDoc`) with a regression test before this PR; re-verified no absolute path remains in any generated file.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -73,6 +73,14 @@ What the project does to keep releases trustworthy:
   tree as a backstop. The merged release tree should already be clean; once a
   real release confirms this, drop the flag for the strongest
   "published == committed source" guarantee.
+- **`contents: write` + `persist-credentials` on the `quality` job.** The
+  `quality` job in `ci.yml` runs on `pull_request` with `contents: write` and
+  `actions/checkout`'s `persist-credentials: true`, because the AI lint fixer may
+  push a fix commit back to the PR branch. This is a deliberate trade-off, and it
+  is fork-safe: `pull_request` (not `pull_request_target`) runs with the base
+  repository's read-only `GITHUB_TOKEN` for a fork PR, so the elevated write and
+  persisted credential apply only to same-repository branches, never to code a
+  fork controls. Workflow egress is audited by `step-security/harden-runner`.
 
 ## Running the scanners yourself
 
--- a/deno.lock
+++ b/deno.lock
@@ -1,13 +1,7 @@
 {
   "version": "5",
   "specifiers": {
-    "jsr:@std/yaml@1": "1.1.1",
     "npm:cspell@9": "9.8.0"
-  },
-  "jsr": {
-    "@std/yaml@1.1.1": {
-      "integrity": "a57665ecf3d17b926380593a56625d8a10cc7281802f1e993b5ebc94a48e71f8"
-    }
   },
   "npm": {
     "@cspell/cspell-bundled-dicts@9.8.0": {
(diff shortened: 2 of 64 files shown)


totollygeek:
**Disposition of the AI-review finding** (`11qpdsqlm6pmm` — "multi-entrypoint docs generation could leak file paths"):

Valid concern, and it's already handled. This exact risk surfaced in the pre-PR adversarial review: `deno doc` multi-file mode emits bare `file://<abs-path>` section headers and `reference X: file://…` lines that single-file mode never produced. `cleanDoc` (`packages/docs/src/render.ts`) now strips all three location shapes (`Defined in file://…`, bare `file://…`, and `reference … : file://…`), with a regression test in `render_test.ts`.

Verified, not just asserted:
- `grep` finds **zero** absolute paths (`/Users/`, `/home/`, `file:///`) in any generated file (`llms-full.txt`, `llms.txt`, all READMEs).
- **"Build & verify with Zuke" passed on the `/home/runner/…` CI checkout.** That job runs `apiDocsCheck`, which regenerates the docs and compares byte-for-byte with the committed files. A leaked absolute path would be `/home/runner/work/zuke/zuke/…` on the runner vs the committed `/Users/…`, so the check would fail red. It passed → the output is path-independent.

That cross-machine `apiDocsCheck` is also the standing backstop: if a future `deno doc` version emits a new path shape `cleanDoc` misses, the gate fails immediately rather than silently committing a path. Not suppressing — leaving the finding visible as a useful caution.

PR #247, opened 2026-07-22
fix(ai): close AI-review security regressions and hardening tail

totollygeek:
Implements **PR 1** of `plans/REMEDIATION_PLAN_V2.md` — the `@zuke/ai` security regressions and hardening tail. All fixes are model-reachable defenses on the fixer/reviewer paths.

## Fixes

- **🟠 Git pathspec-magic bypass (N1).** `normalizePath` (`apply.ts`) now rejects any edit path whose first segment begins with `:` (git reserves it for pathspec magic like `:(glob)`, `:/`), and `commitAndPush` stages with `git --literal-pathspecs add` — two independent defenses so a crafted path can no longer make the commit step re-stage the whole tree (the leak #222 closed).
- **🟡 Markdown fence/inline breakout (N3).** New `markdown.ts` with `fenceMarkdown` (a fence one backtick longer than any run inside the content) and `codeSpan` (the inline equivalent, with CommonMark padding). Applied to the agent-output block and the fix-report diff/headers; `cell()` now also collapses newlines. Model/agent output can no longer break out to inject Markdown (fake "approved" banners, phishing links) into PR comments or the job summary.
- **🟡 Suggest-mode diff scope (N2).** Suggest mode now snapshots the working tree **before** the agent runs and scopes `#postSuggestions` to only the agent's newly-changed paths, failing closed if the snapshot can't be taken — a tree dirtied by the developer or earlier steps is never posted as "agent" suggestions.
- **🟡 `fetchBase` fail-open (N4).** `#resolveDiff` distinguishes a failed base-diff fetch from a genuine no-change diff. An empty working-tree fallback after a fetch failure now **fails the gate** (or skips visibly under `onError: "warn"`) instead of silently passing an empty assessment. A genuinely empty successful diff still passes.
- **⚪ Env seam (N7).** `Reviewer.#publish` routes through the injected `this.#env` instead of the global reader.
- **🟡 Azure pagination (finding 6 residual).** Documented why the Azure DevOps PR-threads endpoint needs no pagination loop (it returns the full set — no `$top`/`$skip`/continuation), making the asymmetry with the other hosts intentional-on-the-record.
- **🟡 Suppression fingerprint change (N5).** Migration note in `docs/ai-review.md`: the 32→64-bit fingerprint change invalidated recorded suppressions; re-record them.

## Verification

- `deno task ci` green (check, tests, coverage ≥95%); `./zuke apiDocsCheck` green.
- **Adversarial review** (four attack dimensions — pathspec, markdown injection, suggest/fetchBase, regressions — each finding reproduced against the real code before verdict):
  - Found and **fixed** a real gap the pathspec/fence work missed: `cell()` didn't neutralize backticks, so a model-controlled `loc.file` in the `####`/`**Files:**` inline code spans could still inject inline Markdown. Fixed with `codeSpan` (variable-length delimiters) + tests.
  - Three other findings were **verified and refuted** against the code (a hypothetical `lang` info-string with no attacker-reachable caller; a rename/copy "leak" that produces a pure-insertion diff which `diffToSuggestions` skips, so nothing is posted; a "missing test" that was correct behavior). I still added the `lang` neutralization and the empty-successful-diff regression test as cheap contract-locking.

## Deferred

- **⚪ Symlink traversal in `applyEdits` (N6, pre-existing).** A robust `realpath` guard needs a subprocess/e2e test (its own cwd) to exercise real symlinks hermetically; disproportionate here. Filed for a follow-up — it's pre-existing and requires an attacker-planted in-tree symlink, unlike the regressions above.

Every fix ships unit + integration tests in this PR.

### DIFF
--- a/docs/ai-review.md
+++ b/docs/ai-review.md
@@ -158,6 +158,11 @@ outage isn't a silent skip either.
   suppressed finding is still listed (under "Suppressed (not gating)") so the
   dismissal is auditable — it mutes the gate, it never silently buries a finding.
 
+  > **Migration:** the finding-fingerprint format was widened (32→64-bit), so
+  > IDs recorded before that change no longer match. If previously dismissed
+  > findings resurface, re-record them: copy the new ID shown in the report back
+  > into your suppress file (default `.zuke/ai-suppress.json`).
+
 ## GitHub Actions summary
 
 Under Actions, a review appends a Markdown section (score, severity, and a
--- a/docs/ai-review.md
+++ b/docs/ai-review.md
@@ -9,9 +9,10 @@ blob of prose.
 
 ## How it plugs in
 
-A reviewer is a [`Validation`](./authoring.md#validations--validatebefore--validateafter):
-an object with a `validate(ctx)` method. You build one fluently, then attach it
-to a target with `.validateBefore(...)` (gate before the body) or
+A reviewer is a
+[`Validation`](./authoring.md#validations--validatebefore--validateafter): an
+object with a `validate(ctx)` method. You build one fluently, then attach it to
+a target with `.validateBefore(...)` (gate before the body) or
 `.validateAfter(...)` (check after a successful body). The target decides _when_
 it runs; the reviewer decides _what_ it checks.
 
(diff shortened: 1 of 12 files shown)


totollygeek:
Thanks — addressed the AI security-review finding (`docs/ai-review.md`, suppression-fingerprint migration).

That note is the deliberate mitigation for the fingerprint-width change (remediation item 1.5); the change itself pre-dates this PR (#235). The reviewer's concern was that dismissed findings could reappear, so I reworded the note to make the **fail-safe direction** explicit: a stale suppression is *ignored* (the finding reappears in the report and re-trips the gate), never silently applied to hide a finding. It's a one-time re-record of the new ID.

The generic review is clean, and all checks are green.

totollygeek:
Re: the new security-review finding — "Untrusted model text is shown in docs as a migration note without explicit sanitization guidance" (`docs/ai-review.md:157`): **false positive.**

I verified it against the lines. That whole region is hand-authored documentation prose (the cost-controls bullet, the `.suppress(...)` bullet, and the migration note). No model output or untrusted text is embedded or rendered there — there is nothing to sanitize. The runtime paths that *do* embed model/agent output (`fenceMarkdown` / `codeSpan` in the fixer report and agent output) are already hardened in this PR and were confirmed clean by the generic review.

The substantive point from the previous run (a stale suppression could confuse) is already addressed by the reworded note. Not editing the paragraph again, since it contains no untrusted text — doing so would only re-trigger the reviewer without any real change.

Generic review is clean (0 findings) and all checks are green.

PR #253, opened 2026-07-22
fix(core): flag dead ordering edges and document the fan-out ordering limit

totollygeek:
Implements **PR C** of `plans/17-consolidated-plan.md` — the `orderWith` × fan-out trap (upstream skill feedback #2). The pilot fed a service dependency graph into per-repo fork ordering; because fan-out sub-targets don't exist at plan time, the edges were dead on arrival — no error, no warning, invisible to `deno check` and `--list`/`graph`.

## Fixes

- **Diagnostic (feedback #3).** `orderWith`/`extraEdges` see only class-field targets, and `planEdges` silently drops any edge whose endpoint isn't in the run's execution set. The executor now **logs an edge as ignored** when its endpoint is neither in the run nor a declared build target — a fan-out per-item name, a typo, or an ad-hoc `target()`. An edge to a *declared* target simply not in this run (a conditional target) stays silently ignored, as before, so the diagnostic is quiet for legitimate use.
- **Docs (the direct fix).** The cheatsheet, `docs/authoring.md`, and `docs/orchestration.md` now state that per-item ordering across a fan-out is **not expressible** with `orderWith`/`extraEdges`, and give the workaround: order whole fan-out **waves** — one `.forEach()` per wave — chained with `.dependsOn`.

## Verification

- `deno task ci` green (check, tests, coverage ≥95%); `./zuke apiDocsCheck` green.
- Unit (`executor_test.ts`) + integration (`orderwith_test.ts`) tests cover: a dead edge to a non-build target is flagged; a conditional (declared-but-not-in-run) target is **not** flagged (no noise); and a malformed nullish endpoint is tolerated without crashing.
- **Adversarial review:** confirmed no false positives across legitimate `orderWith`/`extraEdges` patterns (object identity is consistent via the single shared `discovered` map), and caught a **crash regression** — `endpoint.name_` threw on a type-bypassed nullish endpoint, which the old `planEdges` tolerated. Fixed with a guard and pinned by a `@ts-expect-error` regression test.

## Deferred (noted in feedback #2)
Re-resolving ordering after fan-out materialisation, or per-item edges in `.forEach` settings, is a larger feature left for later — the wave workaround covers the need today.

### DIFF
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -697,6 +697,17 @@ override async orderWith(t: Map<string, Target>): Promise<OrderingEdge[]> {
 }
 ```
 
+The `targets` map holds only **class-field targets**. A
+[`.forEach()`](./orchestration.md#fan-out-over-a-list--foreach) fan-out's
+per-item sub-targets are materialised only while a run executes, so
+they are never in `targets` and **per-item ordering across a fan-out cannot be
+expressed** with `orderWith`/`extraEdges` (`t.get("parent[item].stage")` is
+`undefined`). Order whole fan-out **waves** instead — one `.forEach()` per wave,
+the waves chained with `.dependsOn` — so ordering lives between class-field
+targets. An edge whose endpoint is not a target in the build (a fan-out item
+name, a typo, or an ad-hoc `target()`) is logged as ignored when the run plans,
+rather than silently doing nothing.
+
 A field literally named `default` is the **default target**, run when no target
 is named on the command line.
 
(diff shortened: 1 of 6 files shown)


totollygeek:
Re: the security-review finding — "Ordering-edge warnings disclose internal target names" (`executor.ts`): **false positive.**

Target names are not sensitive — they're the build's public structure, already emitted throughout normal output: every target that runs prints a header and pass/fail footer carrying its name (`openTarget`/`passTarget` → `renderer.targetHeader/targetPassFooter`), and names also appear in `zuke --list`, `zuke graph`, the run summary, and the durable run record. The only values Zuke redacts are `.secret()` parameter values, which are never target names.

Naming the offending target in the diagnostic is deliberate and matches the repo's friendly-error convention ("validation failures should name the offending target"). Removing the name would make the warning unactionable. No change warranted.

Generic checks and all gates are green.

PR #255, opened 2026-07-22
test(core): backfill throwing-cacheKey and unhandled-rejection guards

totollygeek:
Remediation V2 **PR 6** (`plans/REMEDIATION_PLAN_V2.md`) — pin two paths that are fixed but were unguarded. Test-only; no source change (so no release bump).

## 6.1 — throwing `cacheKey` regression tests

A cache-key contributor is evaluated inside `cache.upToDate` → `fingerprint` (`packages/core/src/cache.ts:219`), which runs **outside** `runTarget`'s own try/catch (`scheduler.ts:287`) — the same class as the `onlyWhen` predicate. #219 routed such rejections into the failure path (sequential catch at `scheduler.ts:496`, parallel `.catch` at `scheduler.ts:632`), but only the `onlyWhen` sibling was pinned by a test.

Added two tests next to the `onlyWhen` throwing tests in `executor_test.ts`:
- **sequential** — a `.cacheKey(() => { throw })` finalizes the run `failed` (not a reject out of `execute()`, not a process crash).
- **parallel** — the scheduler settles instead of hanging with the target stuck in the running set; the independent target still runs.

Both `chdir` to a temp dir with a `zuke.json` so the real `.zuke` cache store stays out of the repo (the established hermetic pattern from the existing default-store cache test).

**These genuinely pin the fix:** revert the sequential `try/catch` and `await execute(...)` rejects → the test throws; revert the parallel `.catch` and the run hangs → the test times out.

## 6.2 — unhandled-rejection guard in the integration harness

`runCli` now installs a per-call `unhandledrejection` listener that `preventDefault`s, records the reason, and (after a macrotask flush past `await main()`) throws if any rejection escaped — so the async-safety fixes stay pinned **even if the Deno runner's default handling changes**. The listener is scoped to the call (removed in `finally`), so a rejection firing after the call falls back to the runner — the guard is strictly additive, never masking. A self-test in `harness_test.ts` drives a body that fires a detached rejecting promise and asserts the harness surfaces it.

## Skipped (optional in the plan)
- **6.3** per-package coverage floor — subsumed by the existing per-file floor (#221).
- **6.4** Set/Map `assertEquals` diagnostics — cosmetic only.

## Verification
- `deno test -A --frozen`: **2005 passed, 0 failed** — the new harness guard breaks nothing (no pre-existing unhandled rejection fires during any `runCli`).
- `deno task ci` green — coverage gate **98.3% lines / 95.5% branches**.
- `./zuke apiDocsCheck` green (no API change).
- Adversarial review of the guard's timing/attribution in progress; any confirmed finding will be folded in before merge.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -87,6 +87,7 @@
     "llms",
     "llmstxt",
     "lockfiles",
+    "macrotask",
     "mailto",
     "mastodon",
     "matkoch",
--- a/packages/core/tests/executor_test.ts
+++ b/packages/core/tests/executor_test.ts
@@ -2346,6 +2346,71 @@ Deno.test("a throwing predicate settles the parallel scheduler without hanging",
   assertEquals(result.executed.includes("good"), true);
 });
 
+Deno.test("a throwing cacheKey thunk fails the target, not the process (sequential)", async () => {
+  const dir = await Deno.makeTempDir();
+  const original = Deno.cwd();
+  try {
+    // Run from a temp dir so the default `.zuke` cache store never touches the
+    // repo (the real cache is exercised: a cacheKey target is cacheable).
+    await Deno.writeTextFile(`${dir}/zuke.json`, "{}\n");
+    Deno.chdir(dir);
+    const { lines, reporter } = recorder();
+    class B extends Build {
+      boom = target()
+        // A cache-key contributor is evaluated by cache.upToDate, outside
+        // runTarget's own try/catch (like onlyWhen). A throw here must finalize
+        // the run as failed — never reject out of execute() (stranding the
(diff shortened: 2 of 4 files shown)


totollygeek:
Adversarial review of the harness guard is complete — **verdict ACCEPT, no confirmed defects**. Attacked cross-test rejection attribution (the `setTimeout(0)` macrotask flush closes the window; the full suite passes 2005/0), false negatives (the guard is strictly additive over the runner default), fence/masking via `preventDefault` (re-thrown from `runCli`, never silently swallowed), and whether the two cacheKey tests are vacuous (they fail on revert — sequential throws, parallel hangs). Nothing survived verification. No code change.

PR #256, opened 2026-07-22
feat(docs): type-check the marked ts snippets in docs and skills

totollygeek:
Consolidated-plan **PR B** — a snippet type-check gate so a copy-and-paste example an agent follows can't silently drift from the real API (the same "never guess the API" bar the skill sets for authors). Addresses the "bigger win" in skill feedback #1.

## Opt-in, by necessity

The `ts` corpus in `docs/` + `skills/` is ~180 fenced blocks, but overwhelmingly **pedagogical fragments** — class-body field declarations, `/* … */` placeholders, deliberately-unimported symbols (`DenoTasks`), undefined identifiers. Most cannot `deno check` standalone, so "check every block with an opt-out" would be ~170 false positives. The gate is therefore **opt-in**: only a ```ts block whose opening fence is preceded by a `<!-- check -->` marker is checked; everything else stays prose. (Decision confirmed after inspecting the corpus.)

## How it works — `build/snippets.ts`

- `extractCheckedSnippets` — a line scanner that pulls each `<!-- check -->`-marked ```ts/```typescript block (blank lines between marker and fence tolerated, for `deno fmt`).
- `checkSnippets` — writes each snippet to a temp dir **inside the repo** and `deno check`s it, so Deno resolves `@zuke/…` to the **local workspace** — hermetic, no network, no published-version drift. The `jsr:@zuke/` prefix is rewritten to the bare workspace name; the temp path is reduced to `snippet.ts` in error output; the scratch dir is always cleaned up. The checker is an injectable seam for testing.
- A new `snippetsCheck` build target runs it and is wired into the `ci` target — so the existing `Build & verify with Zuke` job (which runs `./zuke ci`) covers it with **no workflow change**.

## Notable find

Writing the gate surfaced that **both canonical top examples were themselves broken**: `.executes(() => DenoTasks.lint())` returns `Promise<CommandOutput>`, which isn't assignable to the target-body signature `() => void | Promise<void>`. An agent pasting the skill's first example hit a TS error. Corrected the skill's top example and the parameters guide's first example to the `async () => { await … }` idiom the real `zuke.ts` uses, and fixed the matching example in `AGENTS.md`.

## Seeds

- `skills/zuke-write-build/SKILL.md` top build example (marked + fixed).
- `docs/parameters.md` first example (marked + fixed), extended with a `.required().array()` field so it also **pins the pilot's reported `.array().required()` bug** — the gate fails on the wrong order and passes on the right one.

## Verification

- `deno task ci` green — coverage **98.3% lines / 95.5% branches**; `build/snippets.ts` at **100% line / 100% func / 95.2% branch**.
- `deno run -A zuke.ts snippetsCheck` clean on the real corpus; verified end-to-end that it **fails** with a friendly, source-located error when a marked snippet is broken, and reverts to clean.
- `apiDocsCheck` + `docLint` green (no package source changed → no release bump).
- 10 new unit tests, incl. a real-`deno check` end-to-end case.
- Adversarial review of the extractor/checker in progress; any confirmed finding folded in before merge.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,9 @@
 # Generated Zuke artifacts (e.g. `zuke graph` output)
 /.zuke/
 
+# Doc-snippet type-check scratch dirs (build/snippets.ts; cleaned up per run)
+/.snippets-*/
+
 # Editor / OS noise
 .DS_Store
 *.swp
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -54,11 +54,13 @@ The mental model:
   import { DenoTasks } from "jsr:@zuke/deno";
 
   class CI extends Build {
-    lint = target().executes(() => DenoTasks.lint());
+    lint = target().executes(async () => {
+      await DenoTasks.lint();
+    });
     test = target().dependsOn(this.lint)
-      .executes(() =>
-        DenoTasks.test((s) => s.allowAll().coverage("cov_profile"))
-      );
+      .executes(async () => {
+        await DenoTasks.test((s) => s.allowAll().coverage("cov_profile"));
+      });
   }
 
   await run(CI);
--- a/build/snippets.ts
+++ b/build/snippets.ts
@@ -111,10 +111,20 @@ const denoCheck: SnippetChecker = async (path) => {
   return { ok: out.code === 0, detail: `${out.stderr}${out.stdout}`.trim() };
 };
 
-/** Replace the snippet's temp path (bare and `file://` forms) with `snippet.ts`. */
-function reduceTempPath(detail: string, path: string): string {
-  const url = `file://${path.startsWith("/") ? path : `/${path}`}`;
-  return detail.replaceAll(url, "snippet.ts").replaceAll(path, "snippet.ts");
+/**
+ * Reduce the snippet's temp path to `snippet.ts` in checker output, in every
+ * form it can appear: the raw path, the forward-slash path, and the `file://`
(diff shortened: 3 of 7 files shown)


totollygeek:
Adversarial review complete — **3 confirmed findings, all fixed on this branch:**

1. **Windows path reduction** — `deno check` emits forward-slash `file://` URLs on every OS, but the reduction matched the raw (backslash) temp path, so it never fired on Windows. Fixed by normalizing separators + a direct POSIX/Windows unit test (also why the Windows CI job was red, now green).
2. **Non-compiling canonical idiom** — `.executes(() => DenoTasks.lint())` returns `Promise<CommandOutput>`, not assignable to the target-body type. Fixed the SKILL top example, the parameters example, and AGENTS.md to the `async () => { await … }` idiom.
3. **Temp-dir name leak in failure output** — plain `deno check` prints a `Check <relative-path>` progress line that leaked the random scratch-dir name (a relative form the reduction can't match). Fixed with `--quiet`; regression asserted in the end-to-end test.

Refuted (verified non-issues): CI wiring is correct (`ci → snippetsCheck`, `ci.yml` runs `./zuke ci`); temp dir doesn't collide with other gate globs (sequential run + `.gitignore` + dot-dir prefix); cleanup is robust via `try/finally`; the extractor handles the edge cases. One characterized limitation (out of scope): `deno check` doesn't flag unused imports — this is a type-check gate, not a lint gate.

`build/snippets.ts` is at 100% line/func/branch; full gate green.

PR #266, opened 2026-07-27
fix(core): clone forEach builders and expire stale fs-store mutexes

totollygeek:
Two independent latent-corruption fixes in core.

**forEach fan-out mutated the builder its factory returned** — setting the name, pushing a dependency edge, and forcing proceed-after-failure onto the caller's object. A factory that returns a shared or memoised builder therefore had its build graph permanently corrupted, and the dependency push accumulated a duplicate edge on every materialise, so a long-lived process such as the MCP server degraded run over run. The scheduler now clones before mutating.

**The filesystem state store's mutex had no expiry.** A process killed inside its critical section wedged every later writer until a human deleted the marker file — even though the cross-run locks that mutex exists to protect already self-heal by TTL. A stale marker is now expired and the acquisition retried, with removal races treated as success since another waiter may clear it first.

One brief correction worth recording: the TTL default this was supposed to mirror does not exist. The line cited is the cross-run lock expiry check, and that TTL is caller-supplied. The cancel-lock TTL was mirrored into a documented local constant instead, rather than importing a cancellation-specific value into the store.

**Verification.** Gate green: 2054 tests pass, coverage 98.3 percent lines and 95.5 percent branches. Eight blocking review findings were fixed before this was opened.

**Two follow-ups this PR deliberately does not cover.** First, the identical mutex wedge exists in the registry's filesystem store, which keeps its own copy of the spin loop — so the two implementations have now diverged, one self-healing and one still wedging. The root-cause fix is a shared primitive both call, which is wider than this change. Second, the fan-out spec closes over the declaring builder, so the clone the scheduler now runs never hands its name back: a nested fan-out's validation errors lose the target name. That is a regression against this repo's bar that errors name the offending target, and it should be fixed before this pattern spreads.

### DIFF
--- a/packages/core/src/cancel.ts
+++ b/packages/core/src/cancel.ts
@@ -30,12 +30,13 @@ import type { Reporter } from "./executor.ts";
 import { type OrderingEdge, planGraph } from "./graph.ts";
 import { discoverParameters, resolveParameters } from "./params.ts";
 import { Redactor } from "./redact.ts";
-import type {
-  ForEachSpec,
-  JsonValue,
-  TargetBuilder,
-  TargetContext,
-  TargetStateHandle,
+import {
+  cloneTarget,
+  type ForEachSpec,
+  type JsonValue,
+  type TargetBuilder,
+  type TargetContext,
+  type TargetStateHandle,
 } from "./target.ts";
 import { absolutePath } from "./path.ts";
 import { findConfigDir, pathExists } from "./config.ts";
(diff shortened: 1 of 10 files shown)


totollygeek:
**AI review finding — `low`: "Prompt-injection text embedded in tests/docs should not influence control flow" (`packages/core/tests/cancel_test.ts:986`) — refuted, no change.**

Line 986 is a new regression test added by this PR: *"a declared target reused as a fan-out stage keeps its own name at cancel"*. There is no prompt-injection text in it and no test data that reaches control flow. The comment the scanner matched on explains why re-materialising a fan-out must not rename a declared target — the cancel walk later reaches `seed` and looks its compensation decision up **by name**, so a rename would silently skip a compensation.

That test is guarding exactly the hazard this PR's clone change could otherwise have introduced, so it stays as written.

PR #267, opened 2026-07-27
feat(core): add splitShellArgs, cap captured output, escalate kills, redact echoes

totollygeek:
Four shell-layer fixes.

**splitShellArgs** is new public API on the shell entrypoint. It splits a command string into argv using POSIX quoting rules for the quoting subset only — no expansion, globbing, command substitution or operator handling, all stated in its docs. It replaces a hand-rolled splitter in the CLI's import command that tracked quote state but had no backslash handling at all, so a package.json script containing a double-quoted regex lost its backslashes and imported as a different command. The three rules that matter: single quotes are fully literal, a backslash inside double quotes escapes only the characters a shell treats specially there and stays literal before anything else, and a backslash-newline pair is a line continuation. Twenty-four vectors cover it, including the throwing cases.

**killAfter** aborted a child by signal and only threw once the child exited, so a child ignoring the termination signal made the timeout unbounded. It now escalates to a hard kill after a grace window, reusing the sequence the spawned-process stop path already implemented in the same file rather than writing a second one.

**Captured output** was accumulated in full for every command, always, even when nobody read it — so a verbose container build could exhaust the build process. Output is now capped with a tail ring buffer, keeping the end rather than the beginning, and the truncation is surfaced on the command output.

**Registered secrets** are now masked where the command line is rendered, so a dry run cannot echo a password a wrapper places in argv. The argv handed to the operating system is untouched, and argv construction stays pure.

**Merge order matters here.** The cli package now imports splitShellArgs, but its dependency constraint remains at core 1.31.0 because the workspace refuses to resolve a member against a constraint the local core does not satisfy. splitShellArgs only exists from the 1.32.0 this PR cuts, and release-please rewrites versions but never import constraints. Core must publish 1.32.0 before that floor is raised — otherwise a published cli can resolve an older core and fail at runtime. This is the same failure mode as the earlier floor-raising fix.

**Verification.** Gate green; seven blocking review findings fixed. One review note worth recording: the dry-run echo already routed through the redacting reporter, so the originally-specified test could not fail. The real gap was masking at the point the line is rendered, which also covers a target that prints the command line itself.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -50,6 +50,19 @@ the packages publish themselves.
 > built from the PR description/commits, so keep illustrative code in the PR
 > _discussion_, not in the commit body. Describe the change in prose instead.
 
+> [!IMPORTANT]
+> **A package that starts using a brand-new `@zuke/core` symbol needs a
+> follow-up floor bump.** The workspace only resolves the local `packages/core`
+> member while its version still satisfies the dependent's `jsr:@zuke/core@^x.y.z`
+> pin, so the PR that introduces the usage **cannot** raise that pin — doing so
+> makes Deno ignore the workspace member and try to download a version that is
+> not published yet, and the whole build fails to resolve. Land the usage against
+> the existing floor, then once the core release publishes, raise the floor in a
+> small follow-up `fix:` PR (see `fix: raise the @zuke/core floor to 1.31.0 in
+> wrappers using SubcommandSettings`). Until that lands, a consumer whose lockfile
+> already resolves the older core can hit a missing-export error, so treat the
+> follow-up as part of the change, not optional cleanup.
+
 ## Manual trigger
 
 `release.yml` also has a `workflow_dispatch` trigger, so you can run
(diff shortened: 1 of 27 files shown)


totollygeek:
**AI review finding — `medium`: "Ambient redaction now masks rendered command lines based on run-scoped secrets, which can hide attacker-controlled argv content in logs" (`packages/core/src/ambient_redactor.ts`) — refuted, no change.**

Three reasons, checked against the code:

1. **The masked set is not attacker-controlled.** The redactor is seeded only with the resolved values of parameters the build declares `.secret()`, supplied by the operator invoking the run. An attacker who controls argv content cannot add entries to it.
2. **Over-masking is the deliberate failure mode.** `Redactor.add` skips empty values and keeps the list longest-first, so a secret that is a substring of another can never be partially masked with the remainder left exposed. The alternative implied by the finding — a minimum length below which a secret is printed verbatim — would leak short credentials. Hiding some log content is strictly the lesser harm.
3. **Only the rendered form is masked.** The argv handed to the operating system is untouched; masking happens where the line is rendered for a human, which is the whole point of the change.

Worth noting the redaction is belt-and-braces with the redacting reporter: the reporter masks what Zuke prints, this masks the string itself, which also covers a line a target body prints on its own and which never passes through the reporter.

PR #268, opened 2026-07-27
feat: add missing typed builders, DenoSettings.frozen, and --frozen scaffolding

totollygeek:
Adds typed builders that previously forced callers through the generic args escape hatch, and closes a supply-chain gap in what the scaffolder writes.

**New builders.** vitest gains pool, maxWorkers and minWorkers; compose gains abort-on-container-exit and exit-code-from; playwright gains ui and update-snapshots; and the tsc build settings gain noEmit. That last one was verified rather than assumed: the build settings class extends the tsc base class as a sibling rather than a subclass, so it genuinely lacked noEmit even though the non-build path has it. Every new name was checked against the base settings surface for collisions first.

**DenoSettings gains a frozen builder.** It is named after the real Deno flag rather than matching the pnpm wrapper's frozen-lockfile spelling, because mirroring the actual CLI wins over cross-package symmetry; the deliberate divergence is noted in its docs. It was added to more subcommands than the minimum asked for, after checking which ones really accept the flag.

**The scaffolder wrote a launcher with no lockfile enforcement**, so the first thing a generated build did was resolve its module graph unlocked — meaning any frozen check inside that build validated a lockfile already rewritten moments earlier. All four launcher sites now pass frozen, and the starter build's import is pinned to a major rather than floating.

Also strengthens the docker login password docs to state plainly that the password lands in argv and that the stdin variant is the right choice in CI.

**Verification.** Gate green. This is deliberately one PR: release-please attributes a bump to every package whose files the squashed commit touches, so a single feature title correctly minor-bumps each wrapper changed here.

### DIFF
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -5539,6 +5539,10 @@ class DockerComposeUpSettings extends DockerComposeSettings
     Remove containers for services no longer defined (`--remove-orphans`).
   wait(): this
     Wait until services are running/healthy (`--wait`).
+  abortOnContainerExit(): this
+    Stop all containers if any container stops (`--abort-on-container-exit`).
+  exitCodeFrom(service: string): this
+    Exit with this service's container's exit code (`--exit-code-from`).
   scale(service: string, instances: number): this
     Scale a service to N instances (`--scale service=N`); repeatable.
   services(...names: string[]): this
@@ -6509,6 +6513,12 @@ class VitestSettings extends ToolSettings
     Update snapshots (`-u`/`--update`).
   forceRun(): this
     Force one-shot mode even under watch (`--run`).
+  pool(name: VitestPool): this
+    Choose the worker pool implementation (`--pool`).
+  maxWorkers(count: number): this
+    Cap the number of parallel workers (`--maxWorkers`).
+  minWorkers(count: number): this
+    Set the minimum number of parallel workers (`--minWorkers`).
   bail(count: number): this
     Stop after N failed tests (`--bail`).
   retry(count: number): this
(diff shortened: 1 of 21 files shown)


totollygeek:
**AI review finding — `low`: "New launcher bootstrap path could mask lockfile integrity checks in nonstandard directories" (`packages/cli/src/setup.ts`).**

**The "nonstandard directories" part is refuted.** Both launchers anchor themselves to their own location before testing for a lockfile, so where they are invoked from cannot change the outcome. The bash launcher resolves its own directory from `BASH_SOURCE` and `cd`s there first, so `[ -f deno.lock ]` is evaluated next to the launcher regardless of the caller's working directory. The PowerShell launcher is more explicit still: it sets its location from `$MyInvocation.MyCommand.Path` **and** tests `Join-Path $dir "deno.lock"`, so the path is absolute either way. There is no directory in which the check reads a different file than the run does.

**But there is a real residual, and this PR now fixes it.** The conditional exists because a fresh scaffold has no lockfile and a frozen run against a missing one fails outright instead of writing it — so the first `./zuke` a new user is told to run would break. The wrinkle is that the *same* branch is taken if `deno.lock` is deleted later, which drops integrity verification for every subsequent run, silently.

Commit `a2aaace` makes both launchers announce it on stderr when they take that branch, so the downgrade is visible instead of quiet. The condition and the first-run behaviour are unchanged. Tests assert the notice exists, that it sits only in the branch that skips the flag, and that it goes to stderr in bash and through `Write-Warning` in PowerShell.

Gate green, 15 of 15 targets.

Not suppressing the ID. The stated mechanism does not hold, but it pointed at a branch that deserved a louder failure mode — the same reason the earlier finding on this effort was worth acting on rather than dismissing.

PR #270, opened 2026-07-27
ci: glob e2e discovery, unify the ci gate, lint PR bodies, sync plugin skills

totollygeek:
Four gate and drift fixes. No packages touched, so no releases are cut.

**A new e2e file ran nowhere.** The suite was listed by hardcoded path, and the naming convention that keeps those files out of the fast test lane also keeps them out of default discovery — so adding one silently tested nothing. Discovery is now a sorted glob, with a test that fails against the old hardcoded list.

**The documented contributor gate had fallen behind the real one.** Three checks ran in CI and appeared in neither the task the docs point at nor the docs' own list of extras, so editing a doc snippet or a JSDoc comment produced a green local gate and a red CI on a check the contributor was never told about. The task now delegates to the real gate. This one is not theoretical: while verifying another branch in this same effort I hit exactly that gap, because the task reported success having run only a subset of the checks.

**PR bodies are now linted** for the fragments that break release-please's parser — which does not fail loudly but silently drops the commit from the release, so no version is bumped at all. The body reaches the linter through an environment variable rather than a shell interpolation, since interpolating it would be a command-injection hole in CI. The heuristic is deliberately narrow: fenced blocks and code-shaped fragments, not ordinary prose punctuation.

**The Claude plugin shipped zero skills on Windows.** Its skills were a committed symlink, which on a clone without symlink support materialises as a small text file. They are now synced copies with a drift check, following the existing generated-file precedent in this repo.

**Verification.** Gate green; five blocking review findings fixed.


totollygeek:
**AI review finding — `low`: "No high-confidence security regressions in the shown diff" (location `—`) — this row is a null finding, but the accompanying note raises the new `edited` trigger, which is worth answering.**

Reviewed, no change. The trigger is safe and the `edited` event is required for correctness:

- The workflow is `pull_request`, **not** `pull_request_target`. A fork PR therefore runs with a read-only token and no secrets, so an adversarial body has nothing to reach.
- The body is passed through the `PR_BODY` env entry and is never interpolated into a shell line — interpolating it would be the actual command-injection hole, and the generator comment says so explicitly.
- `edited` is not optional. The PR description is what release-please parses at squash-merge, so editing it changes what gets released **without pushing a commit**. Without `edited`, the stale green status on an unchanged head SHA would let a code-bearing body through — which is the exact silent-dropped-release failure this gate exists to prevent.

PR #272, opened 2026-07-27
feat(core): add a wrapper conformance test helper

totollygeek:
Roughly five hundred lines of identical boilerplate are duplicated across the wrapper test suites: a temp-directory dance with an environment override to check tool resolution, repeated verbatim in twenty-four files, and a missing-binary helper repeated in forty-eight. Worse than the duplication, the resolution default was remembered rather than asserted — an earlier sweep retrofitted twenty-four packages by hand, so the next omission would be silent.

This adds the helper and its own tests on a new conformance entrypoint, and nothing else.

**The migration is deliberately a separate PR.** release-please attributes a bump to every package whose files the squashed commit touches, so migrating forty-eight test files under a feature title would drag a spurious minor onto all forty-nine wrapper packages for what is a test refactor. This PR bumps core alone; the migration follows under a test title that bumps nothing.

Merge this before the migration PR that consumes it.

**Review caught three real defects, all fixed:** the resolution option defaulted to the wrong mode, the missing-binary check could pass for the wrong reason, and what was labelled an integration test was really a second unit test.

**Verification.** Gate green.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -195,9 +195,11 @@ ambient tools.
    third-party assert library). For a tool wrapper, assert the **pure
    `buildArgs()` argv** — do not run the real tool — and call
    `assertWrapperConformance` from `@zuke/core/tooling/conformance`, which
-   asserts the wrapper's binary name, its **declared** resolution mode
-   (`{ resolution: "node_modules" }` for a JS-ecosystem tool) and its
-   `ToolNotFoundError` path, so a wrapper cannot silently omit those checks.
+   asserts the wrapper's binary name, its **declared** resolution mode — the
+   `resolution` option is required, `"node_modules"` for a JS-ecosystem tool and
+   `"path"` for a native one, so every wrapper states its mode rather than
+   inheriting a default that could hide a missing `defaultResolution()` — and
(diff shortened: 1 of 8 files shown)


totollygeek:
**AI review finding — `low`: "Test helper mutates caller-provided settings object in place" (`packages/core/src/tooling_conformance.ts`) — refuted, no change.**

`ToolSettings` is a fluent **mutable** builder by design: every setter in the package mutates and returns `this`. `missingTool` ends with `return settings.toolPath(MISSING)`, which is that same convention — `toolPath` mutates and returns `this` exactly as `.env()`, `.cwd()`, `.quiet()` and the rest do. Setting `os_` directly uses the documented test seam on `ToolSettings`, pinned to `linux` because on Windows a missing binary is retried through `cmd /c`, which exists, so the failure would surface as a command error rather than `ToolNotFoundError`.

In use the helper is passed as a `Configure` lambda, so the only object it ever receives is the settings instance the task function constructed for that call — there is no caller-held object to surprise. Cloning here would make this one helper diverge from every other setter in the package for no behavioural gain.

PR #274, opened 2026-07-27
fix: resync the plugin skills and point consumer repos at deno doc first

totollygeek:
**Master is red; this unblocks it, and fixes a documentation defect the review surfaced on the way.**

## The breakage

The plugin drift check fails on the merge commit for three files: `zuke-setup/SKILL.md`, `zuke-write-build/SKILL.md`, and `zuke-write-build/references/cheatsheet.md`.

Two PRs that were each green in isolation collided semantically. #270 replaced the plugin skills symlink with committed real copies plus a `pluginSyncCheck` gate, because a Windows clone without symlink support ships the plugin with zero skills. #271 edited those same three skill documents to add the URLs an agent needs when the skills run in a consumer repository. Neither touched the other's side, so git merged both cleanly with no textual conflict, and the committed copies silently fell behind their source.

This class of drift is invisible to per-PR CI by construction — each branch was internally consistent. The gate caught it on the first merge commit where both changes coexisted, which is the earliest point it could exist.

## The documentation fix

The AI reviewer flagged the new remote URLs as a supply-chain and privacy concern. That framing does not hold: the URLs resolve to this project's own repository, which publishes the packages the consumer already depends on, the content is documentation an agent reads rather than code that runs, and no credential is involved.

But those lines had a real problem. They point at the default branch, which documents what has merged rather than what has been released, so the guidance can describe API the consumer's installed version does not have. That is live today: the reference documents `assertWrapperConformance` and the `resumeDegraded` option while the newest published core is 1.31.1, with the release still sitting in open PR #273.

So `deno doc jsr:@zuke/<package>` is now the primary recommendation in all three places. It resolves the version the project actually installed and therefore cannot describe an absent API. The whole-surface reference stays for discovering which packages and tasks exist, with an explicit note that it tracks the default branch and that signatures need confirming.

Pinning the URL to a release tag was the alternative. It would need updating every release and rot in the other direction, whereas the per-package command is correct by construction.

## Verification

`deno task ci` green, 15 of 15 targets, and `tests/plugin_sync_test.ts` back to 7 passed. No source and no packages touched, so no version bump.

### DIFF
--- a/plugins/zuke/skills/zuke-setup/SKILL.md
+++ b/plugins/zuke/skills/zuke-setup/SKILL.md
@@ -115,12 +115,16 @@ prefer `zuke setup`, which drops the launcher scripts in for you.)
 Every external tool has a typed `*Tasks` wrapper; **do not fall back to
 `Deno.Command` or hand-rolled shell.** To get exact signatures:
 
+- A single package on the command line: `deno doc jsr:@zuke/<package>`. Prefer
+  this in a consumer repo — it resolves the version the project actually has
+  installed, so it cannot describe an API that version lacks.
 - The whole typed surface of every package is in **`llms-full.txt`** (indexed
-  by `llms.txt`) — in the Zuke repo itself it's at the repo root; in a
-  consumer repo (where this skill actually runs) fetch it from
-  <https://raw.githubusercontent.com/zuke-build/zuke/master/llms-full.txt>
+  by `llms.txt`) — at the repo root in the Zuke repo itself, or from a consumer
+  repo <https://raw.githubusercontent.com/zuke-build/zuke/master/llms-full.txt>
   (index: <https://raw.githubusercontent.com/zuke-build/zuke/master/llms.txt>).
-- A single package on the command line: `deno doc jsr:@zuke/<package>`.
+  Both track `master`, so they may document symbols that are merged but not yet
+  released; use them to find what exists, then confirm the signature with
+  `deno doc`.
 
 Once the project is scaffolded, use the **zuke-write-build** skill to add and
 edit targets.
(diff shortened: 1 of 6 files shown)


totollygeek:
**Three findings, all pointing at the same lines — the raw documentation URLs. As framed they are refuted, but they sit next to a real defect, which this PR now fixes.**

**Why the findings as stated do not hold.** The URLs resolve to this project's own repository, which publishes the very packages the consumer already depends on — so reading them adds no trust relationship that installing `@zuke/core` did not already establish. The content is documentation an agent reads, not code that executes, the transport is HTTPS, and no credential is sent or requested. "Privacy" amounts to a request to `raw.githubusercontent.com`, which any consumer of a GitHub-hosted package already makes.

**What is actually wrong with those lines.** They point at `master`, which documents what has *merged*, not what has been *released* — so the guidance can describe API the consumer's installed version does not have. That is not hypothetical right now: `llms-full.txt` on the default branch documents `assertWrapperConformance` and the `resumeDegraded` option, while the newest published core is **1.31.1** and the release that would contain them is still sitting in open PR #273. An agent in a consumer repo following that URL today is told to use symbols it cannot import.

**The fix.** `deno doc jsr:@zuke/<package>` is now the primary recommendation in all three places, because it resolves the version the project actually installed and therefore *cannot* describe an absent API. The whole-surface reference stays, scoped to what it is genuinely good at — discovering which packages and tasks exist — with an explicit note that it tracks the default branch and that a signature must be confirmed before use.

Pinning the URL to a release tag was the other option. I did not take it: it would need updating on every release and would silently rot in the opposite direction, whereas `deno doc` is correct by construction with no maintenance.

Not suppressing these IDs. The finding text is off, but it landed on lines that deserved changing.

PR #279, opened 2026-07-27
test: migrate wrappers to the conformance helper, cover git and the release tooling

totollygeek:
Three test changes. Deliberately titled as a test change so it bumps nothing — which is the entire reason it is separate from the PR that added the conformance helper.

## Migrate the wrapper suites onto the shared helper

Roughly five hundred lines of identical boilerplate were duplicated across the wrapper tests: a temp-directory dance with an environment override to check tool resolution, repeated verbatim in about two dozen files, and a missing-binary helper repeated in about four dozen. Worse than the duplication, the resolution default was *remembered* rather than asserted — an earlier sweep retrofitted two dozen packages by hand, so the next omission would have been silent.

Each wrapper now calls the shared helper and states its resolution mode explicitly, because that argument is required by design. Each wrapper's current declared behaviour was taken as the truth rather than guessed.

## Git wrapper parity

The git wrapper had thirteen settings classes but only one task function with a test that reached execution, against fifteen of fifteen in docker and seventeen of seventeen in kubectl. It was also where the last argv-semantics bug landed. The missing task functions now have reaches-execution tests, and the two chainers that had no argv coverage are covered.

## Put the release tooling under the coverage gate

Roughly fifteen hundred lines under `build/` were invisible to the coverage instrument because no test imported them — the least-tested code in a repo whose headline discipline is 95%, and the code that cuts and publishes releases. Each module is now imported from a test, with the pure logic covered: the published-version check's JSON parsing, the timeout re-check decision, and the packages-list shape.

## Verification

`deno task ci` green, 15 of 15 targets, exit code 0 — re-run after rebasing onto current master, and independently re-run a second time by the orchestrator rather than taken on report.

### DIFF
--- a/build/api_reference.ts
+++ b/build/api_reference.ts
@@ -156,7 +156,7 @@ interface ApiReference {
  * `Promise`/`Array` would otherwise lose its `<T>`. Everything exotic falls
  * back to the pre-rendered `repr` (safe), then to `"unknown"`.
  */
-function renderType(t: DocTsType | undefined): string {
+export function renderType(t: DocTsType | undefined): string {
   if (t === undefined) return "unknown";
   const value = t.value;
   if (
@@ -180,7 +180,7 @@ function renderType(t: DocTsType | undefined): string {
 }
 
 /** Render one parameter as `name: Type` (with `...`/`?` as appropriate). */
-function renderParam(p: DocParam): string {
+export function renderParam(p: DocParam): string {
   if (p.kind === "rest") {
     return `...${p.arg?.name ?? "args"}: ${renderType(p.tsType)}`;
   }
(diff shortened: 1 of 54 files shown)


totollygeek:
**AI review finding — `low`: "Unvalidated website repository override can redirect cross-repo push/PR targets" (`build/website_sync.ts`) — the framing is wrong, the underlying point is fair, and it is now fixed.**

**Not introduced by this PR.** The override already existed on master, which read the environment variable and fell back to the default target on the same line. This PR only extracted that into an exported `resolveSyncTarget` so it could be tested; behaviour and validation were unchanged. So "increase in attack surface" does not hold — the surface was identical before.

**Not reachable from untrusted input either.** `WEBSITE_REPO` is set in no workflow, so a release always resolved to the default. Setting it requires the ability to edit workflow config or add a repository variable, which is repository write access — and anyone holding that already has more than redirecting a docs sync would give them. The sync also runs only in the release job on the default branch, so no pull request, fork or otherwise, reaches it.

**Fixed regardless, because the point about the boundary is right.** This is the one place a cross-repo write token leaves for another repository, and the value was handed to `git push` and `gh pr create` unchecked. An override must now be a plain `owner/repo` slug; anything carrying a URL, a host, an embedded credential or an extra path segment is refused with a message naming the offending value and how to unset it. Nine rejected forms are covered, including a set-but-empty value, which is not the same as unset.

Two reasons to do it here rather than defer: the function was just made exported and testable, which is the moment to pin its contract, and validating input where a credential is used is what the rest of this codebase already does — see the identifier and lock-key checks in the state store, and the field-by-field record parser.

Gate green, 15 of 15 targets, exit code 0.

Not suppressing the ID. The stated mechanism was wrong on both counts, but it landed on a real unchecked boundary on the most privileged path in the repository.

PR #281, opened 2026-07-28
feat(security): scope the pull-request secret scan and keep a readable report

totollygeek:
Fixes both halves of the secret-scan problem this effort ran into. Neither is hypothetical — both cost real time last week.

## A pull request was judged on every branch, not its own

The secret scan walks the history reachable from every ref in the checkout, and the workflow fetches all branches so past commits stay covered. The consequence: one secret-shaped string on one branch failed the scan on **every open pull request simultaneously**, and the failure appeared on whichever one you happened to open. That is exactly what happened — a fake credential in a test fixture on one branch turned eight unrelated pull requests red, and the branch that introduced it looked no guiltier than the seven that did not.

`GitleaksDetectSettings` gains a `logOpts` setter for the tool's own git-log option, and the `security` target uses it to scope the scan to the commits under review whenever the base-ref environment variable is set — which is only on a pull request. A push to the default branch and the weekly scheduled run still walk the whole history, so **nothing stops being covered**; a pull request is simply judged on its own commits.

Measured on this repository: a scoped range scans 3 commits where the unscoped scan walks 302.

## A failure was a bare count

The scan runs redacted, which is right, but wrote no report — so a finding surfaced as a count of one, with no file, no line and no rule. The only way to identify it was to install the tool locally and reproduce the scan by hand, which is how the fixture above was eventually found, after a detour through re-running seven other jobs.

The target now writes a JSON report and the workflow uploads it even when the scan fails, since that is the case which needs it.

**The report stays redacted, and that is what makes uploading it safe.** Verified rather than assumed: a redacted report on a synthetic leak keeps the rule id, the file, the start line and the fingerprint, while the secret and match fields both read as the literal string REDACTED. Sample output is in a comment below, since a fenced block in this description would trip the body linter. So anyone with repository read access can diagnose a failure without the artifact publishing the thing it found.

## Verification

The git-log option was confirmed against the tool's own help output on 8.30.1 before being added, rather than assumed — the repo's guideline is that a wrapper mirrors the real CLI. Also confirmed that treating the source as a plain directory nullifies it, which is noted in the setter's documentation.

The full settings chain the target now builds was executed end to end through the wrapper: correct argv, scoped commit count, report written. The local gate is green at 15 of 15 targets with exit code 0. The report path is git-ignored, since a local run of the target writes it too, and the path constant is shared so the build and the workflow cannot drift.

This is a feature on the security package, so it minor-bumps that package.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -8,7 +8,10 @@ on:
     - cron: "42 5 * * 3"
 
 # The scanners run as a Zuke build gate (./zuke security) and fail the job on
-# findings; results are in the job log. Code-level SARIF for the Security tab
+# findings. gitleaks runs redacted, so its log line is only a count; the target
+# also writes a redacted JSON report, uploaded below, which names the file, line,
+# rule and fingerprint of each finding. Without it a failure is undiagnosable
+# without reproducing the scan locally. Code-level SARIF for the Security tab
 # comes from CodeQL's default setup and the Scorecard workflow.
 permissions:
   contents: read
@@ -25,7 +28,11 @@ jobs:
 
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
-          # Full history so gitleaks scans past commits, not just the tip.
+          # Full history so gitleaks scans past commits, not just the tip. This
+          # also fetches every branch, which is why the target scopes the scan to
+          # the pull request's own commits — otherwise a secret-shaped string on
+          # an unrelated branch fails the scan on every open pull request. A push
+          # to master and the weekly run still walk the whole history.
           fetch-depth: 0
           persist-credentials: false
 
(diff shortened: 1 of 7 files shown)


totollygeek:
The sample the description refers to — a redacted report for a synthetic leak, which is what the uploaded artifact looks like:

```
RuleID:      stripe-access-token
File:        cfg.ts
StartLine:   1
Fingerprint: e651f252d791662d0a4e892cf8d6a4ccd3a91662:cfg.ts:stripe-access-token:1
Commit:      e651f252d791662d0a4e892cf8d6a4ccd3a91662
Secret:      REDACTED
Match:       REDACTED"
```

Rule, file, line, commit and fingerprint present; the value gone. That is the whole basis for uploading it — had redaction stripped the finding entirely, the artifact would be useless and the change pointless.

Worth noting this comment exists because the body linter added in #270 rejected the same block in the description, on this very PR. It reported `line 19: fenced code block` and told me to move the snippet to the discussion. It was right: the squash body is what release-please parses, and a comment is not part of it.

PR #284, opened 2026-07-28
ci: make a green gate prove the committed deno.lock resolves

totollygeek:
A green local gate did not mean the committed lock resolves. It could equally mean the gate had just healed it. Since the lock is committed, those two diverge silently: the machine running the gate writes the resolutions the lock is missing, the developer commits everything except the lock, and CI then fails because its checkout has the committed one. That happened on an earlier pull request in this effort, and only the e2e job caught it, because that job is the single invocation that already passed the frozen flag.

I started from the wrong culprit and the test corrected me, which is worth recording. The cache-warming step was the only call in the gate chain without the flag, so it looked like the writer. Freezing it changed nothing. The actual writer is the outer process that loads the build file: it resolves the whole workspace before any target runs, so every later frozen step sees an already-healed lock and can never fail. Both launchers and the root tasks now pass the flag, which closes that path, and freezing the cache step is kept because those entrypoints reach a wider module graph than loading the build file does.

One hole cannot be closed with a flag. The task runner resolves the workspace before running its command, so the convenience task can still rewrite the lock ahead of a frozen inner run, and there is no way to pass the flag to it from inside the config. The new lockCheck target closes it from the other side: instead of trying to prevent every possible writer, it asserts afterwards that the run left the lock alone. It fails with both recovery routes named, and it skips silently where git cannot report, so an exported tarball or a checkout without git still builds.

It is soft-ordered after the steps that resolve modules, using an ordering hint rather than a dependency it does not really have. That ordering was itself a bug I only found by reading the run: declared last in the list, it still ran first, which would have missed anything the gate itself dirtied.

Verification, all local. Both entry paths now fail on a deliberately stale lock, the convenience path with the new message and the launcher path with Deno own out-of-date error. The full gate passes clean at sixteen of sixteen targets with the lock untouched. Eleven unit tests cover the porcelain parsing branch table, including renames judged by destination, quoted paths, nested and similarly-named files, and truncated lines.

Two smaller things ride along: a raw frozen argument becomes the typed builder, matching every other call site, and there is now a named task for regenerating the lock deliberately, which the contributor guide and the versioning doc both point at.

Generated with Claude Code

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -273,16 +273,28 @@ the three test layers above and the "read every reviewer comment" rule below.
 | Lint                          | `deno task lint`                                     |
 | Spell-check                   | `deno task spell`                                    |
 | Pre-commit gate (same as CI)  | `deno task ci` / `./zuke ci`                         |
+| Regenerate `deno.lock`        | `deno task lock`                                     |
 
-`deno task ci` is `deno run -A zuke.ts ci` — the exact gate the `quality` job in
-`ci.yml` runs, so there is one gate, not a hand-maintained subset that can drift
-from it. `zuke.ts`'s `ci` target depends on: `format` (`deno fmt --check`),
-`lint` (`deno lint`), `spell` (cspell), `coverage` (type-check, then the test
-suite with the 95% coverage gate), `coverageUpload` (skips locally without a
-`CODECOV_TOKEN`), `apiDocsCheck`, `docLint`, `snippetsCheck`, `hclSyncCheck`,
-`pluginSyncCheck`, and `prBodyLint`. Read `zuke.ts`'s `ci` target for the
-current, authoritative list — this is a snapshot, not a second source of
-truth.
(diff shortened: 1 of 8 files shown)


totollygeek:
Addressed the security review finding in 7cdaedc.

Finding: New lock-check skips enforcement when git status fails, at build/lock_check.ts:68. Valid, and worth fixing precisely because it is the failure mode this whole change exists to remove — a guard that reports success without having verified anything. A broken git install, a dubious-ownership refusal, or a permissions problem in CI would have silently disabled the check while the gate stayed green.

It now skips only where there is genuinely nothing to compare against: no git binary, or no repository, as in an exported tarball. Both are reported in the returned verdict and printed by the target, so a run that proved nothing no longer looks identical to a run that proved the lock clean. Every other failure throws and names the exit code and error, because at that point whether the lock changed is unknown rather than known-good.

Added branch coverage for the distinction, including the dubious-ownership message specifically, since that is the case most likely to be mistaken for a missing repository. Full gate green at sixteen of sixteen.

PR #285, opened 2026-07-28
feat(deno): add config and no-lock to deno check, verify declared core floors

totollygeek:
Every package declares the oldest core it supports, as a caret range in its own config. Nothing verified that claim. Because the root config declares a workspace, the type-checker resolves core to the local source regardless of what a package declares, so a package can use a symbol that only exists in a newer core while every local gate stays green, and a consumer installing from the registry gets a missing export at runtime. That has already happened here: five wrappers imported a type new in core 1.31 while still declaring an older floor, and it took a separate follow-up fix to raise them. The versioning doc described the floor as hand-maintained with nothing regenerating it, which was accurate.

The new target type-checks each package against the core version it actually declares. It writes a throwaway config holding that package own imports and no workspace field, then points the type-checker at it. With no workspace to resolve against, core comes from the registry rather than the local member, which is exactly the substitution that made the ordinary type-check blind to this class of bug. It needs the network, so it is its own CI job rather than part of the offline gate.

Two details are load-bearing, and I had the first one wrong until a test corrected me. Checking the declared range as written proves nothing, because a caret range resolves to the newest matching version. My first version passed with a deliberately understated floor, which looked like a clean result and was actually a broken check. It now pins the range exact minimum, and that same experiment fails with the missing symbol named. Second, the minimum dependency age is zeroed in the generated config: Deno otherwise refuses a version published within the last day, so left at the default every package declaring a freshly released core would fail spuriously for a day after each release. Both are commented where they live and documented in the versioning guide.

Verification, all local. Fifty-four packages pass against their real declared floors in about fourteen seconds, needing only three registry fetches since the floors collapse to three distinct versions. Understating one floor fails with the offending package, its declared range, and the missing export. Twenty-five unit tests cover the parsing and formatting branches, including ambiguous ranges that are reported rather than guessed at, malformed configs that are skipped rather than thrown, and the absence of the workspace field in the generated config. The full gate passes at sixteen of sixteen.

The wrapper gains the two flags this needs, named after the real CLI flags as the guidelines require, with argv tests and regenerated API docs.

Stacked on the lock-integrity pull request because both touch the build file; it will retarget to the default branch once that merges.

Generated with Claude Code

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,10 +114,29 @@ jobs:
     name: Core version floors
     runs-on: ubuntu-latest
     steps:
+      # Egress is blocked, not audited, because this is the one job that resolves
+      # modules with `--no-lock`. The generated config carries only `jsr:@zuke/*`
+      # specifiers, but that allowlist governs the *import map* — it cannot stop a
+      # source file in a pull request from importing an absolute URL directly,
+      # which `deno check` would then follow, unpinned by the committed lock.
+      # Blocking egress is what actually closes that: an arbitrary host is
+      # dropped at the network layer regardless of what the checked-out code asks
+      # for. The allowlist is the minimum this job needs — the launcher's Deno
+      # bootstrap, JSR for the published core, and GitHub for the checkout. Add a
+      # host here if a run reports it blocked.
       - name: Harden the runner
         uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
         with:
-          egress-policy: audit
+          egress-policy: block
+          allowed-endpoints: >
+            deno.land:443
+            dl.deno.land:443
+            jsr.io:443
+            github.com:443
+            api.github.com:443
+            codeload.github.com:443
+            objects.githubusercontent.com:443
+            release-assets.githubusercontent.com:443
 
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,26 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
           PR_BODY: ${{ github.event_name == 'pull_request' && github.event.pull_request.body || '' }}
 
+  # Verifies each package against the @zuke/core version it *declares*, not the
+  # workspace-local core. Its own job because it reaches JSR for the published
+  # core, and the `./zuke ci` gate must stay runnable offline. Secret-free and
+  # read-only: it only type-checks.
+  core-floors:
+    name: Core version floors
+    runs-on: ubuntu-latest
+    steps:
+      - name: Harden the runner
+        uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Check declared core floors with Zuke
+        run: ./zuke coreFloorCheck
+
   test-os:
     name: Tests (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -274,6 +274,7 @@ the three test layers above and the "read every reviewer comment" rule below.
 | Spell-check                   | `deno task spell`                                    |
 | Pre-commit gate (same as CI)  | `deno task ci` / `./zuke ci`                         |
 | Regenerate `deno.lock`        | `deno task lock`                                     |
+| Verify declared core floors   | `./zuke coreFloorCheck` (needs network)              |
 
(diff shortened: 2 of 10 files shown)


totollygeek:
Addressed the security review finding in 0e7e6e2.

Finding: New networked core-floor job can be abused to force outbound requests from PR code, at .github/workflows/ci.yml:112. Valid. The generated config was built from a package own manifest, which a pull request controls, and the check resolves it with no lockfile — so a pull request could point an import at any URL and have the runner fetch it. Type-checking never executes what it fetches, so this was not code execution, but it was a genuinely new capability: every other job resolves through the frozen lock, and this one deliberately does not.

The generated config now carries only workspace registry specifiers and drops everything else. That mitigation is free here: every import across all fifty-five packages is already a workspace one, because the library is dependency-free. I verified that before relying on it, and all fifty-four packages still pass against their real declared floors afterwards. The entry naming core must now also resolve core itself rather than another package wearing the name, reported as unverifiable rather than silently checked.

The job carries no secrets and a read-only token, and it is the harden-runner audit egress policy rather than block, which is unchanged. If you would prefer the stricter posture, the natural follow-up is an explicit allowed-endpoints list for that job limited to the registry host.

New tests cover the dropped specifiers, a lookalike scope that must not satisfy the prefix match, and the pinned floor winning over the declared range. Full gate green at sixteen of sixteen.

totollygeek:
Addressed the second security review finding in d45a7d8.

Finding: networked CI job can fetch attacker-chosen imports if package manifests bypass the new allowlist, at build/core_floor.ts:128. Valid, and the concrete hole is narrower and more definite than the finding words it. The specifier allowlist governs the import map only. It cannot stop a source file in a pull request from importing an absolute URL directly, which the type-checker follows whatever the map says, and this job is the one place that resolves with no lockfile, so nothing pins what it fetches. The allowlist was necessary and not sufficient.

Egress is now blocked rather than audited for this job, allowing only the launcher Deno bootstrap, the registry for the published core, and GitHub for the checkout. That drops an arbitrary host at the network layer regardless of what the checked-out code asks for, which is the control that actually covers the direct-import case. This is the stricter posture I flagged as the natural follow-up in my previous comment, so it is now done rather than deferred.

I also corrected my own comment on the allowlist, which claimed it removed the arbitrary-fetch capability. It does not on its own. The comment now states that the two are layers of one control and that neither should be removed on the assumption the other suffices.

One thing to watch on the first run: if the block-mode allowlist is missing a host the job legitimately needs, harden-runner will report it as blocked and the job will fail. I kept the list to the minimum I could justify from the sources rather than padding it, on the basis that a reported blocked host is cheap to add and an over-broad allowlist defeats the point. Full gate green at sixteen of sixteen locally.

totollygeek:
Third security review finding, refuted against the code, with one real fix that came out of it. 9a7ffb4.

Finding: generated check config can still inherit attacker-controlled non-@zuke imports from package manifests, at build/core_floor.ts:150.

Not reproducible. The config builder iterates the declared map and copies an entry only when it satisfies the allowlist, then overwrites the core entry with the pinned floor. A non-workspace specifier cannot reach the generated config by that path. Two unit tests already pin this: one asserts an absolute URL and a registry package outside the workspace scope are both dropped while the workspace sibling survives, and one asserts a lookalike scope does not satisfy the prefix match. I re-read the function before answering rather than relying on the tests alone.

The finding does have a real cause, and it is mine. The field holding the raw declared map was still documented as the full import map copied into the throwaway config. That was true when written and stopped being true when I added the filter, and read on its own that line describes precisely the vulnerability reported. So this is a documentation defect that reads as a security defect, which is worth fixing at least as much as the code would have been.

That field is now documented for what it actually is: held unfiltered on purpose, because the core entry has to be inspected as written in order to report a mapping that names core while resolving something else, and explicitly not what reaches the generated config.

No code change, since the code was already correct. Full gate green at sixteen of sixteen.

totollygeek:
Fourth security review finding, refuted. No code change, and deliberately no further push.

Finding: prompt-injection text embedded in code comments could mislead future automation, at .github/workflows/ci.yml:112. The flagged text is the closing sentence of the comment explaining the egress allowlist, telling a maintainer to add a host if a run reports it blocked.

Three reasons this does not hold.

First, it predates this pull request. That sentence is copied verbatim from the equivalent comment on the quality job, which is on master today at line 50 of the same file. I mirrored the existing convention on purpose so the two egress blocks read the same way. This change did not introduce the pattern, so treating it as newly introduced risk is not right.

Second, prompt injection needs an untrusted input channel, and a committed workflow file is not one. It is trusted, review-gated content. Anyone able to edit that comment can edit the allowlist two lines below it, so the comment is not the weak link — it is strictly less powerful than the config it describes.

Third, and decisively for this repo: the only automation here that applies changes and pushes them on its own is the lint self-healing fixer, and it is path-restricted in zuke.ts to packages, tests, and the build file. It cannot write anything under the workflows directory. So there is no automation in this repository that could act on that comment to widen egress, even if it were inclined to.

I am not editing the comment. Removing a maintainer-facing explanation of how to extend an allowlist would make the security control harder to operate correctly, which is a worse trade than the speculative risk. I am also deliberately not pushing again: this is the fourth review round on this pull request and each one was triggered by my own push, with the finding weaker each time. The score has gone from three to two to one out of ten and the generic reviewer reports nothing.

Suppression is a maintainer decision about accepted risk, not mine, so I am leaving the identifier for you rather than adding it myself. It is 37ynljxkxva6d.

State: all ten checks green, mergeable against master, awaiting review.

PR #288, opened 2026-07-30
chore: graduate every package to 1.0.0 and drop @zuke/tsgo

totollygeek:
## What

Two things, deliberately in one change because the second decides what the first publishes.

**Every package graduates to 1.0.0.** The 45 packages still on `0.x` are set to `1.0.0` in their `deno.json` and in `.release-please-manifest.json`, and `bump-minor-pre-major` is turned off. There is now one maturity tier: all 54 packages follow full semver, so a minor or patch release never breaks a public symbol and a breaking change bumps the major. `docs/versioning.md`, the README maturity note, `docs/comparison.md`, `RELEASING.md` and each package's `CHANGELOG.md` are rewritten to describe that single tier, and the root `CHANGELOG.md` records the milestone.

**`@zuke/tsgo` is dropped** rather than graduated, so it never receives a semver promise. TypeScript 7.0 shipped on 2026-07-08 with the native Go compiler as the `tsc` of the ordinary `typescript` package, so `@zuke/tsc` already drives it — and `TsgoSettings` duplicated `TscSettings` method for method. The `tsgo` nightly stays reachable by pointing the `tsc` wrapper's `toolPath` at it. The package is removed from the workspace, the release-please config and manifest, the publish list, the README table, `docs/tools.md`, both cheatsheets, the release-config test and `deno.lock`. Version `0.1.3` remains resolvable on JSR for anyone already pinned to it; the package will be archived there.

`AGENTS.md`'s stale claim that Deno cannot delegate type-checking is also corrected: `deno check --unstable-tsgo` exists now, but Deno documents it as an unstable preview that is not for CI, so the default checker stays authoritative for this repo.

## Why the versions are set by hand

release-please needs a triggering commit per package path to cut a release, and this change carries no feature or fix for any of them. `.release-please-manifest.json` is release-please's source of truth for a package's current version — when no matching GitHub release exists it backfills the tag from the manifest — so setting the manifest is a supported bootstrap. The PR title is a `chore`, so release-please cuts nothing; `publishJsr` reads each `deno.json` version directly and skips versions already on JSR, so merging publishes the 45 packages at `1.0.0`. The next ordinary `feat` or `fix` bumps from `1.0.0` with no further manual edits.

## Adversarial review

Two defects were found and fixed before opening this PR, each with a regression test or a verified fix.

1. `packages/cli/src/version.ts` still held `0.8.1` while its `deno.json` said `1.0.0`. That file is a release-please extra-file, so only release-please normally rewrites it; the CLI prints the constant, so `@zuke/cli@1.0.0` would have permanently reported `0.8.1` on an immutable JSR version. Fixed, and `tests/release_config_test.ts` now asserts that every extra-file version marker matches the manifest. Verified the new test fails on the defect and passes with the fix.
2. `docs/getting-started.md` still described a 55-package workspace.

Also checked and clean: no other version is duplicated anywhere; no live reference to `@zuke/tsgo` outside historical changelogs; both cheatsheet copies stay in sync; the scaffolder emits no `0.x` specifier; and the leanest packages have real task surfaces rather than stubs. While verifying the last point, `packages/cli/src/import.ts` was aligned with `zuke setup` so an imported build pins the caret major instead of resolving unpinned.

## Verification

`deno task ci` is green, 16 of 16 targets. `tests/release_config_test.ts` passes with 9 tests, covering the six membership lists after the removal.

## Follow-ups for the maintainer

Archive `@zuke/tsgo` on JSR after merge, and create the `component-v1.0.0` tags at the merge commit so the next release's changelog compare links resolve.


totollygeek:
The one AI-review finding — `pm3oldslqkyj`, "generated builds now pin imports to `@^1`, which can pull in future major releases without review" at `packages/cli/src/import.ts:392` — is a false positive, and the direction of the claim is inverted.

`^1` is the range `>=1.0.0 <2.0.0`. A future `2.0.0` cannot satisfy it, so the scaffolded import cannot resolve a future major. What that line replaced was an **unpinned** `jsr:@zuke/core` specifier, which resolved the latest version including any future major — so this change narrows supply-chain exposure rather than widening it. It also brings `zuke import` in line with what `zuke setup` has always scaffolded, and with the pinning guidance in `docs/versioning.md`; a consumer's committed `deno.lock` still fixes the exact resolved version regardless of the range.

Suppressed with that reasoning recorded next to the existing suppressions in `zuke.ts`. Everything else on this PR is green, including `Core version floors` and all three OS matrices.

totollygeek:
Second AI-review pass raised two findings; both are answered rather than ignored, and the reasoning is recorded next to the suppressions in `zuke.ts`.

**`io22vnfjvb1t` — "generated build scaffolds now hardcode `@^1` imports, which can break consumers and bypass their existing lockfile pinning"** — refuted. This is the previous finding reworded. `^1` is `>=1.0.0 <2.0.0`, so no future major satisfies it, and a range in a generated source file is resolved *through* the consumer's `deno.lock`, not around it — a lockfile entry that satisfies the range is what gets used, and under `--frozen` a new specifier fails loudly instead of being resolved silently. `zuke import` also only writes a new `zuke.ts`; it never rewrites an existing project's dependencies. The line replaced an unpinned specifier that *did* resolve any future major.

**`3bja5rj1xp93t` — "removing `@zuke/tsgo` drops a published package without an in-repo compatibility shim"** — deliberate, not an oversight. JSR versions are immutable, so `@zuke/tsgo@0.1.3` keeps resolving for anyone pinned to it; a shim package would have to carry its own `1.x` semver promise for a wrapper whose settings class duplicated `TscSettings` method for method. The migration — point the `tsc` wrapper's `toolPath` at the `tsgo` nightly — is documented in the root `CHANGELOG.md`. The one residual cost, stated plainly: the package's JSR page keeps its last-published README rather than a "use `@zuke/tsc`" pointer, since no further version will be published. The package will be archived on JSR after merge.

All 10 checks pass on the latest push.

totollygeek:
Third pass, and `398y0770bqo1n` is a genuinely new argument rather than a reword of the earlier caret-pin findings, so it gets a specific answer: *"can break builds if those packages are not actually published at 1.x"*.

The concern is real in principle — `@zuke/core` is already `1.32.1`, but `@zuke/cmd` is `0.3.2` on JSR until this PR merges, so a build scaffolded with `jsr:@zuke/cmd@^1` would not resolve today. It cannot happen in practice, and the mechanism is verifiable rather than theoretical:

- `publishJsr` iterates `PACKAGES` from `build/packages.ts` sequentially, awaiting each publish. `cmd` is at index 8 and `cli` at index 10, so `@zuke/cmd@1.0.0` reaches JSR before `@zuke/cli@1.0.0` — the new scaffolder cannot ship ahead of the package it pins.
- `publishOne` throws when a publish does not reach JSR, which aborts the loop. A failure on `cmd` therefore means `cli` is never published in that run either, so the dangerous ordering has no path.
- Consumers on the current `@zuke/cli` are unaffected: the scaffolder they have emits the unpinned specifier.

`qm02o3kwkjt5` is the compatibility-shim finding again, already answered above and unchanged: JSR immutability keeps `@zuke/tsgo@0.1.3` resolvable, a shim would need its own `1.x` promise for a duplicate wrapper, the migration is in the root `CHANGELOG.md`, and the package will be archived on JSR after merge.

Not adding further suppression entries for these: the reviewer rephrases each run and mints a fresh id, so id-based suppression does not converge, and the check already passes its own gate at 5/10. The reasoning for the earlier ids is recorded in `zuke.ts`, and these two are answered here in the thread.

PR #295, opened 2026-08-06
feat(core): prepare every workflow step for generation

totollygeek:
## Why

Zuke exists so nobody writes CI YAML by hand, yet Zuke's own workflows still carried steps only CI could run: three scanner installs, a token mint, two artifact uploads, a SARIF upload, and a base-branch fetch. None of that belonged in YAML, and while it lived there, `./zuke security` could not run on a laptop at all.

This moves every remaining step into the build and adds the CI-model surface a generated workflow needs. It is deliberately preparation only — the workflows are still hand-written here, thinned. A follow-up replaces them with generated files, which is now a mechanical change.

## What moved into the build

| Step that is gone | Where it lives now |
| --- | --- |
| `pipx install zizmor`, two `go install`s, `$GITHUB_PATH` patching | Pinned toolchain entries in `build/scanners.ts`, fetched and checksum-verified on demand |
| `actions/create-github-app-token` | the `appToken` task on `GhTasks` — signs the app JWT, resolves the installation, mints a scoped token |
| `codeql-action/upload-sarif` | the `uploadSarif` task on `GhTasks` — gzip + base64 to the code-scanning endpoint |
| `upload-artifact` for the gitleaks report | The redacted findings render into the job summary |
| `upload-artifact` for the scorecard results | Redundant once the SARIF reaches code scanning |
| `git fetch` before the AI review | A `reviewBase` target, so the same command works locally |

The scorecard action itself stays, because publishing the public OpenSSF score is something only it can do.

## Consequences worth reviewing

**The security scan is now part of the gate.** It was excluded precisely because it assumed three binaries on PATH; with the scanners provisioned by the build that reason is gone. It adds about two seconds to a warm run, and one download on a cold one.

**Two things had to be built rather than wrapped.** GitHub has no CLI for minting an app token or uploading SARIF, so those tasks use HTTP internally while keeping a settings-lambda surface. Both are tested through a `fetch` seam, and the app-token tests generate a real RSA key rather than committing a fixture, covering both PEM encodings GitHub issues.

**`archive` and `binaryPath` now accept per-platform resolvers.** Without this no Go or Rust tool is declarable cross-platform, since they ship a tarball with a bare binary on Unix and a zip with an `.exe` on Windows. `url` and `checksum` were already resolvers; these two were the inconsistency.

**Hardening and checkout are declarable, and their pinned action reference is required, not defaulted.** A default would be either a floating tag, which supply-chain scanners reject, or a commit SHA baked into a published package that goes stale between releases. Requiring it keeps the pin with the caller.

## Two things the scan itself caught

Both surfaced only once this ran in CI, and both were worth having:

The new test file contained a literal `BEGIN RSA PRIVATE KEY` header in its malformed-input cases, and gitleaks flagged it — correctly, since a scanner that ignores that header where a test happens to sit is a scanner you cannot trust. The PEM label is now assembled rather than written out, so no complete header appears in the source and nothing had to be allow-listed. The follow-up finding was a base64-shaped literal next to a `privateKey` call, which turned out to point at a real gap: a PEM whose body is not valid base64 escaped as a bare `DOMException` instead of a message naming the problem. That path now has its own error and its own test.

It also exercised the job-summary replacement in the exact case it exists for. The failing scan rendered both hits with file, line, rule and fingerprint — the information the artifact upload used to carry.

## Verification

`deno task ci` is green: 18 of 18 targets, including the coverage gate and the newly-included security scan. `./zuke coreFloorCheck` confirms all 53 packages still type-check against their declared `@zuke/core` floor — which is why `@zuke/ai` keeps its own job-summary writer instead of delegating to the new core helper, since its declared floor predates it.

## A correction worth recording

An earlier draft of this change added a gate asserting that a run had not rewritten a declared workflow file, on the belief that nothing verified them in CI. That was wrong, and its own integration test is what proved it: the CLI already syncs declared CI files with `check: isCI()`, so a runner verifies them and fails the build while a local run writes them. The drift was gated all along. That gate and its module are removed rather than left in as harmless duplication.

## Still open

Generating the workflows means Dependabot's weekly action-SHA bumps would land in generated files. Deciding where the pins should live is left to the follow-up; the required `action` field is the hook for whichever mechanism wins.


totollygeek:
Verified each finding against the code. One was a real defect and is fixed; the others are addressed below rather than dismissed.

**Unescaped backticks in the gitleaks summary — fixed.** Real defect. A path may contain a backtick, which ended its code span early and garbled the row. That row is now the only place a failing scan reports *where* a secret is, so a value that hides its own location defeats the point. The delimiter is now sized to one longer than the longest backtick run inside the value, per CommonMark, with a space pad when the value starts or ends with one. Three regression tests cover the single backtick, the leading/trailing case, and a run of two.

**Runtime-installed scanners — this narrows the trust boundary rather than widening it.** What it replaced was `pipx install zizmor==1.25.2` plus two `go install ...@version` invocations, which resolved and built from PyPI and the Go module proxy with no integrity check at all. Each asset is now pinned to an exact version *and* verified against a committed SHA-256; a mismatch throws and installs nothing, and a cached copy is re-hashed before reuse, so a tampered binary in a shared tools directory is re-downloaded rather than trusted. The network dependency itself is not new to the gate either: the spell target already fetches `npm:cspell@9` at runtime, so a fresh clone has always needed network for its first run.

**App private key in the sync step — confirmed, and mitigated in a follow-up commit.** The framing naming a fork or pull-request path does not hold: that workflow triggers only on a push to the default branch and on workflow_dispatch, and fork pull requests receive no secrets regardless. The observation underneath is right, though, and it is the one thing in this change that genuinely widens a blast radius — the build process now holds the key rather than a token minted from it, and a minted token lasts an hour and reaches one repository with two permissions. That job now blocks egress instead of auditing it, with an allowlist verified against the hosts the last successful release actually contacted. It is mitigation, not elimination: keeping the private key out of the build process entirely would mean keeping the marketplace action for that one step, which is a defensible exception if the isolation is worth a step.

**Base fetch removal — by design, and it degrades to no review rather than a wrong one.** The option defaults to keeping the fetch, so nothing changes for anyone who does not opt out. Where it is opted out, the build's own target performs the fetch, which is what makes the same command work locally where no workflow step exists. If that fetch fails it warns and the base stays unresolved, so the reviewers' own diff step errors and, under their warn-on-error setting, the review is skipped — a missing review, not a review against a misleading base.

totollygeek:
Verified this round's three findings. One was real and is fixed; the other two are answered.

**Minting for any `WEBSITE_REPO` slug — confirmed, fixed.** This was a real escalation, and one I introduced. That variable used to decide only *where* the sync pushed, and a redirect then failed, because the token's scope was a literal in the workflow's action inputs and did not cover the new target. Deriving the minted scope from the variable turned a redirect into a redirect with a working credential for wherever it pointed — anywhere the app is installed. Minting is now restricted to the one repository this sync exists to update; redirecting still works and is still tested, but requires supplying `WEBSITE_SYNC_TOKEN`, so whoever redirects the sync provides the credential rather than borrowing the app's. Practical reach was narrow — setting that variable already needs a privileged change — but the structural point was right.

**App private key in the step environment — confirmed earlier, mitigated, and the residual is deliberate.** Fourth report of the same issue; noting for the record that its severity has moved between medium and high across runs on an unchanged diff. The mitigation stands: that job now blocks egress rather than auditing it, with an allowlist checked against the hosts the last successful release actually contacted. What remains is a real trade rather than an oversight — removing the exposure entirely means keeping a marketplace action for that one step so the build receives a scoped one-hour token instead of the key. That is a reasonable exception to the no-steps goal and is flagged for the maintainer's decision, not silently accepted.

**Losing the artifact as a verification path — refuted.** The artifact was never independent of the build: the same step produced both it and the summary, so a build that could falsify one could equally skip the scan or skip the upload. The integrity of the gate rests on the build either way, which is unchanged. On the practical side the summary carries every field the JSON had that matters — file, line, rule, fingerprint, with values redacted in both — and it outlives the artifact's 14-day retention, since it persists with the run. The raw JSON is still written to disk for a local run to inspect.

totollygeek:
**HTML escaping in the job summary — refuted, verified against the rendering path.** Every value a repository can influence (rule, file, commit, fingerprint) is wrapped in a Markdown code span by the same helper. CommonMark requires code-span content to be literal, so the renderer escapes the angle brackets and an `` in a filename appears as visible text rather than an element; GitHub additionally sanitizes summary HTML with the allowlist it uses for comments, which drops event handlers and script tags regardless. Checked by rendering both payloads through the real formatter. The only value not in a code span is the line number, which the parser types as a number and defaults to zero for anything else, so it cannot carry markup. The escaping that *was* missing here was structural rather than HTML — the pipe, the code-span delimiter, and control characters — and all three are fixed with regression tests.

**App private key — fifth report of an accepted, documented trade.** No new information in this round; the mitigation and the residual are described in the two comments above. Worth noting the severity has moved between medium and high across runs on an unchanged diff. This is deliberately *not* being added to the reviewer's suppression list: it is a real finding whose residual risk is a maintainer's call, not a false positive to silence. If the decision is to keep the marketplace action for that one step, the finding goes away on its own; if the decision is to accept the exposure with egress blocked, a suppression entry with that reasoning is the right way to record it.

PR #298, opened 2026-08-06
feat(core): generate every workflow, sourcing pins from the committed files

totollygeek:
The repository now contains **no hand-written CI YAML**. All six workflows are declared in `zuke.ts` and written by `zuke generate-ci`.

## How Dependabot keeps working

This was the open question, and the answer decides the shape of everything else. Dependabot only reads `.github/workflows/*.yml` and a root `action.yml` — a TypeScript constant is invisible to it. Put the pins in code and its weekly bumps stop, silently, and stale pins mean missing the security fixes the actions themselves ship.

So the direction is inverted: **Dependabot edits the committed workflow, and the generator reads each SHA back out of it.** Regeneration reproduces whatever Dependabot last wrote, so its PR lands exactly as it does today with no human step and no drift.

    Dependabot PR edits ci.yml → uses: actions/checkout@NEWSHA # v7.0.2
    ./zuke ci regenerates, reads NEWSHA back from the committed file
    generated == committed → nothing to commit → PR merges

Yes, the generator reads its own output. That is deliberate and confined to the single field Dependabot owns. Everything else about a workflow still comes from the declaration, so structure cannot drift while the pin stays current. `SEED_PINS` covers the bootstrap case where no committed file mentions an action yet.

## This fixed a live bug, not just a hypothetical

Two workflows were *already* generated from hardcoded SHAs — three in `integrationCi`, two more inside `@zuke/ai` — so a Dependabot bump to either file was reverted by the very next run. That already happened: commit `a542d92` bumped `ai-review.yml`, and the constant in `@zuke/ai` had to be hand-edited afterwards to match, or the next run would have undone it.

Both now source their pins from the committed files, and a test fails if any declaration hardcodes a SHA again.

## Two supporting changes

**The YAML emitter can now attach a trailing comment to a scalar.** This is not the rationale-comment case that was previously ruled out — a pin's `# v1.2.3` is *data*. Dependabot rewrites the SHA and that comment together, and its own metadata tracks `dependency-version` from it. Every generated pin carries its version, asserted by a test.

**The harden-runner allowlist renders space-separated on one line**, which is what the folded scalar (`allowed-endpoints: >`) in the previous hand-written files collapsed to. The action passes that input to its agent as an opaque string and documents no delimiter, so the control gating a secret-bearing job keeps the form already known to enforce correctly rather than a newline-separated list that merely looks tidier.

## Verification

Textual diffs against the replaced files are dominated by comment removal and formatting, which says nothing about behaviour. So both versions of each file were parsed and compared structurally, normalising only what carries no behaviour — step display names, `null` against `{}` for a bare trigger, a scalar against a one-element `needs`, allowlist whitespace, and string against number for action inputs:

    ✅ ci.yml — behaviourally identical
    ✅ release.yml — behaviourally identical
    ✅ security.yml — behaviourally identical
    ✅ scorecard.yml — behaviourally identical

zizmor and actionlint pass on all six. A simulated bump applied across every file survives regeneration with the new SHA intact. A *half-applied* bump fails loudly, naming both files and both SHAs, rather than picking one and thereby either reverting the bump or spreading it.

## Two things worth a reviewer's attention

The half-applied-bump guard fails the **whole run**, not just generation, because pins are requested while the build's fields initialise. That is the intended trade — two workflows pinning one action to different commits is a supply-chain inconsistency, and a gate that carried on would be reporting on a repository state nobody chose — but it is a deliberate choice rather than a detail.

The scorecard action stays. Publishing the public OpenSSF score is something only it can do, so that job is `harden → checkout → scorecard action → ./zuke scorecardSarif`. Every other job across all six workflows is now `harden → checkout → ./zuke <target>`.


totollygeek:
Verified both findings.

**Full history on pull requests — refuted, this is unchanged behaviour.** `fetch-depth: 0` was already in the hand-written `security.yml` before this change; the generated file reproduces it, which is what the structural comparison in the description asserts. Nothing about the checkout depth is new here.

The concern behind it is also already handled, and was the reason the setting exists in that form: gitleaks needs history to scan past commits, but a full fetch brings every branch, so the `security` target scopes the scan to the pull request's own commits via `GITHUB_BASE_REF` — otherwise a secret-shaped string on an unrelated branch would fail the scan on every open pull request, blaming whichever one was looked at. A push to the default branch and the weekly run still walk the whole history, so nothing stops being covered.

**Self-referential trust loop — accurate, and the deliberate trade.** Worth stating plainly rather than arguing away: the generator does read its own output, and that is the cost of the option chosen here.

What bounds it is scope. Exactly one field is sourced this way — the pinned SHA — and it is the one field an external bot legitimately owns. Everything else (triggers, permissions, egress policy, job graph, steps) comes from the declaration in `zuke.ts` and is overwritten on every run, so structure cannot drift while the pin stays current. Editing a pin in a workflow is a reviewed change to a committed file, the same trust boundary that applied when the pin was hand-written; what changed is which file holds it, not who can change it. A half-applied bump fails loudly rather than resolving to either SHA.

The alternative was pins in a TypeScript constant, which is strictly worse for the property that matters: Dependabot cannot see a `.ts` file, so its weekly bumps would silently stop, and stale pins mean missing the security fixes the actions themselves ship. This change also removes two *existing* instances of that failure — `integrationCi` and `@zuke/ai` both generated from hardcoded SHAs, so a bump to either file was reverted by the next run, which had already happened once.

If the loop is judged unacceptable regardless, the alternative that keeps Dependabot working is a root `action.yml` holding the pins (one of the two paths it scans). That trades the loop for making this repository marketplace-visible as a composite action — a product decision rather than a technical one, and the reason it was raised for a maintainer's call rather than chosen unilaterally.

totollygeek:
Both findings assert something about the generated files, so both are checkable. Neither holds.

**`branch_protection_rule` was not removed.** It is in the generated file:

    "on":
      push:
        branches:
          - master
      branch_protection_rule: {}
      schedule:
        - cron: "18 3 * * 2"

The declaration sets `branchProtectionRule: true`, and the trigger renders as `{}` — an event with no filter, which is how GitHub spells "every occurrence". The previous hand-written file wrote it as a bare `branch_protection_rule:` with nothing after it, which YAML parses as `null` and GitHub treats identically. The structural comparison in the description normalises `null` against `{}` for exactly this reason, and reported the file unchanged.

**The CI job does not run untrusted code with a write-scoped token.** Two separate reasons, and the second is the load-bearing one.

It is not new: `persist-credentials: true`, `contents: write`, and the head-ref checkout were all in the hand-written `ci.yml` before this change. The comparison found that file behaviourally identical, so "now runs" is not what changed.

More importantly it is not unsafe, because the trigger is `pull_request` — not `pull_request_target`. On a fork pull request GitHub caps the token to read-only whatever the workflow requests, and withholds every secret; the `permissions:` block is a ceiling, not a grant. The ref expression makes the same distinction explicitly: it checks out `github.head_ref` only when `head.repo.full_name == github.repository`, so a fork PR gets the default merge-ref checkout and the fixer has nothing to push with. On a same-repo branch the token is write-scoped, but the author already has push access, so the workflow grants nothing they did not have.

`pull_request_target` is the trigger where this finding would be correct — it runs the base repo's workflow with full secrets against fork code. That is precisely why this workflow does not use it.

totollygeek:
All three findings assert that this change altered something. Each is checkable against the diff, and none of the three holds.

**The publish job is not newly exposed, and it is not ungated.** Its `needs:` edge is in the generated file:

    publishJsr:
      needs:
        - release

and `syncWebsite` in turn needs `publishJsr`, so the chain is intact. The trigger is unchanged too — `push: [master]` plus `workflow_dispatch` is exactly what the file carried before this pull request. Nor is a push to the default branch the gate: `publishJsr` skips every version already on JSR, so it is a no-op unless release-please actually cut one in the job before it.

**The composite action cannot reach a secret.** It declares six inputs, none of them a secret, and the file contains zero references to `secrets.` — a composite action has no access to them unless a caller passes one in explicitly, and nothing here does. Its one `run:` step is `./zuke ${{ inputs.target }}`, guarded by `if: inputs.target != ''`, so a caller that omits the target gets hardening and checkout and nothing else. That step also runs the repository's *own* build file, on a runner the caller controls, which is the entire purpose of the action rather than an escalation. "Secrets-friendly defaults" describes something that is not in the file.

**The security workflow's pull-request trigger is unchanged.** Before this change it read `pull_request:` with nothing after it; it now reads `pull_request: {}`. Those are the same thing — YAML parses a valueless key as `null`, and GitHub treats a null and an empty mapping identically, both meaning "every pull request, no branch filter". The generator emits the explicit form because it does not write valueless keys. This is one of the normalisations the structural comparison in the description applies, which is why it reported the file unchanged; the earlier round raised the same equivalence about `branch_protection_rule`.

Worth noting across the review rounds on this pull request: the recurring pattern in these findings is a formatting difference in generated YAML read as a behavioural change. The description's structural comparison exists precisely because a textual diff of a hand-written file against a generated one is dominated by exactly that.

totollygeek:
One of these is backwards on the fact; the other is fair and is now covered by a test.

**Credentialed checkouts are not the default — the action is stricter than the one it wraps.** The input defaults to `"false"`:

    persist-credentials:
      description: "Leave the token in git config, for a later push."
      required: false
      default: "false"

`actions/checkout` itself defaults that to `true`, so a caller using this action gets a *non*-credentialed checkout where using checkout directly would leave the token in git config. One job opts in — the gate, so its lint fixer can push to the pull-request branch — and that is stated where it happens. The finding describes the underlying action's default, not this one's.

**A single point of failure, correctly noted — and now tested.** The observation is right in one specific way that is worth closing rather than debating: the pin pattern requires a full 40-character SHA, so a *mangled* edit matches nothing, the action falls through to a workflow or the seed, and the manifest quietly stops being the source with nothing reporting it. A wrong-but-well-formed SHA would still be used, as it would from any pin location.

There is now a test that resolves pins from the manifest alone, with no workflow directory to fall back to, and asserts each expected action has a full SHA and the version comment Dependabot rewrites beside it. It was checked against a manifest with a deliberately mangled SHA to confirm it fails there, rather than only passing on a good file.

On the broader point: one manifest is a smaller failure surface than the alternative, not a larger one. Before it, six generated files each had to agree, and a bump reaching some but not others was a real state the code had to detect and report. The manifest replaces that with one place to be right — and it is the only arrangement where nothing the generator writes feeds back into what it writes.

totollygeek:
This one was right, and the concrete defect behind it was worse than the summary said. Fixed in `6677ed9`.

**There was an actual code injection in the action.** Its run step was:

    run: ./zuke ${{ inputs.target }}

which splices the caller's string into the command text. zizmor rates that a high-confidence `template-injection`, and the reasoning matches the finding exactly: because the action is consumable from another repository, a caller who derives that input from something untrusted — a pull-request title, an issue body — would have been handing arbitrary code to this action, and the blast radius would have been theirs. The value now rides in through the environment and is quoted in the shell, which is the difference between passing a value and passing code:

    env:
      ZUKE_TARGET: ${{ inputs.target }}
    run: ./zuke "$ZUKE_TARGET"

**The more useful half is why it reached CI at all.** The `security` target scoped zizmor to `.github/workflows`, so the root action was never audited — despite being subject to the same rules and being the more exposed of the two files. `action.yml` is in scope now, and the gate was checked by reintroducing the injection to confirm it fails, rather than only observing that it passes in the fixed state.

The other inputs are safe by construction and stay as they are: `egress-policy`, `allowed-endpoints`, `persist-credentials`, `fetch-depth` and `deno-version` are passed as `with:` inputs to other actions, never into a `run:` block, so there is no shell to inject into.

On the framing — "a new supply-chain trust boundary" is a fair description and worth keeping in mind rather than dismissing. The manifest has to be a root `action.yml` for Dependabot to see it, so the boundary is the cost of the pins arrangement rather than something incidental; what is in our hands is that the file is audited by the same scanner as the workflows, which it now is.

totollygeek:
This is the same finding as two earlier rounds, but it names a specific mechanism — an "explicit fork guard" — so it is worth answering on that mechanism rather than by repetition.

**No explicit guard is needed, because the trigger is `pull_request`, not `pull_request_target`.** On a fork pull request GitHub caps the token to read-only regardless of what the workflow requests, and withholds every secret. The `permissions:` block is a ceiling, not a grant: asking for `contents: write` on a fork PR yields a read-only token. An `if:` condition would not add protection here — it would only skip a job that already cannot write.

`pull_request_target` is the trigger where an explicit fork guard is load-bearing, because it runs the base repository's workflow with full secrets against fork code. The AI-review workflow in this repository *does* use such a guard (`github.event.pull_request.head.repo.fork == false`) for a different reason: not to prevent writes, but because a fork PR receives no API keys, so the reviewers would skip anyway and the job is better not started.

The gate job also distinguishes fork from same-repo explicitly, in the one place where it matters — the checkout ref:

    ref: ${{ (github.event.pull_request.head.repo.full_name == github.repository) && github.head_ref || '' }}

A same-repo branch checks out its head so the lint fixer can push to it; a fork PR falls through to the default merge-ref checkout, where the fixer has no credential to push with. That is the guard, and it is expressed where the behaviour differs rather than as a blanket condition.

None of this is new in the pull request either: the trigger, the permissions, and that ref expression are unchanged from the hand-written `ci.yml`, which the structural comparison in the description reported behaviourally identical.

totollygeek:
Retro-attributing the answers on this pull request to their finding ids, and citing the id from here on.

**`2x2nx340d2ee4`** — "CI job now runs with write-scoped token on PR-triggered workflow without an explicit fork guard". Answered in the comment above: no explicit guard is needed because the trigger is `pull_request`, not `pull_request_target`, so GitHub caps a fork's token to read-only and withholds every secret whatever `permissions:` requests. The fork distinction that does matter is in the checkout ref, where the behaviour genuinely differs. Unchanged from the hand-written file.

**`m5aqoc5dxg5g`** — "Root composite action can be consumed remotely without checkout, so its inputs become a new supply-chain trust boundary". Confirmed and fixed in `26560f3`, which now names the id. It was a real code injection: `run: ./zuke ${{ inputs.target }}`. The commit also puts `action.yml` in zizmor's scope, which is why the gate had missed it.

**Whether `2x2nx340d2ee4` is new, honestly: I cannot tell from the record, and that is the gap worth fixing.** Three rounds on this pull request have raised the same write-token concern under three different wordings — at `ci.yml:19`, then `ci.yml:16` — but the reviewer upserts a single comment, so each run overwrote the previous body and the earlier ids are gone. `2x2nx340d2ee4` is not an id I had recorded, which is consistent with either a genuinely new finding or the same one rehashed after its wording changed. I am not going to claim it is a duplicate on that basis.

What I have changed so it is answerable next time: `6326d67` makes quoting the id a rule in `AGENTS.md`, for the reply and for any commit the finding prompts. These are the same identifiers the `suppressions(...)` list in `zuke.ts` already uses to dismiss a false positive, so the practice is the read-only half of one the repo relies on.

Left deliberately unsuppressed rather than dismissed: `2x2nx340d2ee4`, and the app-private-key finding on the release job. Both describe real properties — the second especially — whose residual risk is a maintainer's decision rather than a false positive.

totollygeek:
Two new ids this round — and they *are* new, which the record can now show rather than guess at.

**`1s6pnr891abw8`** — "Security workflow now runs on all pull requests instead of only push/scheduled events". Refuted: that trigger predates this pull request. On the merge base it reads

    on:
      push:
        branches: [master]
      pull_request:
      schedule:
        - cron: "42 5 * * 3"

and on the branch it reads `pull_request: {}` in the same position. A valueless YAML key parses as `null`, and GitHub treats null and an empty mapping identically — both meaning every pull request, unfiltered. The generator emits the explicit form because it does not write valueless keys. Scanning pull requests is also the intended behaviour: it is what catches a secret before it merges, and the target scopes gitleaks to the pull request's own commits so an unrelated branch cannot fail it.

**`2gcii7ybvfdhp`** — "Documentation adds an AI-review reply instruction that can be abused as prompt injection". Worth taking seriously, and the same class as a finding this repository already dismissed with reasoning: `3ud7i3zbigfl0`, recorded in `zuke.ts` as "a static, author-written workflow comment … it does not execute and feeds no runtime model, so there is no prompt-injection surface".

The reasoning holds here too, with one part worth stating rather than waving through. The chain the finding implies is: a pull request crafts content, the reviewer echoes it into its comment, an agent reads that comment. The middle and last links are pre-existing — the rule directly above the new one already instructs agents to read every review comment, and has since before this change. What the new line adds is "quote the finding's id", where an id is six to fourteen lowercase alphanumerics. It tells an agent to copy a short opaque token out of a comment it was already reading; it does not tell it to trust, execute, or act on comment prose.

`AGENTS.md` is also inside the same trust boundary as the code: it changes only through a reviewed pull request, under a ruleset requiring an approval and a code-owner review. An attacker who can edit it can already edit the build.

I have not added `2gcii7ybvfdhp` to the suppress list, though the precedent would support it. Suppressing a prompt-injection finding is a maintainer's call, and it is better made deliberately than by me while answering it.

PR #301, opened 2026-08-08
feat: version the Marketplace action from the build, and let it check out a ref

totollygeek:
First of three. This one adds what must exist **before** the workflows can adopt the action.

- **#301 (this one)** — `action.yml` gains a `ref` input and a refusal; `actionRelease` versions the action from the build.
- **next** — `@zuke/core`'s generator emits the Zuke action in place of the harden/checkout pair, and `@zuke/ai` stops hand-rolling those steps.
- **then** — this repository's own workflows adopt it.

## `ref` input, and the refusal that comes with it

A job that pushes a fix back to a pull request needs the head branch, not the detached merge ref. The action had no way to ask for that.

It also hands every consumer the classic escalation, so `ref` arrives with a refusal step: on an event whose trigger content a contributor writes, checking out a ref they control is refused outright, with an `::error::` annotation naming the event and the fix.

Three things about that guard came out of review, because the first version had all of them wrong:

- **It ignores `target`.** Requiring one guarded a single *spelling* of the mistake. A caller who omitted the input and ran the build themselves in the next step of their own job got the identical outcome — same secrets, same arbitrary code. Checking the code out is the bug; what runs it afterwards need not be this action.
- **It is an allowlist, not a denylist.** The first version named `pull_request_target`; the second named eight events. Both are denylists, and a denylist is correct only until GitHub adds an event — at which point the guard silently stops firing rather than failing. It now names the events whose ref can only come from someone who could push the same code directly, and refuses everything else. The safe set is small and stable; the dangerous set is open-ended.
- **The message goes through `env`.** An earlier draft interpolated the event name into the script text, which is the exact thing the file's own comment forbids two steps further down.

Allowed: `pull_request` (a fork gets no secrets and a read-only token), `push` and `merge_group` (write access required), `workflow_dispatch`, `schedule`, `release` (triggered by someone who could push directly). Refused: everything else, including `issue_comment` and the slash-command pattern, `workflow_run`, `pull_request_target`, and the discussion and review events.

## Versioning as a target

`./zuke actionRelease` cuts the next patch, but only when `action.yml` has changed since the last release. That guard is also the loop breaker for the third PR, where the target commits its own output back to master.

It refuses to run from anywhere but a clean `master` that matches its remote. The tag it force-moves is the one every consumer of `@v1` follows and there is no undo short of another force-push — a feature branch would publish unmerged work, and a dirty tree would tag a commit that does not contain the change that triggered it.

Publishing to the Marketplace stays manual. GitHub gates it behind a 2FA confirmation no token can perform, and app tokens are the case that still requires it, so the target prints the draft-release link rather than pretending to have finished.

## Where the self-pin lives

`build/action_version.json`, a new committed file. Not `action.yml` — it cannot pin the action to itself. Not a generated workflow — that is the feedback loop `build/action_pins.ts` exists to avoid.

`assertPinnedActionMatches` compares that pinned release against the working tree, because **nothing else notices when they diverge**: every test here asserts against `action.yml`, so they stay green while the artifact the workflows actually run is missing an input or a guard. It is wired into the gate in the third PR, where it starts mattering.

## Review findings addressed

Two independent adversarial reviewers before the PR, plus the `@zuke/ai` reviewers on it.

| Finding | Fix |
| --- | --- |
| Refusal bypassable by omitting `target` and running the build in the caller's own step | Guard no longer keys on `target`; a test asserts the clause cannot come back |
| Refusal covered one of several secret-bearing events | Rewritten as an allowlist |
| `1mdlehy1n2nrl` — denylist potentially incomplete | Inverted, so an unrecognised event is refused rather than permitted |
| `31avekkyncjz5` — raw capture access on the tag list | Kept deliberately, and a truncated capture now fails loudly; a partial tag list would otherwise re-cut an existing version with no symptom |
| `changedSince` failed open — a git exit of 128 read as "changed", cutting a release on every faulted run | Throws on anything but 0 or 1 |
| `pinFor` validated the SHA rigorously and the version not at all, so a trailing newline would emit a stray line into every generated workflow | Version anchored and whitespace-rejected |
| No branch, worktree or remote guard before a force-push | `assertReleasable` |

Reviewers also found that the workflow adoption would pass `ref:` to an action version without that input. GitHub warns and continues, so the gate job would have lost its head-branch checkout and the AI lint fixer would have had nothing to push to. That is why the adoption is a separate PR, gated on `actionPinCheck`.

## Tests

14 in `tests/action_release_test.ts`, 7 in `tests/action_manifest_test.ts`. A reviewer independently mutation-tested the release suite and confirmed every assertion fails when its subject breaks.

`./zuke ci` green, 17/17.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -81,17 +81,25 @@ runs:
     # mistake rather than the mistake.
     #
     # Hence: the refusal ignores `target`, and names the events rather than the
-    # steps. `pull_request` and `push` are absent from the list deliberately — a
-    # fork pull request receives no secrets and a read-only token, and pushing
-    # requires write access already. `workflow_dispatch`, `schedule` and
-    # `release` are absent too: their ref is chosen by someone who can already
-    # push.
+    # steps. It is an allowlist, not a denylist, because a denylist of the
+    # events that are dangerous today is wrong the moment GitHub adds one —
+    # and the failure is silent, since the guard simply does not fire. Listing
+    # the events where the ref can only come from someone who can already push
+    # means anything unrecognised is refused instead.
+    #
+    # Why each of these is safe: `pull_request` gives a fork no secrets and a
+    # read-only token; `push` and `merge_group` require write access;
+    # `workflow_dispatch`, `schedule` and `release` are triggered by someone
+    # who could push the same code directly. Everything else — `issue_comment`
+    # and the `/build` slash-command pattern, `workflow_run`,
+    # `pull_request_target`, the discussion and review events — lets a
+    # contributor's text choose what gets checked out while the job holds this
+    # repository's secrets.
     - name: Refuse an untrusted checkout on a secret-bearing event
       if: >-
-        inputs.ref != '' && contains(fromJSON('[
-          "pull_request_target", "issue_comment", "issues", "workflow_run",
-          "discussion", "discussion_comment", "pull_request_review",
-          "pull_request_review_comment"
+        inputs.ref != '' && !contains(fromJSON('[
+          "pull_request", "push", "merge_group",
+          "workflow_dispatch", "schedule", "release"
         ]'), github.event_name)
       shell: bash
       # Through the environment, not the script text — the same rule the run
--- a/action.yml
+++ b/action.yml
@@ -15,9 +15,13 @@ branding:
 # It is also a usable composite action in its own right, published to the GitHub
 # Marketplace. A repository that builds with Zuke can make this its first step —
 # `uses: zuke-build/zuke@<sha>` needs no prior checkout, since a remote action is
-# fetched by the runner. This repository cannot use it that way for hardening:
-# `uses: ./` resolves inside the workspace, so it would require the checkout it
-# is meant to precede.
+# fetched by the runner. That is how this repository's own generated workflows
+# now start, so the action is exercised on every run here rather than only by
+# strangers. It is the remote form that makes that possible: `uses: ./` resolves
+# inside the workspace and would need the very checkout this is meant to
+# precede, which is why the workflows inlined these steps until it was
+# published. Their pin comes from `build/action_version.json`, since this file
+# cannot pin the action to itself.
 #
 # The action is versioned by a plain `v1` tag (and the `v1.x.y` it points at),
 # which is the tag consumers write. That namespace is the action's alone: the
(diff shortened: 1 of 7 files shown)


totollygeek:
Addressing the three findings from this run, by id.

**`1mdlehy1n2nrl` — Potentially incomplete denylist for secret-bearing custom refs — accepted and fixed** (b986917)

Correct, and the incompleteness is worse than a missing entry: a denylist is right until GitHub adds an event, and then the guard stops firing *silently* rather than failing. Inverted to an allowlist of events whose ref can only come from someone who could push the same code directly — `pull_request`, `push`, `merge_group`, `workflow_dispatch`, `schedule`, `release` — so anything unrecognised is refused. The safe set is small and stable; the dangerous set is open-ended, which is the wrong one to enumerate.

A test now asserts the `!contains(...)` form and that the dangerous events are *absent*, so a future "simplification" back to a denylist fails rather than quietly narrowing the guard.

**`31avekkyncjz5` — Inconsistent Command Output Access — partly refuted, underlying risk fixed** (b986917)

The TypeError does not reproduce: `CommandOutput` is a class with `stdout` as a readonly property (`packages/core/src/shell.ts:85`), so destructuring it cannot throw. And the inconsistency is deliberate — `text()` prefixes a human-readable truncation notice, which for a newline-delimited tag list would parse as a tag.

But the finding points at something real underneath, so thanks for it: a truncated capture has *no symptom*. The newest release simply would not be in the list, `latestVersion` would compute from an older base, and the release would collide with a tag that already exists. The cap is 8 MiB against ~4.7 KB of tags so it is unreachable today; it now throws rather than depending on that staying true.

**`6nroiawy3hge` — Loose Exported Functions Violate Namespace Guidelines — refuted**

The guideline it cites (AGENTS.md #6, "Public API is task-shaped — no standalone utility functions") governs a **package's published surface**: "A package exposes its operations through a namespaced `*Tasks` object". `build/` is not a package. It is not in the `deno.json` workspace, not in `PACKAGES`, and not published to JSR — it is the internal helper layer behind `zuke.ts`'s targets, which AGENTS.md describes as exactly that ("The reusable helpers behind these targets live in `./build/*.ts`").

Every existing module there exports loose functions for the same reason: `collectActionPins` and `actionPin` (`build/action_pins.ts`), `installCli`, `publishOne` and `publishPackage` (`build/publish.ts`), `lintPrBody`, `checkSnippets`, `checkCoreFloors`, `assertLockUnchanged`. Wrapping these in a `*Tasks` object would make `build/action_release.ts` the only file in the directory that does, without exposing anything to a consumer.

Not suppressing it — it is a reasonable thing for a reviewer to flag, and the distinction between the published packages and the build layer is not visible from the diff alone.

PR #302, opened 2026-08-08
feat(core): generate the Zuke action as every job's prelude

totollygeek:
You asked why the generators were emitting steps the Zuke action already contains. They were, and the fix belongs in the generator — not in each caller, which is where my two earlier attempts put it.

## What changed

`@zuke/core`'s GitHub renderer emits one `zuke-build/zuke@<sha>` step where it used to emit harden-runner, checkout, and sometimes setup-deno. The `harden` and `checkout` options are unchanged as the authoring surface; they now become the action's inputs. A build declares the same things it always did — only the step count differs.

Net effect on this repository: **−44 lines of generated YAML**, and every workflow opens with one step. Six workflows, nine jobs.

## Why the generator and not the callers

My first two attempts added a `bootstrap` option to `@zuke/ai` and a `zukeAction()` helper to `build/workflows.ts`. Same idea, implemented twice, in consumers — and neither removed the duplicated pins. `@zuke/ai` now declares what it wants and **deletes the two pinned constants it carried**, which retires a trap this repo had already been caught by and documented in `zuke.ts`: those constants lived inside a published package, so a Dependabot bump landed in `ai-review.yml` and the next regeneration reverted it from the stale value. Someone had to hand-update the constant to match.

The adoption in this repo's own workflows then fell out of the generator change with **no edit to `build/workflows.ts`** — apart from folding `setup-deno` into the prelude via the new `denoVersion` option, which is the third of the three steps.

## Opt-outs, and one I got wrong first

- **Naming an action opts out.** If a caller pins `harden.action` or `checkout.action`, they asked for those specific actions; rendering one that runs neither would discard a pin they chose — silently, since the workflow would still look hardened. My first rule ignored them. The decision is made before the `pins` resolver fills anything in, because once it has, a resolver-supplied action is indistinguishable from a deliberate one and the two mean opposite things.
- **`harden: false` / `checkout: false` still fall back** to the separate steps. One action cannot do half of itself. This repo's `test` matrix depends on it — it deliberately exercises the bootstrap launchers — and `ci.yml` still shows a lone `actions/checkout` there.
- **`bootstrap: false`** declines entirely, for anyone who cannot consume a Marketplace action.

## Two inputs are always stated

Even when they match the action's defaults, exactly as the steps they replace stated them. `checkoutStep` had a deliberate comment about this — whether a checkout leaves a usable credential behind "is exactly the kind of thing that should not be left to a default" — and my first pass silently reversed it in the name of terser YAML. Whether a job records egress or enforces it, and whether it persists credentials, are what a reader most needs to see without opening the action.

## The pin

Resolved through the same `pins` hook as every other action, so this repository sources it from `build/action_version.json` — the file `actionRelease` writes. The constant behind it is a fallback and carries its version comment, so a bot's bump to a generated file reads as newer rather than being reverted by it. That is the same trap as above, one level up.

`actionPinCheck` joins the gate: it compares the pinned release against the working tree. Nothing else notices when those diverge, and the failure is quiet in both directions — an input added to `action.yml` is ignored rather than honoured until released, and a guard added to it protects nobody. Every test asserts against the file, so they stay green either way. It caught exactly this while I was building it, before `v1.0.1` existed.

## Tests

Core 72, `@zuke/ai` 22. New cases cover: the prelude resolving through `pins`, the pin carrying its own version comment without a resolver, `bootstrap: false`, per-job `harden: false` falling back, every option mapping onto its input name, and both security-relevant inputs staying explicit.

`./zuke ci` green, 18/18.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,5 +10,14 @@ updates:
     groups:
       actions:
         patterns: ["*"]
+    ignore:
+      # This repository's own action, which every generated workflow now opens
+      # with. Its pin lives in `build/action_version.json` and is written by the
+      # `actionRelease` target when the action is tagged; a Dependabot bump would
+      # edit the generated workflows instead, drifting them from that file. The
+      # pins Dependabot exists to bump are the ones *inside* `action.yml`, which
+      # it still sees and still bumps — and which reach consumers when that
+      # target cuts the next version.
+      - dependency-name: "zuke-build/zuke"
     commit-message:
       prefix: "ci"
--- a/.github/workflows/ai-review.yml
+++ b/.github/workflows/ai-review.yml
@@ -14,10 +14,12 @@ jobs:
     if: "${{ github.event.pull_request.head.repo.fork == false }}"
     timeout-minutes: 15
     steps:
-      - name: Harden and check out with Zuke
-        uses: zuke-build/zuke@7a62523bb4569e4d90e972dee74bf9cf09cd436f # v1.0.1
+      - name: Harden the runner
+        uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920
         with:
           egress-policy: audit
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
+        with:
           persist-credentials: "false"
       - name: AI review with Zuke
         run: ./zuke review
--- a/.github/workflows/ai-review.yml
+++ b/.github/workflows/ai-review.yml
@@ -14,12 +14,10 @@ jobs:
     if: "${{ github.event.pull_request.head.repo.fork == false }}"
     timeout-minutes: 15
     steps:
-      - name: Harden the runner
-        uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920
+      - name: Harden and check out with Zuke
+        uses: zuke-build/zuke@7a62523bb4569e4d90e972dee74bf9cf09cd436f # v1.0.1
         with:
           egress-policy: audit
-      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
-        with:
           persist-credentials: "false"
       - name: AI review with Zuke
         run: ./zuke review
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,15 +22,12 @@ jobs:
       contents: write
       pull-requests: write
     steps:
-      - name: Harden the runner
-        uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
+      - name: Harden and check out with Zuke
+        uses: zuke-build/zuke@7a62523bb4569e4d90e972dee74bf9cf09cd436f # v1.0.1
         with:
           egress-policy: block
-          allowed-endpoints: "deno.land:443 dl.deno.land:443 jsr.io:443 registry.npmjs.org:443 api.openai.com:443 api.github.com:443 github.com:443 codeload.github.com:443 objects.githubusercontent.com:443 release-assets.githubusercontent.com:443 cli.codecov.io:443 api.codecov.io:443 ingest.codecov.io:443 storage.googleapis.com:443"
-      - name: Checkout
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-        with:
           persist-credentials: "true"
+          allowed-endpoints: "deno.land:443 dl.deno.land:443 jsr.io:443 registry.npmjs.org:443 api.openai.com:443 api.github.com:443 github.com:443 codeload.github.com:443 objects.githubusercontent.com:443 release-assets.githubusercontent.com:443 cli.codecov.io:443 api.codecov.io:443 ingest.codecov.io:443 storage.googleapis.com:443"
           ref: "${{ (github.event.pull_request.head.repo.full_name == github.repository) && github.head_ref || '' }}"
       - name: Run ci with Zuke
         run: ./zuke ci
@@ -43,15 +40,12 @@ jobs:
     name: Core version floors
     runs-on: ubuntu-latest
     steps:
-      - name: Harden the runner
-        uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
+      - name: Harden and check out with Zuke
+        uses: zuke-build/zuke@7a62523bb4569e4d90e972dee74bf9cf09cd436f # v1.0.1
         with:
           egress-policy: block
-          allowed-endpoints: "deno.land:443 dl.deno.land:443 jsr.io:443 github.com:443 api.github.com:443 codeload.github.com:443 objects.githubusercontent.com:443 release-assets.githubusercontent.com:443"
-      - name: Checkout
-        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-        with:
           persist-credentials: "false"
+          allowed-endpoints: "deno.land:443 dl.deno.land:443 jsr.io:443 github.com:443 api.github.com:443 codeload.github.com:443 objects.githubusercontent.com:443 release-assets.githubusercontent.com:443"
       - name: Run coreFloorCheck with Zuke
         run: ./zuke coreFloorCheck
   test:
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -22,17 +22,11 @@ jobs:
           - macos-latest
           - windows-latest
     steps:
-      - name: Harden the runner
-        uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0
+      - name: Harden and check out with Zuke
+        uses: zuke-build/zuke@7a62523bb4569e4d90e972dee74bf9cf09cd436f # v1.0.1
         with:
           egress-policy: audit
-      - name: Checkout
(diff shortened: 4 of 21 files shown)


totollygeek:
Both findings by id. The high one is refuted on its facts, but it named a real gap that the medium one shares, and I have closed that.

**`20f1xk3uppdvu` — "New composite bootstrap action now owns checkout and runner hardening without visible equivalent safeguards" (high) — refuted**

The reviewer's stated reason is that the change "does not show the action implementation". That is true of the diff and not true of the repository: the action is `action.yml` at the root, and it runs `step-security/harden-runner` then `actions/checkout` — the same two actions, at the same pinned SHAs, that the steps it replaces used. I verified those SHAs are byte-identical to the ones the workflows inlined before this change.

The safeguards are not merely equivalent, they are the same code. What the diff changes is who invokes it. Point by point:

- **Hardening still precedes everything.** It is the first step inside the action, before the checkout, which is the ordering property that matters — an egress policy has to be installed before the code it governs is fetched.
- **`persist-credentials` still defaults to `false`**, and this change makes it *more* visible, not less: both it and `egress-policy` are now written into every generated job even when they match the action's defaults. That was a deliberate correction during this work — an earlier pass omitted them as redundant, which reversed an existing comment in `checkoutStep` saying whether a checkout leaves a credential behind "is exactly the kind of thing that should not be left to a default".
- **Nothing gained a privilege.** Job `permissions:` blocks are untouched by this diff. `contents: write` is on the same job it was on before.
- **The pin is a 40-character SHA**, so "if the new action is compromised" requires compromising a specific immutable commit, not the tag.

**`kvi7saiseoh4` — "Generated CI now defaults to a repo-owned action pin for the prelude, increasing supply-chain blast radius" (medium) — accepted as an accurate description of a deliberate trade**

This one is correct and I am not dismissing it. One action is now step 1 of every job in five workflows, and because it subsumes hardening, a compromised version of it could neuter its own egress policy before anything observes it. Reachable from there: `release.yml` holds an app private key, `publishJsr` holds JSR OIDC for 54 packages, `ci.yml` holds `contents: write`.

What holds it: the SHA pin is immutable, so moving the `v1` tag cannot reach this repository's own workflows; `.github/dependabot.yml` ignores the action so a bot cannot move the pin either; and only `actionRelease` writes `build/action_version.json`, from a clean master that matches its remote.

The trade is that the blast radius concentrates *and shrinks*: three pinned SHAs per workflow became one, in a file that is versioned, tested, and gated. The previous arrangement had already failed in exactly the way concentration is meant to prevent — `ai-review.yml`'s pins came from constants inside a published package, so a Dependabot bump landed in the file and the next regeneration reverted it. Someone had to hand-update the constant.

**The gap both findings gesture at, now closed**

Neither finding names it, but reviewing them made it obvious: nothing verified that the pinned commit's `action.yml` is the `action.yml` in the tree. `actionPinCheck` now does, and it is in the gate. Without it, an input added here is ignored rather than honoured until released, and a guard added here protects nobody — while every test stays green, because they assert against the file rather than the artifact the workflows run. It caught precisely that during this work, before `v1.0.1` existed.

Not suppressing either. `kvi7saiseoh4` in particular is worth leaving visible on the record.

totollygeek:
Second security run. New ids — the two from the previous run (`20f1xk3uppdvu`, `kvi7saiseoh4`) are gone from GitHub, since the reviewer upserts one comment. My earlier reply is the only surviving record of them.

**`2v8czaotors71` — "Jobs that contain any matching `uses:` step can now lose the bootstrap entirely" (medium) — correct, and fixed** (5f8bbf3)

This is a real defect in the guard I added an hour ago, and the reviewer found it before anyone hit it. I reproduced it first:

A job whose steps were a `run`, a checkout of `other/repo` into a subdirectory, then a build — vendoring a dependency — rendered with **no hardening and no checkout of its own repository at all**. The guard matched the vendored checkout anywhere in the list and concluded the job had built its own prelude. Silent, because the job still had a checkout in it and still looked plausible.

Now only the job's **first** step counts, because a prelude is positional: it is what a job does before anything else, and hardening that ran after a build step would not have hardened it. That still covers the case the guard exists for, since a job that builds its own prelude necessarily puts it first. Regression test added; `ai-review.yml` still renders exactly one prelude, and no other workflow changed.

**`21r358fq3i0bb` — "New default bootstrap action is trusted to perform checkout and hardening without an equivalent verification gate" (high) — refuted, and it is the same concern as `20f1xk3uppdvu` from the previous run, reworded**

The premise is that there is no verification gate. There is, and it is in this PR: `actionPinCheck` runs in `ci` and fails the gate when the pinned commit's `action.yml` differs from the one in the tree. It caught exactly that during this work, before `v1.0.1` existed. What the action does is not taken on trust either — it runs `step-security/harden-runner` then `actions/checkout` at SHAs I verified are byte-identical to the ones the workflows inlined before this change.

The rest as before: hardening still precedes the checkout inside the action, which is the ordering property that matters; `persist-credentials` still defaults to `false` and is now written into every generated job even when it matches, rather than omitted as redundant; no job's `permissions:` block is touched by this diff; and the pin is a 40-character SHA, so compromising it means compromising a specific immutable commit rather than moving a tag.

Not suppressing either.

totollygeek:
Third run. Both findings dropped to low; one was correct and is fixed.

**`3bvu3qo4hbmpr` — "Potential Unhandled TypeError in actionPinCheck" (low) — correct, and fixed** (eb93482)

Real, and I reproduced it before fixing. `actionPinCheck` read the SHA off `ACTION_PIN.ref.split("@")[1]` and used it. `ActionPinFile` describes the shape of a committed JSON file, and a type cannot hold a hand edit to its contents — so a ref with no `@` yields `undefined`, the `git show` against `undefined:action.yml` fails, and the branch that reports *that* failure calls `.slice()` on it:

```
TypeError: Cannot read properties of undefined (reading 'slice')
```

A TypeError from inside the gate is the least useful place to learn a file needs fixing. Fixed where the pin is read rather than where it is used, so every caller gets the check, and a malformed ref now names both the file and the form it must take. Test added, including the short-SHA case that `pinFor` already rejected on the write side.

**`36g0ym2409cpl` — "New self-hosted bootstrap action expands the trusted attack surface of every workflow" (low) — the same concern as `20f1xk3uppdvu` and `21r358fq3i0bb`, third wording**

Tracking it across runs since the reviewer upserts one comment and the earlier ids are gone from GitHub: this began as high (`20f1xk3uppdvu`), was reworded as high (`21r358fq3i0bb`), and is now low. The substance has not changed and neither has my answer — the action runs `step-security/harden-runner` then `actions/checkout` at SHAs byte-identical to the ones these workflows inlined before, `actionPinCheck` verifies the pinned commit against the tree on every gate run, the pin is a 40-character SHA rather than a tag, and no job's `permissions:` block is touched by this diff.

I am content to leave it visible at low rather than suppress it. It is a fair description of a deliberate trade: the surface concentrates into one action, and in exchange three pinned SHAs per workflow become one, in a file that is versioned, tested, and gated.

totollygeek:
Fourth run. One finding was correct and serious — it caught a check of mine that was passing while verifying nothing.

**`18jkvxd9lkkhx` — "CI gate skips the pin check on shallow checkouts, allowing the release integrity check to be bypassed" (medium) — correct, and fixed** (e24faa7)

Not theoretical. From the run log of this very PR, before the fix:

```
ℹ Pinned action commit 7a62523 is not in this checkout — skipping the comparison.
✔ actionPinCheck succeeded in 0.0s
```

`actionPinCheck` asked git for the pinned commit's `action.yml`. A CI checkout is shallow, so that commit is absent, the lookup failed, and it took the skip branch — one I wrote deliberately, reasoning that failing the gate over a fetch depth would be unfixable for whoever hit it. What that reasoning actually bought was a check that reported success while verifying nothing, in the one place the drift it exists to catch would land. It worked only on a full local clone, which is why every run I watched passed.

Severity is understated, if anything: this is the check added two commits earlier specifically to close the "a guard added to `action.yml` protects nobody until it is released" gap. It was closing nothing in CI.

The pin file now records a SHA-256 of `action.yml` as the release contains it, and the check compares that against the working tree. No git, no network, no dependence on fetch depth — the two things compared are the pin file and the file beside it, both always present. The recorded digest was verified against `git show 7a62523:action.yml` rather than just the current tree, so it describes the released artifact.

Proof it is no longer skippable — a directory with **no git history at all**, strictly worse than CI's shallow checkout:

```
✘ actionPinCheck failed
  action.yml has changed since v1.0.1, which is the release every generated workflow runs.
```

Line endings are normalised before hashing, so a CRLF checkout on Windows does not report a drift that does not exist.

**`3bsz0q4833zj1` — "Self-pinning bypasses normal pin resolution and can silently lock workflows to a stale in-repo action version" (medium) — the bypass is deliberate, and the "silently" is what `18jkvxd9lkkhx` just fixed**

The early return in `actionPin` is the point: `action.yml` is the manifest for every *other* action and cannot pin itself, and reading this pin back out of a generated workflow is the feedback loop that module exists to prevent. So it comes from `build/action_version.json`, which only `actionRelease` writes.

The real substance is "silently", and the two findings compound: staleness was detectable only by `actionPinCheck`, and that check was a no-op in CI. With the digest fix it now fails the gate on any divergence between the pinned release and the tree, on every run and every checkout depth. Dependabot is also configured to ignore this action, so a bot cannot move the pin sideways.

**`25zr7irkn0mq7` — centralised prelude (high) — fourth wording of the same concern**

Previously `20f1xk3uppdvu` (high), `21r358fq3i0bb` (high), `36g0ym2409cpl` (low). Answer unchanged: the action runs `step-security/harden-runner` then `actions/checkout` at SHAs byte-identical to the ones these workflows inlined, the pin is a 40-character SHA rather than a tag, no job's `permissions:` block is touched by this diff, and `persist-credentials` is now written into every generated job rather than omitted. The verification gate this finding says is missing is `actionPinCheck` — which, as of this commit, is one that actually runs.

PR #305, opened 2026-08-08
feat(ai): let core render the review workflow's prelude

totollygeek:
The follow-up promised in #302, now that core 1.34.0 is on JSR.

## What changed

`ai-review.yml` was the last workflow still spelling out a harden and a checkout step, and `@zuke/ai` was the last place holding pinned SHAs for them. It now declares what it wants and lets core render the prelude — one `zuke-build/zuke` step, same as every other workflow here.

Its declared core floor is what kept it there. `^1.25.0` has no `harden` or `checkout` on a job at all, so there was no way to declare a prelude rather than build one. That is why #302 left this out: raising the floor needed the core that #302 itself released.

## The constants are gone, and that is the point

Both `HARDEN_RUNNER` and `CHECKOUT` are deleted. They lived inside a **published package**, which made `ai-review.yml` the one generated workflow whose pins came from a release rather than from the file Dependabot edits. The failure mode was real and had already happened here: a bump landed in the workflow, the next regeneration wrote the stale constant back, and the upgrade was silently reverted — someone had to hand-update the constant to match what the bot had set.

Nothing in this package names an action SHA now. `zuke.ts` also stops passing its own pins in, since there is nothing left to override.

## Backward compatibility

`hardenRunner` and `checkout` keep working and keep their meaning: supplying either renders the two separate steps. That is the same rule core applies — naming an action means those specific actions were asked for, so a caller who pinned them does not silently get one that runs neither.

## Verification

- `coreFloorCheck` passes against the real JSR release — this is the check that failed on #302 and forced the split, so it is the one that matters here.
- `deno.lock` moves with the floor in the same commit, since a dependency range is what changed. One line.
- 22 tests in `packages/ai/tests/workflow_test.ts`.
- `./zuke ci` green, 18/18.

The generated diff is the whole story:

- two steps, `step-security/harden-runner@…` and `actions/checkout@…`
- one step, `zuke-build/zuke@7a62523… # v1.0.1`, carrying `egress-policy: audit` and `persist-credentials: "false"` as inputs

Every workflow in this repository now opens with the action, and the only remaining `actions/checkout` is the deliberate `harden: false` fallback in the test matrix.

### DIFF
--- a/.github/workflows/ai-review.yml
+++ b/.github/workflows/ai-review.yml
@@ -14,12 +14,10 @@ jobs:
     if: "${{ github.event.pull_request.head.repo.fork == false }}"
     timeout-minutes: 15
     steps:
-      - name: Harden the runner
-        uses: step-security/harden-runner@bf7454d06d71f1098171f2acdf0cd4708d7b5920
+      - name: Harden and check out with Zuke
+        uses: zuke-build/zuke@7a62523bb4569e4d90e972dee74bf9cf09cd436f # v1.0.1
         with:
           egress-policy: audit
-      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1
-        with:
           persist-credentials: "false"
       - name: AI review with Zuke
         run: ./zuke review
--- a/deno.lock
+++ b/deno.lock
@@ -1540,7 +1540,7 @@
     "members": {
       "packages/ai": {
         "dependencies": [
-          "jsr:@zuke/core@^1.25.0"
+          "jsr:@zuke/core@^1.34.0"
         ]
       },
       "packages/biome": {
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -10678,19 +10678,14 @@ interface AiReviewWorkflowSpec
     `zuke review` then works locally, where no workflow step exists to run.
   hardenRunner?: string
     The pinned `step-security/harden-runner@<sha>` to harden the runner with.
-    Defaults to a pin baked in here.
 
-    Pass it when the build sources pins from somewhere that stays current — a
-    generated workflow whose SHA comes from a constant in a published package is
-    a trap: a bot bumps the committed file, the next run regenerates it from the
-    stale constant, and the bump is silently reverted.
-
-    A bare `owner/repo@<sha>`, without the `# vX.Y.Z` comment the other
-    generated workflows carry: attaching one needs a core newer than this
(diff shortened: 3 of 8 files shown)


totollygeek:
Both findings by id.

**`10h7anaqwb76w` — "Removed pinned action constants for generated workflow output" (medium) — refuted, and the removal is the fix rather than the regression**

There is no unpinned output. The pin moved up a level: core resolves the prelude action through the same `pins` hook as every other action, and this repository feeds that from `build/action_version.json`. The generated `ai-review.yml` in this PR carries a full 40-character SHA and its version comment:

```
uses: zuke-build/zuke@7a62523bb4569e4d90e972dee74bf9cf09cd436f # v1.0.1
```

Core also holds a fallback constant for callers with no resolver, and it deliberately carries a version comment so a bot's bump to a generated file reads as *newer* than the constant.

The constants deleted here had no such protection, and that is why they went. They lived inside a published package, which made `ai-review.yml` the one generated workflow whose pins came from a release rather than from the file Dependabot edits. The failure had already happened in this repository: a bump landed in the workflow, the next regeneration wrote the stale constant back, and the upgrade was silently reverted — someone had to hand-update the constant to match what the bot had set. The comment on the old `hardenRunner` option described exactly this trap. Removing the constants is what closes it.

**`22hg0b63rbi5k` — "Replace explicit pinned checkout/hardening steps with unreviewed composite action" (high) — sixth wording of the same concern**

Previously `20f1xk3uppdvu`, `21r358fq3i0bb`, `36g0ym2409cpl`, `25zr7irkn0mq7`, `q8hhrswo11z4` — all now erased from GitHub by the upsert, which is why I keep listing them.

"Unreviewed" is the part that is checkable and false: the action is `action.yml` at the root of this repository, it has been through review on #299, #300 and #301, and it runs `step-security/harden-runner` then `actions/checkout` at SHAs byte-identical to the ones this workflow inlined before. The pin is a 40-character SHA rather than a tag, `actionPinCheck` verifies the pinned release against the tree on every gate run, and this diff touches no job's `permissions:` block.

---

Separately, and not from these findings: `actionPinCheck` as merged has a defect I am fixing next. It compares a whole-file digest, so it fails on **#304** — Dependabot's harden-runner bump to `action.yml` — and would fail on every future bump, since the release can only be tagged at a commit that exists after merge. The invariant worth enforcing is narrower: no generated workflow may use an input the pinned release lacks. A SHA bump changes no input names, so that version lets the bot through while still catching the case that actually broke.

PR #306, opened 2026-08-08
fix: check the inputs the workflows use, not a digest of the whole action

totollygeek:
`actionPinCheck` as merged blocks **#304** — Dependabot's harden-runner bump — and would block every future bump the same way. That is a defect I introduced, and it defeats the automation the pin manifest exists to serve.

## Why it was unpassable

The check compared a digest of the whole `action.yml` against the pinned release, so *any* change failed it. But a release can only be tagged at a commit that contains the change, and that commit does not exist until the PR has merged. So a PR touching `action.yml` could never go green — by construction, not by accident.

## The narrower invariant

What actually broke before was a workflow passing an input the pinned release does not have. GitHub does not fail a job for that — it warns and carries on — so the step runs with the input simply **absent**. A job asking for a specific `ref` silently gets the event's default checkout, which is how the gate's auto-fixer would have lost the branch it pushes to.

So the pin now records the input names its release accepts — `target`, `egress-policy`, `allowed-endpoints`, `persist-credentials`, `fetch-depth`, `ref`, `deno-version` — and the check asserts no generated workflow passes one outside that set. A bump to an action that `action.yml` wraps alters no input name, so the bot goes green while the case that actually breaks still fails, naming the input and the fix.

## Verified both directions, against the real repository

A simulated harden-runner bump, the #304 shape, reports that the workflows use six inputs and all are present in v1.0.1, and passes. A pin whose release predates `ref`, while the workflows pass it — the original H1 shape — fails, saying the generated workflows pass `ref` to an action version that does not accept it, and that GitHub ignores an undeclared input rather than failing on it, so those jobs would quietly run without it.

## Hardening, from review finding `17mnhkozf5nbd`

Both readers are line-based, and the shapes they do not handle are valid YAML: a flow mapping and a quoted key. Either would have read as a block with no entries, so the inputs inside would go uncounted — and an input the check cannot see is an input nobody checks. Both are now errors rather than empty answers, naming the line and what to do. This is the third check in this work that could have passed while verifying nothing, so I would rather over-correct.

Writing those guards turned up an adjacent bug: the manifest reader assumed a two-space indent, so a four-space manifest yielded no inputs at all, and `pinFor` would then have blamed the release when the problem was the indent. The indent is now taken from the first entry.

## Two things that fell out of writing it

Reading the workflows must be scoped to this action's own steps. Counting every `with:` in the file would compare `actions/checkout`'s inputs against this list and reject every workflow for having a checkout in it. Tested.

An empty input list is refused when pinning, since a release declaring nothing would reject every workflow that configures anything.

Also worth recording: the build regenerates the workflows on every run, so the check reads what the build *produces*. That is the right thing to check — the failure path is a build rendering a workflow that uses an unreleased input — but it means a hand-edit to a generated file is not what this catches, and my first attempt at testing it was invalid for exactly that reason.

22 tests in `tests/action_release_test.ts`. `./zuke ci` green, 18/18. Rebased onto master after #305, so the check now covers all six workflows including `ai-review.yml`.

Once this merges, #304 should go green on a re-run with no change to it.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,7 +97,12 @@ regenerate them in the same PR.
   future npm distribution (1:1 name mapping).
 - **No runtime dependencies.** The library is dependency-free; tests use a local
   assertion helper (`packages/core/tests/_assert.ts`) rather than a third-party
-  assert library so the suite runs with zero network access.
+  assert library so the suite runs with zero network access. This is a claim
+  about the **published packages** — every `packages/*/deno.json` — and it is
+  enforced by them declaring none. The build layer is separate and does have
+  dependencies: cspell and release-please are installed from npm on demand by
+  `installCli`, and the root `deno.json` imports `@std/yaml` for `build/`.
+  Nothing under `packages/` may import any of them.
 
 ### TypeScript 7 / `tsgo`
 
--- a/build/action_release.ts
+++ b/build/action_release.ts
@@ -289,12 +289,14 @@ export function workflowActionInputs(yaml: string): string[] {
   const names: string[] = [];
   let inStep = false;
   let inWith = false;
+  let withIndent: string | undefined;
   for (const line of yaml.replace(/\r\n/g, "\n").split("\n")) {
     if (/^\s*(?:-\s+)?uses:/.test(line)) {
       // Entering a step's `uses:` — this action's, or another's, which ends
       // any block we were reading.
       inStep = new RegExp(`uses:\\s*${ACTION_SLUG}@`).test(line);
       inWith = false;
+      withIndent = undefined;
       continue;
     }
     if (!inStep) continue;
@@ -314,8 +316,24 @@ export function workflowActionInputs(yaml: string): string[] {
     // the end dropped every input after it — silently, which is the failure
     // this whole check exists to avoid.
     if (/^\s*(#.*)?$/.test(line)) continue;
-    const entry = /^\s*([A-Za-z0-9_-]+):/.exec(line);
-    if (entry !== null) names.push(entry[1]);
+    const entry = /^(\s*)([A-Za-z0-9_-]+):/.exec(line);
+    // Indent decides what a line is, because a value can look like a key. The
+    // first entry sets the block's own indent; anything deeper belongs to that
+    // entry — the lines of a folded scalar, or a nested mapping — and anything
+    // shallower has left the block.
+    //
+    // Reading by pattern alone got this wrong in both directions at once: a
+    // folded `allowed-endpoints` whose lines are host:port pairs ended the
+    // block, losing every input below it, while a nested value contributed its
+    // own keys as inputs that were never passed.
+    const lead = entry?.[1] ?? /^(\s*)/.exec(line)?.[1] ?? "";
+    withIndent ??= lead;
+    if (lead.length > withIndent.length) continue;
+    if (lead.length < withIndent.length) {
+      inWith = false;
+      continue;
+    }
+    if (entry !== null) names.push(entry[2]);
     else {
       assertNotQuotedKey(line, "with");
       inWith = false;
--- a/build/action_release.ts
+++ b/build/action_release.ts
@@ -61,9 +61,21 @@ export interface ActionPinFile {
    * A digest also fires on a change that alters no input at all — a Dependabot
    * bump to one of the action's own pins, say — and that bump cannot be
    * released until it has merged, so the check blocked the very automation the
-   * pin manifest exists to serve.
+   * pin manifest exists to serve. See {@link digest}, which keeps what that
+   * approach was good at without the deadlock.
    */
   inputs: string[];
+  /**
+   * SHA-256 of `action.yml` as that release contains it.
+   *
+   * Not a gate. {@link inputs} covers what breaks a workflow *now*; this covers
+   * what a workflow cannot see at all — a guard added to the action, a step
+   * reordered — which reaches consumers only when the tag moves. Left
+   * unreleased that is release lag rather than a fault, so it is reported and
+   * not failed: failing it is what made the first version of this check
+   * unpassable on any pull request that touched the file.
+   */
+  digest: string;
 }
 
 /** Where the committed self-pin lives, relative to the repository root. */
@@ -220,11 +232,45 @@ export function declaredInputs(source: string): string[] {
   return names;
 }
 
+/**
+ * SHA-256 of the action's source, hex-encoded.
+ *
+ * Line endings are normalised first. A checkout on Windows can materialise the
+ * file with CRLF, and a value that moved with the checkout's line-ending config
+ * would report a drift that does not exist on one platform and not the other.
+ */
+export async function actionDigest(source: string): Promise<string> {
+  const bytes = await crypto.subtle.digest(
+    "SHA-256",
+    new TextEncoder().encode(source.replace(/\r\n/g, "\n")),
+  );
+  return [...new Uint8Array(bytes)]
+    .map((b) => b.toString(16).padStart(2, "0"))
+    .join("");
+}
+
+/**
+ * Whether the action has changed since its release in a way no workflow can
+ * see — a guard added, a step reordered, a pin bumped.
+ *
+ * Reported rather than failed. It is release lag, not a fault: the change
+ * reaches consumers when the tag moves, and the tag cannot move until the
+ * change has merged. Failing on it is what made the first version of this
+ * check unpassable on any pull request that touched the file.
+ */
+export async function releaseIsOwed(
+  pin: ActionPinFile,
+  workingTree: string,
+): Promise<boolean> {
+  return await actionDigest(workingTree) !== pin.digest;
+}
+
 /** The pin file's contents for a release of `sha` as `version`. */
 export function pinFor(
   sha: string,
   version: string,
   inputs: readonly string[],
+  digest: string,
 ): ActionPinFile {
   if (!/^[0-9a-f]{40}$/.test(sha)) {
     throw new Error(
@@ -252,7 +298,15 @@ export function pinFor(
         `would reject every workflow that configures anything.`,
     );
   }
-  return { ref: `${ACTION_SLUG}@${sha}`, version, inputs: [...inputs] };
+  if (!/^[0-9a-f]{64}$/.test(digest)) {
+    throw new Error(
+      `refusing to pin a digest that is not a SHA-256: ${
+        JSON.stringify(digest)
+      }. It is what tells the gate a release is owed, and an unrecognised ` +
+        `value would say so on every run.`,
+    );
+  }
+  return { ref: `${ACTION_SLUG}@${sha}`, version, inputs: [...inputs], digest };
 }
 
 /**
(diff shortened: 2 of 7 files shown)


totollygeek:
**`17mnhkozf5nbd` — "Line-based workflow parsing can miss or misread inputs in valid YAML edge cases" (low) — correct, and fixed** (6b2c2fc)

Right, and worth more than its severity. Both readers are line-based, and the shapes they do not handle are valid YAML — a flow mapping (`with: { ref: main }`) and a quoted key. Either would have read as a block with **no entries**, so the inputs inside it would go uncounted. Since the check compares what the workflows use against what the release accepts, an input it fails to see is an input nobody checks.

That is a false negative in the direction that matters, and it is the third instance of the same shape in this work: a check that passes while verifying nothing. The first skipped itself on shallow checkouts; the second treated any matching step as a self-built prelude. I would rather over-correct here than find a fourth.

Both are now errors rather than empty answers:

```
cannot read `with` written in flow style: with: { ref: main }
cannot read a quoted key under `with`: "ref": main. Unquote it, or the input it
names is not checked against the release.
```

I also found one adjacent bug while writing the guards: `declaredInputs` assumed a two-space indent, so a manifest written with four spaces yielded no inputs at all — and `pinFor` would then have rejected it with a message about the release when the problem was the indent. The indent is now taken from the first entry.

Verified each shape, including that the ordinary ones still read:

```
workflow: flow `with: {…}`     -> THROWS
workflow: quoted key           -> THROWS
manifest: flow `inputs: {…}`   -> THROWS
manifest: 4-space indent       -> ["target"]
manifest: normal 2-space       -> ["target","ref"]
```

I considered a real YAML parser instead. `jsr:@std/yaml` is in `deno.lock` but nothing imports it — an orphaned entry — so it would be a new dependency in practice, and core's own `yaml.ts` is deliberately writer-only. Given the inputs are files this repository's own renderer writes, failing loudly on what these cannot read seemed the proportionate answer. Happy to switch to a parser if you would rather not carry hand-rolled readers at all.

22 tests, `./zuke ci` green 18/18. Rebased onto master after #305.

totollygeek:
Three findings this run. One was correct and is fixed; two are rewordings I have answered before.

**`27gd3dctj7bya` — "Fragile YAML parsing in workflowActionInputs ignores inputs after comments or blank lines" (medium) — correct on comments, and fixed** (ab29f52)

Reproduced before fixing. A step passing `egress-policy`, then a comment, then `ref` reported only `egress-policy` — the `ref` was invisible to the check, and an input the check cannot see is an input nobody checks. Blank lines were already handled correctly; comments were not, which is the kind of asymmetry that survives review because the two look alike in the code.

Both are now skipped rather than read as the end of the block. Regression test covers a comment and a blank line in the same block.

It is not reachable from the current generator, which writes no comments inside a `with:`. That is not much of a defence. The value of this check is noticing what nobody was watching for, and it cannot do that while quietly reading less than the file says — which is the fourth time in this work I have written something that could pass while verifying less than it claimed.

**`3v3rzob4no0vg` — "Custom workflow YAML parsing can under-report used inputs in valid but unsupported shapes" (low) — this is `17mnhkozf5nbd` from the previous run, already fixed** (6b2c2fc)

Flow mappings and quoted keys now throw rather than returning fewer inputs, in both readers. The commit predates this review run.

**`2ed0orhqli4c5` — "Loose exported functions violate namespaced Tasks API design pattern" (low) — this is `6nroiawy3hge` from #301, refuted there and unchanged**

The guideline it cites (AGENTS.md #6) governs a **package's published surface**: "A package exposes its operations through a namespaced `*Tasks` object". `build/` is not a package — not in the `deno.json` workspace, not in `PACKAGES`, not published to JSR. It is the internal helper layer behind `zuke.ts`'s targets, which AGENTS.md describes as exactly that.

Every module there exports loose functions for the same reason: `collectActionPins` and `actionPin`, `installCli`, `publishOne`, `publishPackage`, `lintPrBody`, `checkSnippets`, `checkCoreFloors`, `assertLockUnchanged`. Wrapping these would make `build/action_release.ts` the only file in the directory that does, without exposing anything to a consumer.

---

Separately: the gate failure on this PR was `prBodyLint` rejecting three fenced code blocks in my own description — the same mistake I made on #301, and nothing to do with the code. Body rewritten as prose. Worth noting the first line of that log looks like a check failure but is my PR body being echoed back, since I had quoted the failure message verbatim.

23 tests, `./zuke ci` green 18/18.

totollygeek:
**`3dbpydt51b4rl` — "Workflow input parser terminates prematurely on multiline or nested YAML values" (low) — correct, and wrong in both directions at once** (51e1f6d)

Reproduced before fixing, and the two failures point opposite ways:

A folded scalar followed by another input reported only `allowed-endpoints` — `ref` was lost. The lines of that scalar are `host:port` pairs, so the first one looked like neither a key nor a continuation and ended the block. That is the silent under-report again: an input the check cannot see is an input nobody checks.

A nested mapping value reported `config`, `key`, `ref` — inventing `key` as an input no workflow passes. That direction is at least loud: it would fail the gate against a release that is perfectly correct.

Indent settles both, the same way it already settles the manifest reader. The first entry sets the block's own indent; deeper lines belong to that entry, shallower lines have left the block, and only lines at the indent are inputs.

**`15zjidi25ge3i` — "Line-based YAML parsing can miss inputs in valid workflow/manifest syntax" (low) — this is `17mnhkozf5nbd` and `3v3rzob4no0vg`, third wording**

Flow mappings and quoted keys throw in both readers as of 6b2c2fc.

---

Worth stating plainly, since this is now four parsing faults found in one reader — three of them silent under-reports:

- a comment ended the block (`27gd3dctj7bya`)
- flow style read as empty (`17mnhkozf5nbd`)
- a folded scalar ended the block (`3dbpydt51b4rl`)
- a nested value invented inputs (`3dbpydt51b4rl`)

Every one came from deciding what a line means by looking at the line, when the shape is what carries the meaning. The reader now uses indent for that, which is the property YAML actually defines blocks with, and each fault has a regression test.

I raised the alternative in an earlier comment and it is worth repeating now that the count is four: `jsr:@std/yaml` would remove this class of bug entirely. It sits in `deno.lock` already but nothing imports it, so it would be a genuine new dependency, and core's own `yaml.ts` is deliberately writer-only. I have kept the hand-rolled readers because the inputs are files this repository's own renderer writes — but if you would rather not carry them at all, say so and I will switch; the tests would carry over unchanged.

24 tests, `./zuke ci` green 18/18.

totollygeek:
**`2w74dwe5r72e2` — "The new check no longer detects non-input drift in the released action" (medium) — correct, and now addressed** (1f19f1d)

This names exactly what narrowing the check gave up, and it was a deliberate trade rather than an oversight — but the trade did not have to be total.

Input names catch what breaks a workflow *now*. They say nothing about a change no workflow can see: a guard added to the action, a step reordered. Those reach consumers only when the tag moves, and until then the released action simply does not have them — which is the `pull_request_target` guard case from #301 exactly.

The whole-file digest caught that, and was unpassable for it. The tag cannot move until the change has merged, so any PR touching `action.yml` was red by construction — which is why #304 is red today.

So: both, differing in what they do about it. The pin records a digest alongside the input names, and a mismatch is **reported, not failed**:

> ✔ the workflows use 6 input(s), all present in v1.0.1.
> ⚠ action.yml has changed since v1.0.1 in a way no workflow passes — a guard, a step, a bumped pin. Consumers get it when the tag moves: run `./zuke actionRelease` on master after this merges.

That is the honest description of the state. It is release lag, not a fault, and the message says what resolves it. Failing would only mean nobody can merge the thing that resolves it. Verified both ways: silent when the tree matches the release, warning when it does not, exit zero either way.

**`o394kebogad4` — "Line-based YAML parsing can silently miss inputs in valid workflow forms" (medium) — fifth wording of the parsing concern**

Previously `17mnhkozf5nbd`, `3v3rzob4no0vg`, `27gd3dctj7bya`, `3dbpydt51b4rl`. Four concrete instances have been found and fixed, each with a regression test: a comment ended the block, flow style read as empty, a folded scalar ended the block, a nested value invented inputs. The reader now decides by indent, which is the property YAML actually defines blocks with, and refuses outright the two shapes it cannot read.

I take the generic restatement seriously rather than as noise — the finding has been right four times. My standing offer holds: `jsr:@std/yaml` would remove the class entirely. It is in `deno.lock` but unimported, so it is a genuine new dependency, and core's own `yaml.ts` is deliberately writer-only. Say the word and I will switch; the tests carry over unchanged.

**`3dbpydt51b4rl`** was fixed in 51e1f6d, before this run.

25 tests, `./zuke ci` green 18/18.

totollygeek:
Switched to `@std/yaml`, per the discussion above. Both readers are now a walk over parsed objects rather than a line scan.

**Why now:** five findings, four fixes, and the fifth was a general restatement that the class was still open — which it was. Three of the four were silent under-reports, and that is the one outcome this check cannot afford, since an input it fails to see is an input nobody checks.

**What went away:** the two guards that refused flow style and quoted keys. Those shapes now simply work, which is what refusing them was standing in for. Net **−81 lines** in `build/action_release.ts`.

Every shape that previously broke, now parsed correctly:

| shape | before | now |
| --- | --- | --- |
| plain | ok | `["egress-policy","ref"]` |
| comment inside `with:` | dropped `ref` | `["egress-policy","ref"]` |
| folded scalar | dropped `ref` | `["allowed-endpoints","ref"]` |
| flow mapping | refused | `["egress-policy","ref"]` |
| quoted key | refused | `["egress-policy","ref"]` |
| nested value | invented `key` | `["config","ref"]` |
| four-space manifest | returned nothing | `["target"]` |

**On the dependency-free commitment**, which is worth being precise about since it is a headline claim:

It is about the **published packages**, and it is untouched. Verified from the lock: no workspace member gains `@std/yaml` — it is recorded against the root alone, where `build/` and `zuke.ts` live. Nor is it the first build-layer dependency; `installCli` already pulls cspell and release-please from npm.

That distinction was doing real work and was written down nowhere, so AGENTS.md now states which half of the repository the claim covers, and that nothing under `packages/` may import any of them.

24 tests. `./zuke ci` green, 18/18.

totollygeek:
**Security review: 0 findings** — first clean run on this PR. The centralisation concern that ran through five wordings (`20f1xk3uppdvu`, `21r358fq3i0bb`, `36g0ym2409cpl`, `25zr7irkn0mq7`, `q8hhrswo11z4`) is gone with it.

**`3dbpydt51b4rl` — "Workflow input parser terminates prematurely on multiline or nested YAML values" (low) — stale; the code it describes no longer exists**

Same id as the finding fixed in 51e1f6d, re-reported. It is doubly moot now: that commit made the reader indent-based, and 0e67115 removed line reading altogether. Verified nothing of the kind survives — no `split("\n")`, no per-line loop, no `inWith`/`withIndent`/`inBlock` state anywhere in the module. `workflowActionInputs` parses the document and walks `jobs[].steps[]`, taking the `with` keys of steps whose `uses` names this action.

"Terminates prematurely" has no mechanism left to describe: there is no scan to terminate.

---

That closes every finding raised across this PR's six review runs:

| finding | outcome |
| --- | --- |
| `2w74dwe5r72e2` non-input drift undetected | fixed — reported as a warning, not failed |
| `27gd3dctj7bya` comment ended the block | fixed, then obsoleted by parsing |
| `3dbpydt51b4rl` folded scalar / nested value | fixed, then obsoleted by parsing |
| `17mnhkozf5nbd` / `3v3rzob4no0vg` / `o394kebogad4` flow style, quoted keys | fixed, then obsoleted by parsing |
| `2ed0orhqli4c5` loose exported functions | refuted — `build/` is not a published package |

Four were real defects in code written during this PR, three of them silent under-reports. The reviewer was right every time it pointed at a mechanism.

All 10 checks pass. 24 tests, `./zuke ci` green 18/18. Ready when you are — and once this merges, **#304 goes green on a re-run with no change to it**, which is what the whole PR is for.

PR #312, opened 2026-08-09
feat: cut the action release from CI, and propose its pin as a pull request

totollygeek:
Cuts the Marketplace action's release from CI and proposes its pin as a pull request, so the only manual step left is the Marketplace tick GitHub gates behind a browser 2FA confirmation.

## Why a pull request rather than a push

The ruleset on master requires one. The tags are the half that reaches consumers and they are pushed directly; the pin — `build/action_version.json` plus the workflows regenerated from it — is this repository catching up with its own release, and it can wait for review without anyone being stuck.

## The credential

This job mints a GitHub App installation token rather than using `GITHUB_TOKEN`. That is not a preference. There is no `workflows` permission to grant `GITHUB_TOKEN`, so it may not write the regenerated `.github/workflows/*`, and a pull request it opened would trigger no workflows — so the required checks could never run and the proposal could never be merged. The job itself is granted `contents: read` and holds no `GITHUB_TOKEN` at all.

Every write goes through the REST API, so nothing needs a credential in `.git/config` where it would outlive the step that used it. Git is still used, but only to read: the tag list, the diff, and the SHA a tag resolves to. The key is withheld from the `generate-ci` subprocess, which has no use for it.

## What the earlier draft got wrong

It carried its own copy of the GitHub API transport, written before the fixes that landed in `@zuke/gh` — no ref validation, no path encoding, a bare catch reporting an authorisation failure as a missing ref, type assertions on responses. That module is deleted; the operations come from `GhTasks.commit`, `.tag` and `.pullRequest`. The duplication is what let the two copies diverge, so the shared transport now lives in one place, `packages/gh/src/api.ts`.

Two defects worth naming, both silent.

A release is two tag writes with nothing binding them together. If the move of `v1` failed after `vX.Y.Z` was created, the release was half-published — and nothing would ever notice, because `changedSince` then compares against a tag naming that very commit and answers unchanged on every later run. Both paths reported success while every `@v1` consumer stayed on the older action, which is precisely the harm the tag exists to prevent. The pointer is now reconciled against the newest release on every run, ahead of the gate that decides whether anything is owed, because that gate says no in exactly this state.

The pin branch was force-reset on every push to master while its proposal was open, not once — so unrelated traffic would discard a commit a maintainer pushed to make it mergeable, and move the head out from under a review. It now looks for an open proposal before writing anything.

Smaller ones: the pull-request lookup filters on base as well as head, since GitHub allows several proposals open from one branch to different bases; a failing lookup no longer replaces the error that says what was refused; `.replace()` without `.from()` is refused rather than ignored; and the app-token mint validates and encodes the owner and repository it interpolates into a path that carries the private-key JWT.

## Known limitation

If `action.yml` changes again while a pin proposal is still open, a second one is opened for the newer version and the older is left alone. Merging them out of order regresses the pin until the next push to master, where the retry path re-proposes the correct one. Closing the superseded proposal automatically needs an operation this does not add.

## Testing

`./zuke ci` green, 18/18. 32 tests over the release state machine, including the half-published release, the aligned pointer, the first release of a major, and a release tag that will not resolve; 4 pinning the generated job's credential shape; and the `@zuke/gh` suites covering the new pull-request operation and `.replace()`.

The `isCI()` half of this target executes for the first time in production, against a tag namespace with no undo. That is worth knowing before merging.


totollygeek:
Both findings, and one fix answers both — b16aba0.

**`2eyo60uhydy5t` — "Release workflow step persists checkout credentials unnecessarily" (medium) — half right, and the half that matters is fixed**

Not removable, I checked rather than assumed: the job pushes a branch, and `git push` needs a credential in git config to do it. Dropping `persist-credentials` breaks the feature outright.

What *was* wrong is that the job audited egress while holding a write-scoped token. This repository does not do that anywhere else — the gate job blocks, and `syncWebsite` blocks, and `syncWebsite` is precisely the job this one is modelled on: push a branch, open a pull request, with a token. Mine was the odd one out.

It now blocks against the same endpoint list. The credential still exists; what changes is that anything reading it out of the environment has nowhere to send it.

**`3dnk9v0tnnexy` — "CI release job now pushes a branch and opens a PR with write credentials" (medium) — accurate description of a deliberate choice, and the same fix bounds it**

Worth stating why this shape and not another, since the finding is really about whether the privilege is justified.

Master has an active ruleset requiring a pull request, whose only bypass is a repository role. The alternative to opening a PR is granting the bot that bypass so it can push to master directly — a standing hole in that protection, for a change that happens a few times a year. Opening a pull request is strictly less privilege than the alternative, and it leaves the change reviewable, which a direct push would not.

The credential's blast radius is now bounded by the egress block, which is the part of this that was genuinely missing.

**One detail worth recording:** the endpoint list is renamed, since two jobs share it and its old name described only the first. Verified it is a rename and not a change — regenerating alters exactly one line across the workflows, the new job's, and the website sync's own allowlist is byte-identical.

`./zuke ci` green, 18/18.

totollygeek:
**`26topmdip8rk2` — "Persisted checkout credentials can be reused by build steps before branch push" (medium) — accurate, and I am not going to claim the egress block solves it**

The narrowing is correct and sharper than the previous wording. The credential is in git config from checkout onward, and `./zuke actionRelease` loads the whole build before it reaches the push — so every module in that graph can reach it for the duration.

The egress block bounds *exfiltration*, not *use*. The token is a GitHub credential and `github.com` is necessarily on the allowlist, so code that read it could still act as the token against this repository. I said in my previous comment that the block means it "has nowhere to send it" — that is true of an arbitrary host and not true of the one that matters. Worth correcting rather than leaving to stand.

**What it is, then:** the same trade this repository already makes on its gate job, which holds `contents: write`, sets `persist-credentials: true` for the lint fixer's push, blocks egress, and runs the entire build in between. This job is that pattern applied to a second case, not a new risk class — and the module graph it loads is first-party code plus `@std/yaml`.

**What would actually remove it**, if you want that: drop `persist-credentials` and create the branch, the commit and the pull request through the API with `gh`, which needs the token in the environment for `gh` alone and never writes it to git config. That is roughly seven API calls — one per file plus the ref and the pull request — against one `git push`, and it would leave the gate job as the only place the pattern remains.

I have not done it here because it is a different shape of change from wiring the job up, and because it would leave this repository inconsistent with itself in the other direction. Happy to do it as a follow-up covering both jobs, which is where it would actually pay: the gate job holds the same credential for a much longer build with a much larger dependency graph.

All 10 checks pass. `./zuke ci` green, 18/18.

totollygeek:
**`2y2yj4phucoli` — "Branch name is interpolated into GitHub ref creation without explicit sanitization" (medium) — confirmed, and worse than the wording suggests. Fixed in #313.**

I traced it before deciding, and then found the real version of it.

**In this PR's call path it is not reachable.** The branch comes from `pinBranch(version)`, which calls `majorTag()` and throws on anything that is not `v<major>.<minor>.<patch>`; the version itself is machine-generated as `v${major}.${minor}.${patch + 1}` from integers parsed by an anchored regex. A test asserts `pinBranch("; rm -rf /")` throws.

**But the finding is about the library function, and that is the right thing to be about.** `proposeChange` takes `branch: string` from any caller and validates nothing, and #313 moves exactly this logic into `@zuke/core`, where arbitrary callers exist. A library that is safe only when every caller validates is not safe.

**And the consequence is not a malformed name.** These names go into request *paths*, and URL normalisation resolves `..` before the request is sent. Verified:

> `/repos/acme/app/git/ref/heads/../../../user/repos` → `/repos/acme/app/user/repos`

A branch name could silently redirect the call to a different endpoint, with a write-scoped token attached. That is a genuine defect in code I wrote today, and the reviewer got to it first.

**Fixed in #313**, where the durable copy lives: every entry point validates by git's own ref rules and refuses **before anything is sent** — a request that goes out and fails has already carried the token somewhere unintended, so the test asserts zero calls were made. It also asserts the URL normalisation itself, so a future reader sees why the check exists rather than just that it does.

I have not duplicated the guard into this PR's `build/propose_change.ts`, because #313's follow-up deletes that file in favour of core's. If you would rather merge this one first and leave a gap of one PR, say so and I will add it here too — the call path is safe either way, so it is a question of how long the unvalidated library function sits in the tree.

totollygeek:
**Marked draft.** Two adversarial reviews found blockers that compound into a silent, unrecoverable half-release. Converting back is one click once these are addressed; I would rather it not be merged in the meantime.

## Blocker 1 — the pull request cannot work with `GITHUB_TOKEN`

Three separate reasons, any one sufficient:

- A pull request opened with `GITHUB_TOKEN` **triggers no workflow runs**. That is documented GitHub behaviour, not a setting. The pin PR would land with zero CI and could not satisfy required checks.
- `POST /pulls` is refused unless "Allow GitHub Actions to create and approve pull requests" is enabled, which is off by default.
- The PR contains `.github/workflows/*.yml`, and `GITHUB_TOKEN` cannot write those. I checked the `permissions:` key list: there is no `workflows` key to grant, so this is not configurable.

The precedent is already in this repository — `syncWebsite`, the other PR-opening job, mints a GitHub App token rather than using the ambient one.

## Blocker 2 — a failure after the tag push is permanent, and reports success

This is the one that worries me. Sequence:

1. Run 1 writes the pin, cuts `v1.0.3`, moves `v1` — consumers now have the release.
2. `proposeChange` fails (blocker 1, a 422, any transient error). The target fails.
3. Run 2 computes `changedSince(v1.0.3)` — against the tag that now points at that very commit — and gets "unchanged since v1.0.3. Nothing to release."

The pin and the regenerated workflows are never proposed again, by any run. The build goes green while the repository stays on the old pin indefinitely. There is no recovery short of deleting the tag by hand.

My own comment in `releaseAction` claims the pin-before-tag ordering makes a failure "loud rather than looking like success". That reasoning covers only the window between writing the pin and cutting the tag. Everything after the tag is swallowed by the guard.

## Blocker 3 — a retry cannot fix it either

`pinBranch(version)` is deterministic, so a re-run after a failed `POST /pulls` hits 422 "Reference already exists" on the branch that run 1 created, and throws before reaching the pull request.

## Also confirmed

- Concurrent unmerged pin PRs revert each other: each is branched from master, so merging an older proposal after a newer one rolls the pin backwards and fails `generate-ci --check` on master.
- This PR's `build/propose_change.ts` has **no ref validation at all**, while #313 added it — and this is the copy that runs in CI with a write-scoped token. The `%2e%2e` bypass fixed in #313 was never closed here.
- `moveTag`'s bare `catch` reports an authorization failure as a ref conflict. Fixed in #313; not here.

## The shape of the fix

Land #313 first, add a pull-request operation to `@zuke/gh`, and rewrite this PR to call `GhTasks.commit`/`GhTasks.tag` rather than carrying a second, weaker copy. That removes the duplication, puts the ref validation on the path that actually runs, and is the placement this work should have had from the start.

Two decisions are the maintainer's rather than mine: whether this job gets the app token (the only credential that can commit workflow files and trigger CI), and whether `changedSince` should compare the committed pin against the newest tag rather than `action.yml` alone, so a failed proposal is retried instead of being declared complete.

totollygeek:
Rewritten. The draft carried its own copy of the GitHub API transport — written before the fixes that landed in `@zuke/gh`, so it had no ref validation, no path encoding, a bare catch reporting an authorisation failure as a missing ref, and type assertions on responses. That module is deleted rather than repaired: the duplication is what let the two copies diverge, so the shared transport now lives in one place.

Four adversarial reviews ran against this branch. Two findings were serious, and both were silent.

**A half-published release was permanent and invisible.** A release is two tag writes with nothing binding them: `vX.Y.Z` is created, then `v1` is moved. If the second failed, `v1` — the tag consumers actually use — stayed on the previous release. Nothing would ever have noticed: once `vX.Y.Z` exists, `changedSince` compares against a tag naming that very commit and answers "unchanged" on every later run, so the release path is never re-entered; once the pin lands, the retry path stops too. Both report success, while every `@v1` consumer silently keeps the older action — which is exactly the harm this module's own docstring says the automation exists to prevent. The old code asserted the opposite in a comment: *"No tag is cut or moved here — they are already right."*

The pointer is now checked against the newest release on **every** run, ahead of the gate that decides whether anything is owed, because that gate answers "no" in precisely the state this repairs. That required the tag deps to name the commit they write rather than implying HEAD — repointing a major tag at an earlier release is not something the old signature could express, which is worth flagging as a trap for anyone calling `push` from a non-HEAD context.

**The pin branch was force-reset on every push to master while its proposal was open**, not once. The retry gate fires whenever the committed pin disagrees with the newest tag, and that stays true until the pull request merges — so unrelated traffic would have discarded any commit a maintainer pushed to make it mergeable, and moved the head out from under a review. My justification comment claimed "only this target ever writes it", which is true of the one-shot retry and false in the steady state. It now looks for an open proposal before writing anything.

Smaller confirmed defects, each with a regression test: the pull-request lookup filters on base as well as head, since GitHub permits several proposals open from one branch to different bases and the duplicate 422 is raised for the pair; a failing lookup no longer replaces the error that says what was refused; `.replace()` without `.from()` is refused rather than silently ignored; `needs:` gained `if: ${{ !cancelled() }}`, because a bare `needs:` would have let a release-please failure *skip* this job, and a skipped job is not a red workflow; and the app-token mint now validates and encodes the owner and repository it interpolates into a path carrying the private-key JWT — the same defect class caught twice before, one file over.

`2gbhthjuauvb6` from this run is fixed in e06be46: the lookup verified the returned head but not the base, an asymmetry when both are what make it the same proposal.

**Two things a reviewer should weigh rather than take from me.** The app token is not a preference — `GITHUB_TOKEN` has no `workflows` permission to grant, so it cannot write the regenerated workflow files, and a pull request it opened would run no checks and could never satisfy the ruleset. That does mean a second job holding the app private key, which was raised as a concern on #298; it is withheld from the `generate-ci` subprocess, and the job holds no `GITHUB_TOKEN` and only `contents: read`. And the `isCI()` half of this target runs for the first time in production against a tag namespace with no undo.

`./zuke ci` green, 18/18. 32 tests over the state machine including the half-published release, 10 over the pull-request operation, 4 pinning the generated job's credential shape.

totollygeek:
`lpimal2fs7bp` — "new action-release job exposes GitHub App key to an additional workflow path" — accurate, and not something I should quietly fix. It is the tradeoff this PR makes, and it is the maintainer's call rather than mine.

The key now reaches a second job. That is real, and it is the same concern raised against `syncWebsite` on #298. What makes it not optional is that the alternative does not work: `GITHUB_TOKEN` has no `workflows` permission to grant, so it cannot write the regenerated `.github/workflows/*`, and a pull request it opened triggers no workflows, so the required checks never run and the proposal can never be merged. There is no third credential.

What is bounded rather than removed: the job holds no `GITHUB_TOKEN` and is granted only `contents: read`; egress is blocked to a list that no longer carries the npm registry, since nothing in this job's module graph reaches it; the key is withheld from the `generate-ci` subprocess, which is a fully-privileged `deno run -A` with no use for it; and the minted token is narrowed to this repository with the three permissions the target actually exercises.

What is not bounded, stated plainly: the mint happens inside build code running under `-A`, and `api.github.com` is necessarily reachable while the minted token carries `pull_requests: write` on a public repository. A compromised dependency in the build graph therefore has a publication channel without needing any non-allowlisted host. The egress block is defence in depth here, not a boundary — the same is true of `syncWebsite` today.

If that is not an acceptable widening, the shape that avoids it is a separate workflow triggered on the tag push, holding the key in a job that runs no build code. Worth deciding deliberately either way; I have not built it, because it trades this exposure for a second automation surface and that is a judgement about the repository rather than about the code.

All checks pass. The remaining `mergeStateStatus: BLOCKED` is review approval.

PR #317, opened 2026-08-10
feat(core): expose per-target outcomes on the target context

totollygeek:
## Why

A target could read another target's published metadata, but nothing told it what the other target actually **did**. An aggregate target that reports one verdict for a fan of checks had no way to compute that verdict from inside the run.

This is the first piece of the durable merge-gate work: a gate that posts a single required status context has to know which checks passed, which failed, and which were skipped.

## What changed

Two members on the target context:

- `outcomeOf(name)` — what another target in this run did, or `undefined` if it has no outcome yet.
- `outcomes()` — every outcome settled so far, keyed by target name.

The status is the run record's vocabulary, so a target whose body ran and one served from the incremental cache both read `succeeded` — the distinction the record keeps, and the one a body branching on "did it work" wants.

## The part worth reviewing

**Outcomes are deliberately not read from the run record.**

Every settlement goes through the writer as `void env.writer?.markTargetSettled(...)`, and the mutator only runs once the serialized chain drains through a store round trip. So a record read taken right after a target settles still says `running`. An aggregate reading that would see a finished failure as not-yet-failed — for a gate posting a required context, "not failed yet" reads as green.

So outcomes come from a status map on the run environment, written synchronously at every settle site, in the record's vocabulary. The record backfills only what the map has never seen, which is what an earlier process settled before a resume.

There is a regression test for exactly this, and it is mutation-checked: reordering the lookup to consult the record first makes it fail, reporting `running` for a dependency that had already succeeded.

Compensations get the same seam, built from the record the walk is already reading, so a cleanup can ask what happened to the run it is undoing even in a process that never executed any of it.

## Tests

- Unit, 5 cases: the view built from a status and a record row, the row contributing detail but never its status, absent detail staying absent rather than undefined-valued, and a record becoming one entry per settled target with pending rows omitted.
- Integration, 5 cases: reading a passed and a skipped dependency, the staleness guard above, outcomes surviving a resume for a target this process never ran, parity without a state store, and outcomes listing only what has settled.

Full gate green locally, 18 of 18 targets.

## Known limitation, addressed separately

A target whose dependency **failed** never becomes ready, so it cannot report on that failure. `.always()` exempts a target from the build-wide halt, not from waiting on its own dependencies. That makes the aggregate gate above unable to post a red verdict, and is fixed in a follow-up PR.

### DIFF
--- a/docs/run-context.md
+++ b/docs/run-context.md
@@ -25,6 +25,8 @@ class Deploy extends Build {
 | `signal`     | `AbortSignal`       | Aborted when the run is cancelled (see below).                       |
 | `state`      | `TargetStateHandle` | Durable per-target metadata — see [Durable run state](./state.md).   |
 | `stateOf`    | `(t) => …`          | The state handle of **another** target — read a dependency's published metadata (e.g. a wait's result). |
+| `outcomeOf`  | `(t) => …`          | What another target in this run **did** — `succeeded`, `failed`, `skipped`, … or `undefined` if it has none yet. |
+| `outcomes`   | `() => ReadonlyMap` | Every outcome settled so far, keyed by target name. |
 | `signals`    | `ReadonlyMap`       | Payloads of external signals received so far (see [waits](./orchestration.md)). |
 | `dryRun`     | `boolean`           | `true` when the run is a dry run (bodies don't execute in a dry run). |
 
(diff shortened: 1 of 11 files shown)


totollygeek:
### Review findings

**`1xs7ip1kjryk8`** — "Potential prompt-injection text embedded in generated documentation" at `docs/run-context.md:35` — **false positive, no change made.**

Line 35 of that file is the `## Cancellation` heading, which this PR does not touch. The section this PR adds sits above it and is ordinary prose plus one type-checked snippet; there is no instruction-like text, and nothing in that file is fed to a model at build time. Not suppressing it in `zuke.ts` — one sighting on an unrelated line is not yet a pattern worth dismissing permanently.

**`1krsgewmjc5d4` / `2p0b8f9zctul8`** (raised against the stacked #318, same root cause) — "workflow dispatch is now persisted after the API call" — **real signal, wrong cause; fixed by rebasing.**

This branch was cut before #315 merged, so it still carried the old dispatch-then-persist order in `packages/gh/src/workflow.ts` — which, diffed against current `master`, reads as reverting that fix. A three-way merge would not actually have reverted it (this branch never modifies that file, so `master`'s version wins), but the stale base made the diff say otherwise. Both branches are now rebased onto `master` and carry the persist-before-dispatch order.

PR #318, opened 2026-08-10
fix(core): let an always target wait for its dependencies to settle

totollygeek:
> Stacked on #317 — the base retargets to master once that merges. The diff to review is the last commit.

## Why

The always modifier exists for work that must happen when something went wrong: cleanup, teardown, or an aggregate that reports one verdict for a fan of checks. It was unusable for the failure case, which is the case it exists for.

Readiness required every dependency to be in the done set, and a failed target never enters it. So a target that depended on the work it was reporting on was held back by exactly the failure that should have triggered it, and the run settled it as skipped. With lint succeeding and test failing, the gate came out skipped and never ran at all.

Its own documentation already said it "still waits for its own dependencies to **complete**". A failed target has completed. The code read that as "succeeded", so this is the documentation and the behaviour disagreeing rather than a deliberate design.

This blocks the durable merge-gate work directly. A gate that posts a required status context could post green when everything passed and **nothing at all** when anything failed. For a required context a missing verdict is worse than a red one: the pull request is not red, it is stuck with no answer, and only an org owner editing branch protection can unstick it.

## What changed

An always target now waits for its dependencies to **settle** rather than to succeed. A failure releases it, the same as a pass or a skip, and it reads what actually happened through the target context's outcome accessors added in #317.

Scoped deliberately:

- An ordinary dependent is **still** blocked by a failed dependency. The new set is consulted only for always targets.
- An always target that passes still does not rescue a failed build.
- A dependency parked at a wait is **not** settled. It has not finished and is going to be resumed, so releasing the target there would have it report on a run still in progress, and the resume would then find it already done.

## Compatibility

This is an observable behaviour change to a public modifier: an always target with a dependency that fails now runs where it used to be skipped. That is the documented contract, and no existing test pinned the old behaviour. A cleanup body that assumed its dependency succeeded should check the dependency's outcome.

## Tests

Five integration cases: the aggregate-gate shape reporting a failure it depends on, the failure's message reaching it without a state store, an ordinary dependent still being blocked, a parked wait holding it back until the resume, and the all-passed path still working.

Full gate green locally, 18 of 18 targets.

### DIFF
--- a/docs/authoring.md
+++ b/docs/authoring.md
@@ -19,7 +19,7 @@ body, which is required before the target can run.
 | `.onlyWhen(condition)`      | `(c: () => boolean \| Promise<boolean>) => this`      | Run only when the condition holds, else skip.                                                      |
 | `.requires(...params)`      | `(...p: Parameter[]) => this`                         | Fail the target unless these parameters are set.                                                   |
 | `.proceedAfterFailure()`    | `() => this`                                          | Keep the build going if this target fails.                                                         |
-| `.always()`                 | `() => this`                                          | Run for cleanup even after the build has failed.                                                   |
+| `.always()`                 | `() => this`                                          | Run even after the build has failed; waits for dependencies to settle, not succeed.                |
 | `.unlisted()`               | `() => this`                                          | Hide the target from `--list`/`--help`.                                                            |
 | `.readOnly()`               | `() => this`                                          | Advertise the target as query-only over [MCP](./mcp.md) (`readOnlyHint`).                          |
 | `.cacheKey(fn)`             | `(fn: () => string \| Promise<string>) => this`       | Extra (non-file) input to the cache fingerprint.                                                   |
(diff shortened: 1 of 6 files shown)


totollygeek:
### Review findings

**`1krsgewmjc5d4`** (security, medium) and **`2p0b8f9zctul8`** (generic, medium) — "dispatch is now persisted after the API call, enabling duplicate dispatch on crash" — **real signal, wrong cause; fixed by rebasing.**

Both findings point at `packages/gh/src/workflow.ts`, which this PR does not touch. The cause was a stale base: this branch and its parent #317 were cut before #315 merged, so they still carried the old dispatch-then-persist order. Diffed against current `master` — which now has the fix — that reads as a reversal of it.

A three-way merge would not have reverted anything, since neither branch modifies that file and `master`'s version wins. But the reviewer was reading a diff that genuinely showed the weaker ordering, so the signal was correct even though the conclusion about this PR was not.

Both branches are now rebased onto `master`. The marker is persisted before the dispatch call on this branch, and the reviewer should see the same on the next run.

The direction of concern is worth recording, because it is the opposite of what happened: the ordering was changed *to* persist-first precisely so that a process killed mid-dispatch re-polls rather than dispatching a second time. The reaper this series is building toward is what makes that window reachable in the first place.

totollygeek:
Post-rebase confirmation for the two findings above.

The security review re-ran against the rebased head and reports **0 findings** (comment updated `08:15:43Z`). The generic review's comment is still the pre-rebase body, last updated `08:06:46Z` — before the rebase landed at `08:15:15Z` — so `2p0b8f9zctul8` is stale rather than reproduced.

Independently checkable: the file it names is no longer in this PR's diff at all. The changed files are `docs/authoring.md`, `llms-full.txt`, `packages/core/README.md`, `packages/core/src/scheduler.ts`, `packages/core/src/target.ts`, and `tests/integration/always_settled_test.ts`. The branch now contains #315, so the persist-before-dispatch order is what is on it.

Not adding either id to the suppress list in `zuke.ts`: they were correct readings of a stale base, and suppressing them would hide the same signal next time it is real.

totollygeek:
Resolved. With #317 merged, this branch's base retargeted to `master` but it still carried the pre-squash outcomes commits, so the diff showed work that was already upstream. Rebased onto `master` — the two outcomes commits dropped as already-applied — leaving one commit and a six-file diff.

Both reviewers re-ran against that diff and now report **0 findings**. `2p0b8f9zctul8` is gone from the generic review's body rather than dismissed, which is the outcome to want: the finding was a correct reading of a stale base, so it should still fire if the ordering ever genuinely regresses.

Gate green locally on the rebased head, 18 of 18 targets.

PR #322, opened 2026-08-10
feat(core): give a run a lease, so slow and dead stop looking alike

totollygeek:
## Why

A run record cannot tell a working process from a dead one. A process mid-step and a process that was `SIGKILL`ed both leave a record that says `running` and then stops changing. Anything that wants to take over abandoned work has to guess, and guessing either way is wrong: too eager and two processes drive one run, too cautious and nothing ever recovers it.

This is R2 of the durable merge-gate work, and it is the precondition for the reaper — which cannot exist until "abandoned" is a thing zuke can observe rather than infer.

## What it does

A run with a state store holds a lease on itself, `zuke-run-<run-id>`, renewed by a heartbeat at half its 60-second TTL and released when the run settles. A lapsed claim means the holder is gone. Nothing polls for expiry; the store evaluates it when somebody next tries to acquire.

The cancellation lock becomes a thin naming of the same thing, signature and key unchanged — its mechanics were never about cancellation.

## The three decisions worth reviewing

**The lease is taken before the record says `running`** — and on a resume, before the record leaves `suspended`. So a `running` record always has a live holder. Acquiring afterwards would leave a window in which a perfectly healthy run looked abandoned, which is precisely the question the lease exists to answer. That is why acquisition lives in `openRunState` ahead of the writer (opening the writer persists `running`) and in `resumeRun` ahead of the compare-and-swap, threaded through `ResumeState`.

**Loss is an `AbortSignal`, not a callback.** A resumer takes the lease before the run it will drive exists, so there is nothing to hand a callback to yet — and a callback would drop the already-lost-before-you-subscribed case. A signal can be read late and still be true.

**A refused renewal is loss; a failed one is not.** Both stores return `false` only when the claim has demonstrably changed hands; they *throw* for a filesystem mutex they could not take in time, an HTTP 503, or a DNS blip. None of those say anything about who holds the lease, so they are retried on the next tick. Treating them as loss would abort a healthy build because the state service had a bad second.

A resume also now refuses a run whose previous holder has not let go. Two processes working one run is worse than a resume that waits, and it is the same answer the existing exactly-one-resumer race already gives, from the other side.

## A caveat, documented rather than hidden

The heartbeat is a timer, so it does not fire while the event loop is blocked in synchronous work. A run that blocks for longer than the TTL can have its lease lapse and be taken over. Losing the claim then stops it, so the outcome is a stopped run rather than two writers — but it is a real limitation and `docs/locks.md` says so.

## Tests

- **Unit, 8:** the key and TTL, acquire and release, refusal when a live holder has it, an explicit refusal counting as loss and stopping the heartbeat, a throwing renewal *not* counting as loss and surviving a blip that passes, release stopping renewal, and the cancel lock delegating without changing its key.
- **Integration, 2:** a run holding its lease while a target runs and having released it once settled, and a resume refusing a run whose holder is still live.
- **E2E:** the existing effect re-drive test now proves both halves of this. A resume **refuses** the killed process's run while its lease is still live, and succeeds once the lease lapses — which also makes explicit what the reaper will have to establish: a lapsed lease *and* a run to move.

Full gate green locally, 18 of 18 targets, plus the e2e suite.

### DIFF
--- a/docs/locks.md
+++ b/docs/locks.md
@@ -110,5 +110,10 @@ holder is gone and the run can be taken over.
   not fire while the event loop is blocked in synchronous work. A run that blocks for longer than the TTL can therefore have
   its lease lapse and be taken over; losing the claim then stops it, so the
   outcome is a stopped run rather than two writers.
+- **Whoever takes the claim gives it back.** A resume releases the lease it took
+  on every path out, including one where the run fails — a claim held by nobody
+  would make a run that has demonstrably stopped look like one still being
+  worked on. A run that took its own lease releases it when it settles; if that
+  process breaks first, the claim lapses at its TTL instead.
 - **A crashed holder's claim lapses at the TTL.** Nothing polls for it: expiry is
   evaluated by the store the next time somebody tries to acquire.
--- a/docs/locks.md
+++ b/docs/locks.md
@@ -76,3 +76,39 @@ Acquisition is atomic. Two runs racing for the same free key → **exactly one**
 acquires; the other gets the conflict. An expired lock is taken over atomically.
 The filesystem backend is single-host (an `O_EXCL` marker serialises
 acquisition); the HTTP backend uses the same optimistic model across hosts.
+
+## The run's own lease
+
+Separately from any lock a target declares, a run with a
+[state store](./state.md) holds a **lease** on itself for as long as it is
+running: `zuke-run-<run-id>`, taken before the record is written `running` and
+released when the run settles. It is renewed by a background heartbeat at half
+its 60-second TTL.
+
+Its whole job is to make "slow" and "dead" different things. A run record cannot
+tell them apart on its own — a process mid-step and a process that was
+`SIGKILL`ed both leave a record that says `running` and stops changing. The lease
+answers it: a live holder keeps renewing, so a claim that has lapsed means the
+holder is gone and the run can be taken over.
+
+- **Ordering matters.** The lease is taken *before* the record says `running`,
+  and on a resume before the record leaves `suspended` — so a `running` record
+  always has a live holder. Acquiring afterwards would leave a window in which a
+  perfectly healthy run looked abandoned.
+- **A resume refuses a run whose holder has not let go.** Two processes working
+  one run is worse than a delayed resume, so the claim is checked before the
+  compare-and-swap and a resumer that cannot take it stops.
+- **Losing it stops the run.** If a renewal is refused — the claim is
+  demonstrably somebody else's now — the run aborts rather than carrying on
+  beside whoever took it over.
+- **A refused renewal is loss; a failed one is not.** A store answers "no" when
+  the claim has changed hands, but it *throws* for a filesystem mutex it could
+  not take in time, or an HTTP 503, or a DNS blip. None of those say who holds
+  the lease, so they are retried on the next tick rather than aborting a healthy
+  build over a bad second.
+- **The heartbeat never keeps a process alive** — and, like any timer, it does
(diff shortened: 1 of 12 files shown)


totollygeek:
### Review finding

**`10l3omurh2s07`** — "Potential stale run lease if resume fails after acquiring it" at `packages/core/src/resume.ts:176` — **confirmed and fixed** in `cb653f0`.

The compare-and-swap path was already covered: losing that race releases the lease before rethrowing. What was not covered is everything after it. The lease was acquired by `resumeRun` and released by `execute`, which is the wrong way round in two concrete ways:

- **`execute` can exit without reaching the release.** Its release sits after the settlement block, so a throw anywhere earlier — `services.stopAll`, `cache.save`, a settlement write — left the resumer's claim held by nobody until the TTL lapsed.
- **Its early return happens before the lease is even read.** `if (!opened.ok)` returns above the line that reads `options.resume.lease`, so that path could never have released it. Not reachable today, because a resume always resolves a store and that refusal needs `stateStore === undefined` — but reachable by construction the moment anyone adds another early return above it.

Why it matters more than "a lock lingers for a minute": a claim held by nobody makes a run that has demonstrably stopped look like one still being worked on, which is the single distinction this lease exists to draw. A sweep would skip it for a minute, having read the one signal that was supposed to tell it otherwise.

Fixed by making ownership explicit rather than by adding another release: **whoever takes the claim gives it back.** `resumeRun` releases in a `finally`, on every path out including one where the run throws; `execute` releases only a lease it acquired itself. That also improves the failing-run case — the claim now goes immediately rather than at the TTL, so an abandoned run is visible as abandoned at once.

A run that took its own lease and then broke mid-run still lets it lapse at the TTL, deliberately: there is no outer scope that outlives that call, and the record is left `running` by a process that really is broken, so a sweep finding it a minute later is the honest outcome. Stated in `docs/locks.md` alongside the rest of the lease's contract.

Regression test: `tests/integration/run_lease_test.ts` — "a resume gives the lease back even when the run fails" resumes into a target that throws, then asserts the next acquirer finds the lease free. Full gate green, 18 of 18, plus the e2e suite.

totollygeek:
### Review finding

**`1q11zmmr8qyd5`** — "Integration test removes the lease file directly instead of exercising normal expiry" at `tests/e2e/effect_redrive_e2e.ts:111` — **confirmed and fixed** in `fca944a`.

The criticism is exactly right, and sharper than "a test took a shortcut". A missing lock and an expired one are **different branches** in `FileSystemStateStore.acquireLock`:

```
const current = await this.#readLock(key);
if (current !== null && current.expiresAt > now) return { ok: false, holder: current.holder };
// …otherwise write a new lock — this is the takeover
```

Deleting the file made `current` null, so the resume took the ordinary fresh-acquire path. The branch that gives a lapse its meaning — a lock record that still exists, still names the dead holder, and is taken over because its expiry has passed — was never executed. The test asserted "re-drive after the lease lapses" while proving "re-drive when there is no lease", which is the weaker claim and not the one production depends on.

Fixed by moving only the clock. The lock record stays exactly as the killed process left it, holder and token intact, and only `expiresAt` is backdated, so the resume goes through the real takeover. Waiting out the 60-second TTL is the only other way to reach that branch, and it would test the same code a minute later.

**Mutation-checked**, since a test that claims a branch should be shown to fail without it. Making an expired lock still conflict:

`if (current !== null) return { ok: false, holder: current.holder };`

now fails this test. Under the previous version it passed, which is the finding restated as evidence.

Full gate green, 18 of 18, plus the e2e suite. Not suppressing the id — "a test manipulates state instead of exercising it" is worth hearing every time.

PR #323, opened 2026-08-10
feat(core): reap abandoned runs and give a run a deadline

totollygeek:
## Why

A run that ends cleanly settles itself, and a run that suspends is picked up by the resume sweep. A run whose process was **killed** does neither: it stays `running`, and a resume only ever acts on `suspended` runs. So anything that run recorded as owed waited forever — which made the durable half of a durable effect a promise nothing kept.

This is the last piece of the durable merge-gate series (#315, #317, #318, #319, #320, #322).

## What it does

`zuke resume --check` now examines this build's `running` runs before sweeping the suspended ones. The first question is always whether anyone is still there — the run's lease answers it, since a live process keeps renewing. A run with no holder goes back to `suspended` with a `reap` audit event, and the same sweep resumes it, so recovery takes one pass rather than an extra interval spent on nothing but a state change.

A build can now declare a wall-clock budget for a whole run. A run with no holder that has passed that budget is settled **`failed`**, compensations and all — it did not stop because anyone asked, it ran out of time, and anything waiting needs an answer. The budget covers *running*: time parked at a gate is given back on resume, so a 45-minute deadline behind a 72-hour approval gate still has its 45 minutes when the approval arrives.

The settlement machinery is now shared. An operator's cancel is one terminal, a passed deadline is the other, and the walk between them is identical — because an abandoned run's side effects need unwinding just as carefully as a cancelled one's. Which terminal a settlement intends is recorded on the run, because the process that finishes it may not be the one that began it.

## `zuke cancel` is unchanged, and that was tested first

The requester's promote plane runs these paths from a pinned image, so a silent change stays dormant until a deployment moves. I wrote a contract test suite **before** touching anything: the returned result field by field, the terminal status, the audit events, and the wording of every line it prints.

That discipline paid twice. It caught my own wrong assumption immediately — I expected the cancel summary event to carry the run id in its args, and it carries an empty object — and it caught me casually rewording two reporter lines, which I then restored verbatim for the cancel path.

## The second commit is the adversarial pass, and it found four real defects

Two ended in **a run whose work had been unwound reporting that it succeeded** — the outcome this series exists to prevent. Two reviewers found the first one independently.

- **The writer only noticed an external settlement for `cancelling` and `cancelled`.** A sweep leaves a run `failed`, so the process still working on it sailed past the check and re-applied its own terminal. Worse, that check only fires on a *conflicting* write: once a writer came back into step with the settled record, its next write landed unchallenged. The observation is now latched, and a process that discovers it reports the run as failed rather than claiming an outcome that is not its to give.
- **The deadline settled runs that were demonstrably alive**, because it was checked before asking whether the run still had a process — running a run's compensations beside the work they were undoing. The lease question now comes first. That costs nothing the deadline was for: a hung process stops renewing its lease, and a run killed repeatedly has no holder at all.
- **The deadline charged time spent parked at a gate**, so the docs' own worked example produced the outcome the docs excluded. Now pushed forward on resume.
- **The sweep mutated other builds' runs.** A store is commonly shared and a listing has no build filter, so a settlement found none of the run's targets, stamped it terminal with its side effects never unwound, and left nothing to retry it. The suspended sweep was safe here only because a resume *refuses* an unmatched build; a reap mutates, so it now checks.

Plus two smaller ones: resolving the build's budget on the resume path computed a discarded value, and a malformed one stranded the run to be reaped and stranded again forever; and a single-run sweep never recovered a run left `cancelling`, which is exactly the sweep shape with nothing else to fall back on.

## Tests

- **Unit:** 14 reaper — live holder skipped, dead holder released, lease not retained, deadline settled with compensations, malformed deadline refused rather than guessed, a run settled between the lease and the write left alone, foreign builds untouched, one bad run not stranding the sweep. Plus 5 cancel-contract and 2 writer.
- **Integration:** 5, including the whole loop — nothing but a sweep, and a killed run's owed effect is performed.
- **E2E:** two real sweeper processes racing one abandoned run; exactly one reap event and exactly one extra drive. Run repeatedly for flakiness.

**Four fixes are mutation-checked** — reverting any one fails its test. That process caught a fifth problem: my first parked-time test passed *without* the fix, so it was rewritten to assert the property directly.

Full gate green locally, 18 of 18, plus the e2e suite.

### DIFF
--- a/docs/orchestration.md
+++ b/docs/orchestration.md
@@ -248,11 +248,19 @@ suspended ones:
   build's run would find none of its targets and settle it with its
   compensations silently skipped.
 
-  Before *settling* a run — which is irreversible — the graph has to agree as
-  well, because a class name is not an identity and `Ci` is a name half an
-  organisation's repos will use. Handing a run back to `suspended` asks only the
-  looser question, since that is reversible and a resume refuses a graph it does
-  not recognise with an error naming the drift.
+  Before *settling* a run — which is irreversible, and runs its compensations —
+  the graph has to agree as well, because a class name is not an identity and
+  `Ci` is a name half an organisation's repos will use. Handing a run back to
+  `suspended` asks only the looser question, since that is reversible and a
+  resume refuses a graph it does not recognise with an error naming the drift.
+
+  What no check here can see: two builds that share a class name, a root-target
+  name **and** a graph shape are indistinguishable in a run record, even though
+  their target *bodies* may do entirely different things. If several repos share
+  one state service and any of them might collide that way, give each its own
+  store — a separate `ZUKE_STATE_DIR`, or a distinct prefix or instance behind
+  `ZUKE_STATE_URL`. That is the only way to make the question unambiguous, and it
+  costs nothing when the runs were never meant to be pooled.
 
 A run left `cancelling` by a settlement whose own process died is finished too,
 in whichever terminal that settlement was heading for — recorded on the run,
--- a/docs/orchestration.md
+++ b/docs/orchestration.md
@@ -229,17 +229,23 @@ wait forever.
 `zuke resume --check` now looks at `running` runs first, before it sweeps the
 suspended ones:
 
-- **Is anyone still there?** The run's [lease](./locks.md#the-runs-own-lease)
-  answers it. A live process keeps renewing, so a lease that can be acquired
-  means its holder is gone; one that cannot means the run is merely slow, and
-  slow is left alone.
+- **Is anyone still there?** — asked first, always. The run's
+  [lease](./locks.md#the-runs-own-lease) answers it: a live process keeps
+  renewing, so a lease that can be acquired means its holder is gone; one that
+  cannot means the run is merely slow, and slow is left alone. A live run is
+  never settled or moved, deadline or no deadline, because doing so would run its
+  compensations beside the work they undo.
 - **Nobody there** → the run goes back to `suspended`, with a `reap` event on
   its audit trail saying why. The same sweep then resumes it, so an abandoned
   run is recovered in one pass rather than waiting another interval for nothing
   but a state change.
-- **Past its `deadline()`** → the run is settled **`failed`**, compensations and
-  all. It did not stop because anyone asked; it ran out of time, and anything
-  waiting on it needs an answer rather than silence.
+- **Nobody there, and past its `deadline()`** → the run is settled **`failed`**,
+  compensations and all. It did not stop because anyone asked; it ran out of
+  time, and anything waiting on it needs an answer rather than silence.
+- **Only this build's runs.** A state store is commonly shared, and a listing has
+  no build filter, so the sweep skips runs belonging to another build. Acting on
+  one would find none of its targets, settle it with its compensations silently
+  skipped, and leave nothing to retry it.
 
 A run left `cancelling` by a settlement whose own process died is finished too,
 in whichever terminal that settlement was heading for — recorded on the run,
@@ -257,11 +263,18 @@ class Ci extends Build {
 }
 ```
 
-A run parked at a `.waitsFor(...)` gate is not spending it — the budget there is
-the gate's own `.timeout()`, which already exists and already fires. Only the
-sweep over runs still marked `running` consults the deadline, so a build with a
-72-hour approval gate and a 45-minute deadline is not killed the moment it
-suspends.
+A run parked at a `.waitsFor(...)` gate is not spending it: the deadline is
+pushed forward on resume by however long the run was parked, so a build with a
+72-hour approval gate and a 45-minute deadline still has its 45 minutes of
+running time when the approval finally arrives. The gate's own `.timeout()` is
+what bounds the waiting.
+
+A live run is never settled for time, either — the sweep asks whether anyone is
+still working on the run before it looks at the deadline, and leaves it alone if
+someone is. That costs nothing the deadline was for: a process that hangs stops
+renewing its lease, and a run killed and abandoned repeatedly has no holder at
+all. What it rules out is settling a run underneath a process that is still
+working on it.
 
 What it is really for is the run that stops making progress without failing: a
 process that hangs, or one killed so hard that its work is picked up and
--- a/docs/orchestration.md
+++ b/docs/orchestration.md
@@ -218,6 +218,57 @@ its recorded `onTimeout` disposition:
 Timeouts are enforced lazily on any `zuke resume`/`resume --check`, so a single
 cron that sweeps suspended runs also enforces every deadline.
 
+### Abandoned runs — reaping
+
+A run that ends cleanly settles itself, and a run that suspends is picked up by
+the sweep above. A run whose process was **killed** does neither: it stays
+`running`, and a resume only ever acts on `suspended` runs. Anything that run
+recorded as owed — a [crash-durable effect](./run-context.md) above all — would
+wait forever.
+
+`zuke resume --check` now looks at `running` runs first, before it sweeps the
+suspended ones:
+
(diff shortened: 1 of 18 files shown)


totollygeek:
### Review findings

**`x80ycu95oi7g`** (high) — "Build ownership check relies on constructor name, allowing cross-build run mutation when names collide" — **confirmed and fixed** in `775e355`.

Correct, and it undercut the very check it is about. `Ci` is a name half the repos in an organisation will use, and the setup this feature is for is a shared state service holding every repo's runs — so a name-only check would let one repo's cron settle another's, which is exactly what the check was added to prevent.

The record carries nothing more specific than the class name, so the fix is a second condition rather than a better identity: the record's root target must exist in this build too. That is what makes a collision harmless instead of merely unlikely — a settlement resolves its compensations *by name*, so a build holding the run's root target holds the targets that run recorded, and the destructive outcome (settling a run with its compensations silently skipped) is closed.

The residual is stated where the check lives: two builds sharing a class name **and** a root-target name are still indistinguishable, and closing that needs an identity in the record rather than a better guess in the sweep. Regression test added and mutation-checked — dropping the root-target condition fails it.

**`lhpyw96zwxp8`** (medium) — "Stranded-settlement recovery ignores the recorded intended terminal and always finalizes as cancelled" — **refuted**, and now pinned by a test so it stays that way.

The literal argument is `terminal: "cancelled"`, which is what the finding read, but that value is only a default and never reaches the record. `recoverStranded` only ever calls for runs already in `cancelling`, so `transitionToCancelling` returns `"recover"`, and that branch re-reads the record and uses its `intendedTerminal` — narrowing to `failed` when that is what was recorded, and to `cancelled` when the field is absent (an ordinary cancel, or a record written before the field existed).

Verified empirically rather than by reading: a stranded run carrying `intendedTerminal: "failed"` settles **`failed`**. The comment at the call site said this was a default, which was evidently not enough, so there is now a test — *"a stranded run settles into the terminal its settlement intended"* — that asserts it.

Neither id is going in the suppress list. The first was a genuine defect; the second is a reasonable misreading of a call site that deserved the test it now has.

totollygeek:
### Second pass on the same two findings

The re-run after `775e355` reported both again, reworded to name what that fix had *left behind* rather than what it fixed — "when build and root-target names collide" and "likewise vulnerable". That is the residual I had documented in the code comment, and reading it back as a live finding is the right call: a settled run with its compensations skipped is, for a deploy, a live change standing behind a record that says the run is over. Documented is not the same as closed.

Closed now in `8182f3c`, by splitting the question in two rather than tightening one check:

- **Handing a run back to `suspended`** asks the looser question — build name plus root target. That step is reversible and self-correcting: a resume refuses a graph it does not recognise, with an error naming the drift and `--force-graph` as the documented override.
- **Settling a run** asks whether the recorded graph matches the one this build plans. Where it matches, the two builds are the same build for every purpose a settlement has, because compensations resolve by name and the names are identical. Where it does not, the sweep leaves the run for the build that recognises it.

The split is deliberate rather than uniform strictness. Requiring a graph match to *reap* would strand a run whose own build had legitimately changed since it started — it would stay `running` with nothing to move it, which is a worse failure than the one being fixed and is exactly the hole this PR exists to close.

Both new cases are tested and mutation-checked: a colliding build with a different graph neither settles a past-deadline run nor finishes a stranded one, and making the graph comparison unconditionally true fails both.

Full gate green, 18 of 18, plus the e2e suite.

totollygeek:
### Third pass: one correction of mine, two refutations

First, a correction to my own previous reply. I wrote that where the graph agrees, "the two builds are the same build for every purpose a settlement has". That is wrong. The target *names* match; the bodies need not. One repo's `rollback` and another's are the same name and different code. Corrected in `eee205e`, in the comment and the docs.

**`3j3gkkbakvlin`** (high) — "Reaper can settle another build's stranded run when build identity collides" — **the collision is real and unclosable by a record comparison; the stated harm is not present on this path.**

A stranded run is finalised through the recover branch, which settles **without running any compensations** — `finalizeCancelled(store, runId, actor, EMPTY_OUTCOME, now, intended)` — and into the terminal the run's *own record* names via `intendedTerminal`. So a colliding build settles a stranded run exactly as the owning build would have: same terminal, no compensations either way. The only difference is the `actor` on the audit event.

The genuine residual is the other path, which this finding does not name: a **past-deadline** settlement does run compensations, and there "same names, different bodies" would run the wrong code. That is why `canSettle` exists, and it is also why no third heuristic is coming — a record cannot distinguish two builds that agree on class name, root target and graph shape. What closes it is a store per build: a separate `ZUKE_STATE_DIR`, or a distinct prefix or instance behind `ZUKE_STATE_URL`. That is now documented in `docs/orchestration.md` as the answer, rather than left as a limitation the reader has to infer.

**`3t3vcrkcj0xyh`** (medium) — "makes terminal status depend on local wall clock without validating stored timestamp format" — **the format is validated; the clock dependency is bounded.**

The validation is there:

`return Number.isFinite(deadline) && Date.parse(now) > deadline;`

An unparseable `deadlineAt` yields `NaN`, `Number.isFinite` is false, and the run is not treated as expired — refusing to guess rather than settling a healthy run over a malformed timestamp. There is a test for exactly this, *"a malformed deadline is not treated as a passed one"*, and it passes.

The wall-clock half is true and inherent: any wall-clock deadline compares a timestamp one process wrote against another's clock. What bounds it here is the ordering — the lease is probed first, so a sweeper with a skewed clock can only settle a run that is *already ownerless*, and an ownerless run needs an answer regardless. A skewed sweeper cannot settle a run that something is still working on.

Neither id is suppressed. The first names a real limitation whose answer is configuration, and I would rather it keep being raised than be silently dismissed.

totollygeek:
### Fourth pass: same residual, third wording — stopping here

**`j7qwhdluoo9y`** and **`31ko47nvm9omo`** are the residual documented in `eee205e`, restated. Tracking the lineage, because the ids change each run and only these replies persist:

| pass | ids | what it named | outcome |
| --- | --- | --- | --- |
| 1 | `x80ycu95oi7g`, `lhpyw96zwxp8` | name-only ownership check | **fixed** — root target required (`775e355`) |
| 2 | `x80ycu95oi7g`, `lhpyw96zwxp8` | the residual that fix left | **fixed** — graph must agree before settling (`8182f3c`) |
| 3 | `3j3gkkbakvlin`, `3t3vcrkcj0xyh` | collision + clock/format | refuted with evidence; my own overclaim corrected (`eee205e`) |
| 4 | `j7qwhdluoo9y`, `31ko47nvm9omo` | the same collision, reworded | this comment |

Three real defects came out of passes 1 and 2 and are fixed and mutation-checked. What is left is one limitation, stated plainly in both the code and `docs/orchestration.md`: **two builds that agree on class name, root-target name and graph shape are indistinguishable in a run record.** No comparison of a record closes that, because the thing that differs — the target bodies — is not in the record. A fourth heuristic would only move the boundary again.

Two answers exist, and which one applies is a decision for whoever operates the store rather than something to guess at here:

1. **A store per build**, which is configuration and costs nothing when the runs were never meant to be pooled: a separate `ZUKE_STATE_DIR`, or a distinct prefix or instance behind `ZUKE_STATE_URL`. Documented as the answer.
2. **A declared build identity in the record** — an override defaulting to the class name, compared instead of it. That closes the question properly rather than bounding it, and it is a public API addition that deserves its own PR with its own tests and docs, not a late amendment to this one.

Worth keeping in view while judging severity: the two paths differ. A **stranded** run is finalised without running any compensations and into the terminal its own record names, so a colliding build settles it exactly as the owning one would — the only observable difference is the `actor` on the audit event. A **past-deadline** run does run compensations, which is the case worth caring about and the reason `canSettle` checks the graph at all.

Not suppressing either id. This should keep being raised until the store is scoped or the identity lands.

PR #324, opened 2026-08-10
feat(core): enforce operator token over whole run plan

totollygeek:
## What & why

Protection of sensitive targets (via `--protect`) now enforces the operator token requirement across a run's entire dependency closure, not just the root target. This closes a critical authorization bypass: a protected target like `deploy` can no longer be invoked without a token by calling an unprotected dependent like `release` that depends on it.

The changes also refine the allow-list semantics:
- **Allow-list gates invocation, not execution**: invoking a target inherently runs its dependencies, so allow-listing a root allows everything that root does.
- **Read-tool visibility follows the allow-list closure**: when an allow-list is configured, the read tools (`list_targets`, `graph`, `describe_build`) narrow to the matched targets' dependency closure, making hidden targets genuinely unreachable rather than merely undisplayed.
- **Audit trail is operator-only**: the `mcp-audit` run record is no longer readable over MCP (preventing agents from confirming their own denials), but remains accessible to the operator via the CLI.

The operator token check now:
1. Resolves the full execution plan for any `run:` tool call
2. Checks whether that plan touches any protected target (by name or in the closure)
3. Advertises `operatorToken` as required in the tool schema if the plan is protected
4. Denies the call with `missing_operator_token` or `invalid_operator_token` if needed
5. Fails closed on unresolvable plans (unknown targets, graph errors)

Similar protection is extended to `signal_run` and `cancel_run` over suspended runs, and to the registry server's `run:<buildId>:<target>` tools.

## Related issues

Closes authorization bypass where protected targets were reachable through unprotected dependents.

## Checklist

- [x] The PR title is a Conventional Commit (`feat(core): ...`).
- [x] Tests added: `mcp_authz_closure_test.ts` (390 lines, 11 test cases covering the bypass, schema advertisement, signal/cancel gating, allow-list closure, read-tool visibility, and audit trail access control) and integration test in `mcp_authz_test.ts`.
- [x] Docs updated: `docs/mcp.md` clarified on protection scope, allow-list semantics, and audit trail access.
- [x] No `any`, `as` casts, or `!` assertions in new `src/` code.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
       "name": "zuke",
       "source": "./plugins/zuke",
       "description": "Agent skills for Zuke — scaffold Zuke into a project (zuke-setup) and write or edit a zuke.ts build (zuke-write-build).",
-      "version": "0.2.0",
+      "version": "0.3.0",
       "author": {
         "name": "Zuke",
         "email": "contact@zuke.build",
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -394,6 +394,28 @@ plugins/zuke/             # Claude Code plugin wrapping the skills
   the website's package grid is generated from it by `syncWebsite`.
 - **Update docs with code.** If behaviour changes, update `README.md`, JSDoc,
   and the spec/acceptance criteria in the same PR.
+- **The agent skills are docs too — and they ship to a marketplace.** `skills/`
+  is the source of truth for `zuke-write-build` and `zuke-setup`, and
+  `plugins/zuke/` is the Claude Code plugin that publishes them. Any change to
+  the authoring surface or to a documented guarantee — a new `target()` method,
+  a new `Build` override, changed CLI or authorization semantics — must be
+  reflected in `skills/zuke-write-build/SKILL.md` and
+  `skills/zuke-write-build/references/cheatsheet.md` **in the same PR**. The
+  cheatsheet is one of the two canonical answers to "does a wrapper exist?", so
+  a new package belongs in its catalogue table as well. Then, in order:
+  1. Run `./zuke pluginSync` to regenerate `plugins/zuke/skills/`. Never hand-edit
+     the copies — `pluginSyncCheck` fails on drift, and it is the only part of
+     this that is automated.
+  2. **Bump the plugin version by hand, in both manifests**:
+     `plugins/zuke/.claude-plugin/plugin.json` and the entry in
+     `.claude-plugin/marketplace.json`. They must agree —
+     `tests/plugin_manifest_test.ts` enforces that much, but nothing can tell
+     you that you *forgot* to bump, because that needs the base branch to
+     compare against. Clients use the version to decide whether an installed
+     plugin is stale, so skills edited without a bump simply never reach agents
+     that already hold the old copy. release-please does **not** manage
+     `plugins/` — it is not a workspace package and has no `deno.json`.
+     Additive skill content is a minor bump; a correction is a patch.
 - **Always read the reviewer comments on every PR.** This repo runs AI reviewers
   (`@zuke/ai`) that post their assessments as PR comments (and human reviewers
   do too). Before considering a PR done — and again after each push — fetch and
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -294,7 +294,8 @@ subset that can drift from it. `zuke.ts`'s `ci` target depends on: `format`
 (`deno fmt --check`), `lint` (`deno lint`), `spell` (cspell), `coverage`
 (type-check, then the test suite with the 95% coverage gate), `coverageUpload`
 (skips locally without a `CODECOV_TOKEN`), `apiDocsCheck`, `docLint`,
-`snippetsCheck`, `hclSyncCheck`, `pluginSyncCheck`, `prBodyLint`, and
+`snippetsCheck`, `hclSyncCheck`, `pluginSyncCheck`, `pluginVersionCheck`,
+`prBodyLint`, `actionPinCheck`, `security`, and
 `lockCheck`. Read `zuke.ts`'s `ci` target for the current, authoritative list —
 this is a snapshot, not a second source of truth.
 
@@ -403,19 +404,24 @@ plugins/zuke/             # Claude Code plugin wrapping the skills
   `skills/zuke-write-build/references/cheatsheet.md` **in the same PR**. The
   cheatsheet is one of the two canonical answers to "does a wrapper exist?", so
   a new package belongs in its catalogue table as well. Then, in order:
-  1. Run `./zuke pluginSync` to regenerate `plugins/zuke/skills/`. Never hand-edit
-     the copies — `pluginSyncCheck` fails on drift, and it is the only part of
-     this that is automated.
+  1. Run `./zuke pluginSync` to regenerate `plugins/zuke/skills/`. Never
+     hand-edit the copies — `pluginSyncCheck` fails on drift.
   2. **Bump the plugin version by hand, in both manifests**:
      `plugins/zuke/.claude-plugin/plugin.json` and the entry in
-     `.claude-plugin/marketplace.json`. They must agree —
-     `tests/plugin_manifest_test.ts` enforces that much, but nothing can tell
-     you that you *forgot* to bump, because that needs the base branch to
-     compare against. Clients use the version to decide whether an installed
-     plugin is stale, so skills edited without a bump simply never reach agents
-     that already hold the old copy. release-please does **not** manage
-     `plugins/` — it is not a workspace package and has no `deno.json`.
-     Additive skill content is a minor bump; a correction is a patch.
+     `.claude-plugin/marketplace.json`. Clients use the version to decide
+     whether an installed plugin is stale, so skills edited without a bump
+     simply never reach agents that already hold the old copy. release-please
+     does **not** manage `plugins/` — it is not a workspace package and has no
+     `deno.json`. Additive skill content is a minor bump; a correction is a
+     patch.
+
+  Two gate targets hold this up, so a miss fails the build rather than shipping
+  quietly: `pluginVersionCheck` fails when a published skill changed against the
+  base branch and the version did not move, and `tests/plugin_manifest_test.ts`
+  fails when the two manifests disagree. `pluginVersionCheck` is the one part of
+  the gate that needs history — it compares against `origin/<PR base>`, or
+  `ZUKE_PLUGIN_BASE_REF` when you set one — and it reports itself *skipped*,
+  never passed, in a clone that has no base to compare against.
 - **Always read the reviewer comments on every PR.** This repo runs AI reviewers
   (`@zuke/ai`) that post their assessments as PR comments (and human reviewers
   do too). Before considering a PR done — and again after each push — fetch and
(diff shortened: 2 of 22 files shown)


totollygeek:
## AI review findings — one fixed, two refuted

All three findings from the security and generic reviews, each verified against the code rather than taken at face value.

### `n6n05sxeq803` — forbidden `as` assertion — **fixed** in 1200607

Correct, and it was the only non-comment `as` cast in all of `build/`. The check read `version` off a parsed JSON value through `(parsed as Record<string, unknown>)`, which asserts a shape the parser never established — exactly what guideline 1 rules out. Replaced with a `isRecord` type guard, so the `null` and array cases are now handled by the narrowing rather than by the assumption behind the cast. The new manifest test lost its cast the same way.

The reviewer reported this at two locations (`:60` and `:Infinity`); both refer to the one call site, now gone.

### `3ngcosrz8x1n` — audit redaction may leak non-string secrets — **refuted**

The change moves strictly in the opposite direction. Before:

```
if (typeof value !== "string") continue;   // a non-string secret was NEVER seeded
```

After, every secret-keyed value is seeded via its audit string form. So a secret supplied as a JSON number previously had its echo into another argument masked *not at all*, and now is masked whenever the string forms match. The matching algorithm — literal substring replace in `Redactor` — is untouched by this patch, so "if matching is imperfect" describes the pre-existing model, not a regression introduced here.

The secret's own slot is also unaffected: the `secret_` branch in the output loop still writes `[redacted]` unconditionally, before any matching is consulted. The only other edit to that loop replaces an inline ternary with the shared `auditText` helper, which is character-for-character the same expression.

Worst case for the new seeding is *over*-masking (a secret of `1` masking a `1` elsewhere in the trail), which fails safe and is why `Redactor` sorts longest-first.

### `3akcx5hclp4n6` — 1 MiB body cap enables denial of service — **refuted as framed**

The cap is the DoS mitigation, not a DoS: without it a single unauthenticated POST to the loopback bridge is buffered whole via `JSON.parse(await request.text())`. The new reader streams and abandons the body at the limit, so an oversized request is never buffered just to be rejected.

On the residual point — that a hard limit could reject legitimate traffic — 1 MiB is generous for a JSON-RPC control message whose payload is a tool name plus scalar arguments, and a parameter large enough to breach it would be better passed by reference than inlined into an MCP call. Refuting rather than fixing, because making it configurable would add a CLI flag, a serve option and docs for a case that has not appeared. Happy to add that knob if a real build hits the ceiling.

---

All 10 checks were green on 9bfecd2; 1200607 re-runs them. Full gate green locally (2,556 tests, 95% coverage gate, spell/lint/format/docLint/snippets/pluginSync/pluginVersion/lock). `security` fails only in my sandbox, where zizmor's `known-vulnerable-actions` audit gets HTTP 403 from the GitHub Advisories API — it passes in CI.

---

totollygeek:
## Second review pass — both findings fixed in d0ba437

Score moved 5/10 → 2/10, medium → low, and the two new findings replaced the earlier set. Recording the previous ids for the trail, since the reviewer upserts one comment and the old bodies are gone: `n6n05sxeq803` (fixed in 1200607), `3ngcosrz8x1n` and `3akcx5hclp4n6` (refuted — reasoning in the [previous comment](https://github.com/zuke-build/zuke/pull/324#issuecomment-5244637657)).

### `ngwd7il7opji` — audit trail hidden without an authorization boundary — **fixed, and it was a real bypass**

This one was worth the review. The refusal compared the requested id to the trail's exactly:

```
if (runId === AUDIT_RUN_ID)
```

`FileSystemStateStore` maps an id straight to `<id>.json`, and `assertSafeId` permits `[A-Za-z0-9._-]` — capitals included. On a case-insensitive volume `MCP-AUDIT.json` *is* `mcp-audit.json`, so `show_run {"runId": "MCP-AUDIT"}` walked past the guard and returned the entire trail: every actor, argument, and denial reason. The control I added last push held on Linux and failed open on **macOS and Windows** — two of the three platforms in the test matrix.

Now compared case-insensitively, in both `show_run` and the `list_runs` filter. The regression test asserts three casings and was confirmed to fail against the exact-match comparison it replaces.

I checked the neighbouring equivalences too: the id charset is ASCII-only so Unicode normalisation on APFS is not a vector, and a trailing dot produces `mcp-audit..json`, a different file.

### `2ead97l1f3w4m` — unresolved-plan denial reveals target existence — **fixed**

Thinner than it reads — reaching it via `run:` requires a target that is already known *and* allow-listed, and via the run-state tools the caller supplies a run id, not a name, and `list_runs` already shows them the `rootTarget`. So the practical leak is small.

Acting on it anyway, because the principle is one this module states explicitly and the fix is three lines: a caller who can tell `not_allowed` from `unresolved_plan` can distinguish a target outside the allow-list from one no longer in the build, and that is the existence signal the opaque `Unknown tool` answer exists to withhold. Clients now get the collapsed reason; the precise one still reaches the audit trail, which is operator-only and where it is actually useful for diagnosis.

The earlier test asserting `unresolved_plan` on the wire is superseded by one asserting it is withheld.

---

Full gate green locally: 2,557 tests, 95% coverage gate, format/lint/spell/docLint/snippets/pluginSync/pluginVersion/lock. `security` fails only in my sandbox, where zizmor's `known-vulnerable-actions` audit gets HTTP 403 from the GitHub Advisories API; it passes in CI.

---

totollygeek:
## Third review pass — one fixed in cd4de76, two refuted

Prior rounds, recorded because the reviewer upserts one comment and the earlier bodies are gone: `n6n05sxeq803` fixed (1200607); `3ngcosrz8x1n`, `3akcx5hclp4n6` refuted; `ngwd7il7opji`, `2ead97l1f3w4m` fixed (d0ba437).

A note on the score, which went 2/10 → 7/10 "high" across a push that only *tightened* the audit guard and collapsed a denial reason. Nothing in d0ba437 widened anything, and two of these three findings describe properties that predate this branch. Treating the findings on their merits rather than the number.

### `1tbq9mda88ha7` — audit exposure via serialized argument text — **fixed, partially valid**

The general claim (a substring redactor can miss a transformed secret) is inherent to that design, predates this branch, and is unchanged by it. But following it into the code found a real gap next to it.

`RegistryMcpServer.auditArgs` elides the value of any key the descriptor does not declare, with the rationale written down: a value mistakenly supplied under a secret parameter's name reads as an unknown key and must never be written verbatim. `McpServer.#auditArgs` did **not** do this. So a caller who typed `tokens` when the build declares `token` had their value recorded in full — it matches no declared secret, so nothing masks it.

That was an inconsistency, not a considered difference. Undeclared keys now keep their name and lose their value, matching the sibling server. Declared parameters are still recorded (naming the repos and slots a deploy touched is the point of the trail) and `dryRun`/`confirm` stay readable. Regression test added.

Worth noting the trail is also no longer readable over MCP as of this PR, so the residual exposure is an operator reading their own audit log.

### `bobg4n3m7es` — prompt injection via tool guidance — **refuted**

`docs/mcp.md` is a documentation file. It is not served to a model, and nothing in it reaches a tool description.

The strings a client actually receives are the static schema text this PR added and `target.description_` from the build file — authored by whoever wrote the build, who already controls execution, so it is not a trust boundary. There is no path by which caller-supplied input reaches tool guidance; the diff adds no interpolation into any description.

### `2h94ggphpii5d` — body parsed before schema validation — **refuted**

Parsing precedes dispatch by construction: JSON-RPC has to be parsed before the method is known. What matters is that it is bounded and guarded, and it is on both counts.

Ordering in `serveMcpHttp` is method check → Origin guard → bearer token → `readBounded` → `JSON.parse`. So with a token configured an unauthenticated caller never reaches the parser at all, and the body is capped at 1 MiB before parsing, read incrementally and abandoned at the cap rather than buffered whole. A nesting bomb raises `RangeError`, which the surrounding `try` already converts to a 400 parse error rather than a crash.

---

Full gate green locally: 2,558 tests, 95% coverage gate, format/lint/spell/docLint/snippets/pluginSync/pluginVersion/lock. `security` fails only in my sandbox, where zizmor's `known-vulnerable-actions` audit gets HTTP 403 from the GitHub Advisories API; it passes in CI.

---

totollygeek:
## Fourth review pass — all three refuted, no code change

Two of these three flag the exact changes the previous round asked for, and the third describes code that no longer exists. Prior rounds for the record, since the reviewer upserts one comment: `n6n05sxeq803` fixed (1200607) · `3ngcosrz8x1n`, `3akcx5hclp4n6` refuted · `ngwd7il7opji`, `2ead97l1f3w4m` fixed (d0ba437) · `1tbq9mda88ha7` fixed (cd4de76) · `bobg4n3m7es`, `2h94ggphpii5d` refuted.

### `3au0im963viv6` — "blocked only for exact run IDs" — **already fixed, in d0ba437**

This was true, and it was this reviewer's own `ngwd7il7opji` from round two. It was fixed two pushes ago. Current code:

```
function isAuditRun(runId: string): boolean {
  return runId.toLowerCase() === AUDIT_RUN_ID.toLowerCase();
}
```

applied at both reachable paths — `show_run` (`runtools.ts:306`) and the `list_runs` filter (`runtools.ts:276`). The regression test asserts `MCP-AUDIT`, `Mcp-Audit` and `mcp-AUDIT` are all refused, and was confirmed to fail against the exact comparison it replaced.

I re-walked every route to that record while checking this: `signal_run`/`cancel_run` on the trail resolve its root target `(audit)`, which is not in the build and is denied; the `resume_check` sweep lists only `suspended` runs and the trail is `running`. No exact-match path remains.

### `3tsa8lh0vntbo` — "collapse hides why a plan is denied but still logs the precise reason" — **refuted; this is the requested design**

That is a description of `2ead97l1f3w4m` from round two, which asked for exactly this and which I implemented last push. Round two's position was that distinguishing `unresolved_plan` from `not_allowed` on the wire leaks target existence; this round's position is that not distinguishing them is a problem. Both cannot hold.

Keeping the round-two behaviour, because it is the one with a security argument behind it: the caller gets a uniform denial, and the precise reason still reaches the audit trail — which is operator-only, so diagnosis is preserved exactly where it is useful and withheld exactly where it is a signal. The finding's own wording ("but still logs the precise reason") confirms the diagnostic path is intact.

### `1615u08bdkfvl` — "eliding undeclared keys hides data if a parameter is added later without updating the schema" — **refuted on the facts**

The scenario cannot occur. There is no schema to fall out of date with:

```
this.#params = discoverParameters(build);
```

The key set is derived from the build instance itself. Adding a parameter to the build *is* adding it to `#params` — there is no second place to update and therefore nothing to drift. A key can only be undeclared if the build genuinely does not declare it, which is precisely the case round three (`1tbq9mda88ha7`) asked to have elided, and which `RegistryMcpServer` has always elided.

---

**On the pattern.** Four rounds, scores 5 → 2 → 7 → 5, with rounds three and four each objecting to the fix the previous round requested. Four findings were real and are fixed; the rest describe pre-existing properties, contradict an earlier round, or misread the code. I am not reverting round two's and round three's fixes to satisfy round four — that would be a loop, and both fixes have concrete security arguments and regression tests behind them.

If the maintainers agree, `3tsa8lh0vntbo` and `1615u08bdkfvl` are good candidates for the `suppressions(...)` list in `zuke.ts`, since they will keep resurfacing against deliberate design. Happy to add them with these justifications — leaving that call to a human rather than dismissing the reviewer's own findings on my own initiative.

CI on cd4de76: 10 checks, all green or in progress, none failing.

---

PR #327, opened 2026-08-11
docs: correct stale claims across the docs, skills and root guides

totollygeek:
## What & why

An audit of every documentation surface against the actual code found roughly seventy statements that had drifted. They group into four kinds.

**Claims that were simply false.** `SECURITY.md` described a bootstrap that fetches and pipes an install script — both launchers download a pinned release archive, verify it against a baked-in checksum, refuse a moving version outright, and demand a matching hash for any override. The document understated the project's own strongest control. It also named release jobs and a CI job that do not exist, and asserted that every job hardens its runner when the one holding no token does not. `RELEASING.md` described the release as a single command, which is the local aggregate; CI runs four least-privilege jobs. Getting-started said the repo does not yet use its own published action, which it now does in all six workflows, and described the untrusted-checkout guard as a refusal for one event when it is an allow-list that also refuses a checkout with no target. A Node example pinned a version range that resolves to nothing.

**Guidance pointing at the wrong place.** Three files sent readers to a one-line pointer file for the coding standards and architecture notes. The PR template asked for a new package to be wired into five places and cited a document that documents none of them; the real count is seven and the real reference is `AGENTS.md`. `CODEOWNERS` still carried its own unresolved placeholder instruction, protected a directory that does not exist, and left the build helpers and the action manifest uncovered.

**Features documented nowhere a reader would look.** Six builder methods were absent from the authoring reference — including the cross-run lock and the suspend gate, which had no entry and no pointer anywhere on that page. Two of four lifecycle hooks were missing, as was every other override on the base class. The state-service contract omitted the three lock routes entirely, so a service built from it could store run records but support neither locks, nor the run lease, nor the reaper that depends on it. The command reference was missing a top-level flag, three server flags, one mutating tool, and the fact that the resume sweep now reaps abandoned runs before checking suspended ones.

**Skills.** Added the run deadline, the read-only marker, the telemetry package that was the only one missing from the catalogue, four flags and the completions command, the four combinations a fan-out rejects at materialisation time, the flag an effect body can read to detect its own re-drive, the second secret source, and the file the scaffolder writes that was never listed. Two rules that contradicted their own files — that a body is always required, and that the launcher and the task are equivalent — now state their exceptions.

**Two gates got sharper rather than just satisfied.** Six anchors pointed at a heading whose slug they got wrong; every relative link and anchor in the corpus was then verified with a checker, and all resolve. The snippet gate went from **4 checked examples to 30** — every fenced block that already type-checks standalone is now held to it, so a reader pasting one gets code that still compiles. Two root documents were excluded from the spell gate on the premise that they are generated, which neither is; both are checked now.

`llms.txt` and `llms-full.txt` needed no change: they are generated from JSDoc, this PR touches only prose, and the drift check confirms they are current.

## Related issues

Follows #324 and #326, closing out the documentation findings from the framework assessment.

## Checklist

- [x] `deno task ci` passes locally (lint, fmt, type-check, tests, spell). The security target fails only in this sandbox, where the vulnerable-actions audit cannot reach the GitHub Advisories API; it passes in CI.
- [x] Tests were added or updated; coverage stays at 95%+ — no source changed, and the full suite is green.
- [x] Docs updated in the same PR — this PR is the docs update.
- [x] Public API changes were regenerated with the apiDocs target; no public surface moved and the drift check passes.
- [ ] A new package was wired into all seven places — not applicable.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RYNxgcvdii7P8pPhD6U8q2)_

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
       "name": "zuke",
       "source": "./plugins/zuke",
       "description": "Agent skills for Zuke — scaffold Zuke into a project (zuke-setup) and write or edit a zuke.ts build (zuke-write-build).",
-      "version": "0.4.0",
+      "version": "0.5.0",
       "author": {
         "name": "Zuke",
         "email": "contact@zuke.build",
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,8 +1,6 @@
 # Code owners — require review on security-sensitive paths. Pair this with a
 # branch protection rule on `master` that requires CODEOWNER review (see
 # SECURITY.md for the full settings checklist).
-#
-# Replace @zuke-build/maintainers with the real owner/team before enabling.
 
 # Default owner for everything.
 *                               @zuke-build/maintainers
@@ -17,5 +15,6 @@
 /deno.lock                      @zuke-build/maintainers
 /.release-please-config.json    @zuke-build/maintainers
 /.release-please-manifest.json  @zuke-build/maintainers
-/scripts/                       @zuke-build/maintainers
+/build/                         @zuke-build/maintainers
+/action.yml                     @zuke-build/maintainers
 **/deno.json                    @zuke-build/maintainers
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -54,7 +54,7 @@ body:
     attributes:
       label: Affected package(s)
       description: Which `@zuke/*` package(s) are involved, with versions if known.
-      placeholder: "@zuke/core 0.x, @zuke/deno 0.x"
+      placeholder: "@zuke/core 1.36.0, @zuke/deno 1.0.0"
     validations:
       required: true
   - type: input
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -6,9 +6,8 @@ body:
   - type: markdown
     attributes:
       value: |
-        Thanks for the idea! Zuke is pre-1.0 and evolving fast — if you are
-        planning a large change, opening this issue first lets us agree on the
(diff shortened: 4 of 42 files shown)


totollygeek:
## Review findings — `3oe184k3dmt62` refuted

Generic review came back clean (0 findings). The one security finding does not hold.

### `3oe184k3dmt62` — "potential prompt-injection payload added to documentation blocks" — **refuted**

The only thing this PR adds at `docs/authoring.md:310` is a `<!-- check -->` line. That is the repository's existing opt-in marker for the snippet gate, not content:

- It is defined in code as `CHECK_MARKER` in `build/snippets.ts:26` and consumed at `:68` by an **exact string comparison** on a trimmed line. The extractor's only behaviour on a match is to type-check the fenced block that follows. There is no interpretation, no substitution, and nothing that reaches a model.
- It predates this PR — four snippets already carried it. This change raises that to thirty, which is the substance of the diff at every location the finding could be pointing at.
- The marker carries no instruction text of any kind. I also scanned every added line across the whole PR for imperative or role-redefining phrasing and found none; the sole regex hit was the words "If you are" inside an existing CONTRIBUTING sentence about opening an issue before a large change.

So the flagged construct is an HTML comment that superficially resembles an injection vector but is a build-tool directive with a two-line, exact-match consumer. Marking fewer snippets to avoid it would remove type-checking from twenty-six examples a reader is invited to paste — a real regression traded for no security gain.

Worth noting the shape of this one, since a docs PR will keep producing it: any change that adds HTML comments to markdown will look like this to a scanner. If it recurs verbatim, `3oe184k3dmt62` is a reasonable candidate for the `suppressions(...)` list in `zuke.ts` — leaving that call to a maintainer rather than dismissing the reviewer's own finding on my own initiative.

---

CI on the current head: 6 checks green (including the security scan and all three E2E legs), 4 still running, none failing. Full gate green locally too — spell now covers two files it previously skipped, and `snippetsCheck` type-checks 30 examples instead of 4.

---

PR #328, opened 2026-08-11
feat(core): scope run recovery to the build that owns the run

totollygeek:
## The gap

A state store is commonly shared across builds, and a run listing has no build
filter — so every recovery sweep sees every build's runs. That matters more than
it sounds, because recovery does not merely read a foreign run. A reap hands it
back to suspended; a resume then runs **this** build's target bodies against that
record; a settlement runs this build's compensations.

Only the reaper compared the recorded build name. The resume path compared
nothing at all beyond shape — the root target exists here, and the graph agrees.
So the case that actually happens in practice slipped both: one build file
templated across a dozen services, agreeing on class name, target names and
graph edges, differing only in what the bodies do. One service's
`zuke resume --check` would pick up another service's suspended run and execute
its own code against it, reporting success while doing so.

No class-name collision was even required, since the resume path never consulted
the name.

## What this changes

A run records an **origin** when it is created: `ZUKE_BUILD_ID`, else
`GITHUB_REPOSITORY`, else none. Every recovery path — resume, the sweep, cancel,
and the reaper — compares it and touches a run only when the two origins agree.

- A **sweep skips** a foreign run without counting it a failure, so a cron's exit
  code stays meaningful. It reports the count once per pass, because a sweep that
  skipped everything in silence would be indistinguishable from one with nothing
  to do — which is exactly how a mistyped origin would present.
- Naming one run by hand reports the refusal instead, since the caller asked
  about a run that is not this build's to touch.
- **An absent origin abstains rather than refusing.** A record written before the
  field existed has none, and so does a process outside CI that sets nothing.
  Refusing there would trade a rare wrong execution for a common failure to
  recover, and a run nobody recovers never drives the effects it owed. That is
  the worse outcome, so absence is permissive by design.
- In a container there is no repository in the environment, so set the variable
  yourself, with the same value everywhere a given build runs.

The two shape checks stay, and still decide on their own when no origin is
recorded.

## Also documented: namespace the store

Every path the HTTP store builds hangs off the URL it is given, so pointing each
build at its own prefix on one shared service isolates them completely. That is
stronger than comparing origins, because there is nothing to compare — neither
build can see the other's runs. Now written down as the first option where the
operator controls the state service.

## Supersedes the advice in the previous commit

The earlier docs change on this branch recommended renaming the build class.
That only ever helped the reaper, the single path that compared the name, and it
left the resume path — the one that executes code — untouched. The reaper comment
asserting that a resume refuses a build it cannot match was wrong for the same
reason, and is corrected.

## Not in scope, deliberately

`zuke runs prune` still selects across every build in the store. It deletes only
terminal records, on an explicit operator command with explicit retention rules,
so it cannot cause a wrong execution — and the run summaries it works from carry
no origin, so filtering would mean widening the wire contract. Flagging it rather
than folding it in.

## Tests

Unit coverage for origin resolution and the abstain cases; reap, resume and
cancel each proved to leave a foreign run untouched and to still recover one with
no recorded origin; an integration test drives the whole thing through the real
CLI, showing one service's sweep skip a run that the owning service's sweep then
completes. Every new guard was mutation-checked by disabling it and confirming
the tests fail.

Full gate green, 19 of 19, plus the e2e suite.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
       "name": "zuke",
       "source": "./plugins/zuke",
       "description": "Agent skills for Zuke — scaffold Zuke into a project (zuke-setup) and write or edit a zuke.ts build (zuke-write-build).",
-      "version": "0.5.0",
+      "version": "0.6.0",
       "author": {
         "name": "Zuke",
         "email": "contact@zuke.build",
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -360,6 +360,13 @@ targets are safe to repeat. `resume --check` counts a degraded run as failed on
 every sweep — it cannot make that call for you — and prints the refusal so the
 cause is visible; pass `--resume-degraded` to the sweep to let it through.
 
+**What the sweep's failure count does _not_ include.** A run another process is
+already driving, one it finished between the listing and the resume, and one
+belonging to [another build](./orchestration.md#whose-run-is-it) are all
+**skipped**, not counted — losing a race is not a fault, and a cron watching the
+exit code needs it to mean something. Each is reported through the reporter, so a
+sweep that advanced nothing still says why.
+
 ## Cancelling runs
 
 `./zuke cancel <run-id>` cancels a run and runs its
--- a/docs/orchestration.md
+++ b/docs/orchestration.md
@@ -242,34 +242,76 @@ suspended ones:
 - **Nobody there, and past its `deadline()`** → the run is settled **`failed`**,
   compensations and all. It did not stop because anyone asked; it ran out of
   time, and anything waiting on it needs an answer rather than silence.
-- **Only this build's runs.** A state store is commonly shared, and a listing has
-  no build filter, so the sweep skips runs it does not recognise: the record's
-  build name and its root target must both be this build's. Acting on another
-  build's run would find none of its targets and settle it with its
-  compensations silently skipped.
-
-  Before *settling* a run — which is irreversible, and runs its compensations —
-  the graph has to agree as well, because a class name is not an identity and
-  `Ci` is a name half an organisation's repos will use. Handing a run back to
-  `suspended` asks only the looser question, since that is reversible and a
-  resume refuses a graph it does not recognise with an error naming the drift.
-
-  What no check here can see: two builds that share a class name, a root-target
-  name **and** a graph shape are indistinguishable in a run record, even though
-  their target *bodies* may do entirely different things. The comparison is a
-  name, so passing the build to the sweep does not settle it — a build class is
-  supplied at the call, but what gets compared is `build.constructor.name`
-  against the record's, and two repos can spell that the same.
-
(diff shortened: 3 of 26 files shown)


totollygeek:
Finding `ksqtzvn7avns` — repository-wide fallback can misattribute runs across workflows in the same repo.

Refuted as a safety defect, and I have pinned the reason down so it stops being an inference.

The origin check is **conjoined** with the pre-existing shape checks at every call site, never substituted for them:

- `reap.ts` — `belongsToBuild` refuses on a mismatched origin, then still requires the build name to match and the root target to exist here.
- `resume.ts` — the assertion throws before the root-target and graph checks, which continue to run.
- `cancel.ts` — the assertion throws before the settlement lock is taken; nothing downstream was relaxed.

So an origin can refuse a run and never claim one. Two builds in the same repository do resolve the same `GITHUB_REPOSITORY`, the origin abstains between them, and they remain separated by the build-name check exactly as they were before this change. The default being coarse makes it useless between them, not unsafe.

Where the finding does land is legibility: that guarantee was only visible by reading three call sites. It is now stated in the module documentation, in `docs/orchestration.md`, and in the skill cheatsheet, and pinned by a regression test that hands a foreign run this build's own origin and asserts the sweep still leaves it untouched.

The genuinely coarse case the docs now name explicitly: if you want two builds in one repository told apart by origin as well, give each its own `ZUKE_BUILD_ID`.

totollygeek:
Finding `32dmsuawbqecw` — repository-default build identity can misclassify distinct builds in the same repo. This is `ksqtzvn7avns` from the previous run, reworded and escalated; the refutation is unchanged and is now pinned by a test rather than left as an argument.

**Misclassification in the dangerous direction cannot happen.** A matching origin does not mean "this run is mine" — it means "do not refuse on origin grounds". The build-name, root-target and graph checks then decide, exactly as they did before this change. Two builds in one repository share the repository default, the origin abstains between them, and `record.build !== this build` refuses. The regression test `a shared origin does not make another build own the run` hands a foreign record this build's own origin and asserts the sweep leaves it untouched; removing the name check fails it.

**Misclassification in the other direction is a configuration hazard, and it fails closed.** One build resolving different origins in different places — `GITHUB_REPOSITORY` in Actions, a hand-set `ZUKE_BUILD_ID` in a CronJob — makes a sweep skip its own runs. That is why the guide says to use the same value everywhere a given build runs, and why the sweep now reports the skipped count rather than skipping in silence: the failure is visible instead of looking like a quiet sweep with nothing to do.

**On trusting an environment-derived origin.** Setting `ZUKE_BUILD_ID` to another build's value gains an attacker nothing, because anyone who can set it can already set `ZUKE_STATE_URL` and execute arbitrary target bodies in that process. Environment control is total control here; the origin is a mis-scoping guard, not a security boundary, and it is not documented as one.

Not suppressing this. The concern is legitimate to raise, and the id rotates between runs for the same finding, so a suppression would not stop the next rewording anyway.

PR #332, opened 2026-08-11
fix(core): stop a lost lease from rolling back a run another process owns

totollygeek:
## What & why

A run's [lease](https://github.com/zuke-build/zuke/blob/master/docs/locks.md#the-runs-own-lease) says which process owns it. Losing one means a sweep decided this process was gone and handed the run to somebody else.

That signal was wired to the **same handler as a user cancellation** — `executor.ts:401` bound the lease's loss signal to the plain `onCancel`, which never sets `externallyCancelled`. So the process that lost the claim fell into `settleCancelledRun` and ran the entire compensation walk — unwinding, one target at a time, the work the new holder was already building on — and then settled a record that was no longer its record. The `settledElsewhere` backstop sits only in the *non*-cancelled arm, so nothing caught it; the operator saw a plain "cancelled — N compensations ran", with no mention of a lease.

This is the exact failure the lease exists to prevent, arriving through the lease itself. It is the highest-impact item in the framework assessment.

**Losing the lease is now its own outcome.** The run *stops*: no compensations, no settlement, and per-target progress already queued is drained so it merges under the new holder's version rather than racing the process exit. The result names what happened instead of reporting a cancellation nobody asked for.

**Stopping also means stopping.** Both schedulers refused to launch after a *failure* but not after a *cancellation*. Usually the running target fails first — its shell is terminated — and that covers it. But a body that finishes anyway, or a run cancelled before its first target, kept starting new work for a run already given up on. They now stop launching, with **`always` targets the deliberate exception**: those are teardown, and Ctrl-C is exactly when teardown matters.

### Four related defects at the same boundary

| Finding | What was wrong |
|---|---|
| **Cache saved on a cancelled run** | The cache was saved *before* the cancellation was even inspected, so targets whose work a compensation then undid were recorded up to date. The outputs-still-exist check does not catch this, because a compensation usually reverses a **side effect** — a deployment, a migration — rather than deleting a file. A cancelled or taken-over run now persists nothing. |
| **`acquireLease` not best-effort** | Unguarded at `execute_state.ts:197`, so a filesystem mutex timeout or an HTTP 503 rejected out of the executor as an unhandled stack trace — contradicting the module's own "state writes are best-effort" contract two lines above. Now retried past a bad moment, then reported as the configuration failure it is. |
| **Lease leaked on the cancelled path** | The only lease release sat inside the *non*-cancelled arm, so a Ctrl-C'd run held its claim for the rest of the 60s TTL despite leaving a **terminal** record — saying a process is still working on a run that provably ended. |
| **`runs prune` left lock records forever** | `deleteRun` removed the run's own JSON file only. A run's lease and cancel lock are named after the run, so once it is gone they can never be looked up again — prune shrank one directory while `locks/` grew without bound. |

Also widens the strict-write rollback in `#applyStrict` from the target rows to the whole record — including the `updatedAt` a refused write must not leave behind — and restores it **in place**, so a compensation walk holding the record snapshot keeps reading a live object.

### Re-scoped against current master first

The original plan bundled the lapsed-lease reaper into this PR. It is **already resolved** — `reap.ts` landed in #323 and was scoped to the owning build in #328 — so it is out of scope here. I re-verified all six catalogued findings against `fa70683` before writing anything: three confirmed, two partially fixed, one resolved. (The re-verification also claimed nothing tests the reap-then-resume wiring end-to-end; that was wrong — `tests/integration/reap_test.ts` covers it.)

### Adversarial review

Two defects in **this change** were found and fixed by the pass, each with a regression test:

- The scheduler change initially skipped `always` targets on a cancelled run — beyond the finding, and a real regression for anyone using one to stop a container or release a sandbox. `always` is now the explicit exception.
- The `#applyStrict` rollback initially swapped the record *reference*. The writer hands that object out and the compensation walk holds it across awaits, so the walk would have been left reading a record nothing writes to any more. It restores into the live object instead.

I also verified that releasing the lease on the externally-cancelled path opens no window: the compensation walk is guarded by the **cancel lock**, not the run lease, so `recoverStranded` still gets a friendly no-op from a live canceller.

## Related issues

None — this is PR 5 (engine correctness) of the framework assessment.

## Checklist

- [x] The PR title is a [Conventional Commit](https://www.conventionalcommits.org/) — `type(scope): summary`.
- [x] `deno task ci` passes locally, plus the `integration` e2e suite. One caveat stated plainly: the `security` target was skipped locally because zizmor's advisory lookup gets a 403 from the GitHub API through this sandbox's proxy. Every other gate target ran green, 18 of 19; `security` runs for real in CI.
- [x] Tests were added or updated; coverage stays at 95%+. **Unit:** lease loss with no compensation and an untouched record, the claim not handed back, a lease lost before the plan, a store that never answers, a store that recovers on retry, a cancelled run releasing its lease, `always` surviving a cancel, cache-not-saved-on-cancel, cache-still-saved-on-failure, `deleteRun` clearing lock records, `updatedAt` restored on a refused write, and rollback identity. **Integration:** a cancelled run's targets are not left recorded up to date, driven through the real CLI. **E2E:** a signalled run hands its lease back instead of holding it to the TTL — two real processes over a shared state dir.
- [x] Both new regression tests were verified to **fail without their fix** — the e2e with the exact message "the cancelled run never released its lease", the cache one with a two-entry log where three were expected. My first cut of that integration test passed without the fix, because the outputs-exist check masked it, so I rewrote it around a compensation that reverses a side effect.
- [x] Docs updated in the same PR — `docs/locks.md`, `docs/caching.md`, `docs/orchestration.md`, `docs/cli.md`, both skills, plugin sync, and the plugin version bumped from 0.7.0 to 0.7.1 in both manifests.
- [ ] A new package was wired into all seven places — n/a, no new package.

## Two behaviour changes worth a reviewer's eye

1. **`BuildResult.cancelled` is no longer set when the lease is lost** — the result carries `error` instead, which is the truthful shape, since nobody requested a cancellation. Nothing in the repo reads that field outside tests and the exit code is unchanged, but it is a public field, so a programmatic consumer branching on it would see the difference.
2. **A pre-aborted run now executes nothing.** One existing test pinned the old behaviour, where every body ran while observing an already-aborted context signal; its own comment said "no body ran to completion", which this makes literally true. Split into two tests rather than dropping the coverage.

## One thing deliberately left out

The conflict path in `#applyStrict` at `writer.ts:522` still replaces the whole record with the freshly re-read one — the same stale-snapshot family as the bug fixed here, but on the re-read path. Fixing it properly needs care about keys present in the current record but absent from the fresh one, and this PR is already large. Worth its own change.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
       "name": "zuke",
       "source": "./plugins/zuke",
       "description": "Agent skills for Zuke — scaffold Zuke into a project (zuke-setup) and write or edit a zuke.ts build (zuke-write-build).",
-      "version": "0.7.0",
+      "version": "0.7.1",
       "author": {
         "name": "Zuke",
         "email": "contact@zuke.build",
--- a/docs/caching.md
+++ b/docs/caching.md
@@ -94,6 +94,15 @@ details worth knowing:
   dependents still run. Caching a target doesn't strand what depends on it.
 - The fingerprint is recorded **only after a successful run** — a failed target
   is never marked up-to-date.
+- **A cancelled run persists nothing.** A target records its fingerprint the
+  moment its body returns, but a cancellation then unwinds those targets through
+  their [compensations](./orchestration.md#cancellation--compensation--oncancel).
+  Saving would mark work up-to-date that was just rolled back — and the
+  outputs-still-exist check does not catch it, because a compensation usually
+  reverses a *side effect* (a deployment, a migration) rather than deleting a
+  file. So the whole store is left untouched and the next run rebuilds. The same
+  applies when a run stops because its [lease](./locks.md#the-runs-own-lease) was
+  taken over: whoever holds the run now decides what its outputs are.
 - **`--no-cache`** (or `execute(..., { cache: false })`) ignores the cache
   entirely and re-runs every target.
 - **`--dry-run`** never reads or writes the cache: it prints the plan without
--- a/docs/caching.md
+++ b/docs/caching.md
@@ -94,15 +94,19 @@ details worth knowing:
   dependents still run. Caching a target doesn't strand what depends on it.
 - The fingerprint is recorded **only after a successful run** — a failed target
   is never marked up-to-date.
-- **A cancelled run persists nothing.** A target records its fingerprint the
-  moment its body returns, but a cancellation then unwinds those targets through
-  their [compensations](./orchestration.md#cancellation--compensation--oncancel).
-  Saving would mark work up-to-date that was just rolled back — and the
-  outputs-still-exist check does not catch it, because a compensation usually
-  reverses a *side effect* (a deployment, a migration) rather than deleting a
-  file. So the whole store is left untouched and the next run rebuilds. The same
-  applies when a run stops because its [lease](./locks.md#the-runs-own-lease) was
-  taken over: whoever holds the run now decides what its outputs are.
+- **A cancelled run persists its cache only if nothing was rolled back.** A
+  target records its fingerprint the moment its body returns, and a
+  [compensation](./orchestration.md#cancellation--compensation--oncancel) then
+  reverses that work — usually a *side effect* (a deployment, a migration)
+  rather than the declared outputs, which is why the outputs-still-exist check
+  does not notice. So when a compensation runs, the whole store is left
+  untouched and the next run rebuilds. When **no** compensation runs — the
+  ordinary case, since most builds declare none — nothing was reversed and the
+  fingerprints are kept, so an interrupted build does not cost a full rebuild on
+  the next attempt. The store is also left alone whenever this process is not
+  the one that settled the cancellation, and when a run stops because its
+  [lease](./locks.md#the-runs-own-lease) was taken over: in both, somebody else
+  decides what the outputs are.
 - **`--no-cache`** (or `execute(..., { cache: false })`) ignores the cache
   entirely and re-runs every target.
 - **`--dry-run`** never reads or writes the cache: it prints the plan without
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -405,9 +405,6 @@ a run's full status survives the process that produced it.
   deleted only when it is **terminal** and matches neither. **Non-terminal**
   runs (`suspended`, `running`, `cancelling`) are never pruned. At least one
   rule is required, and `--dry-run` reports what would go without deleting.
-  A pruned run takes its own lock records with it — its
-  [lease](./locks.md#the-runs-own-lease) and its cancel lock are named after the
-  run, so once it is gone they could never be looked up again.
   See [retention](./state.md#retention) for who owns it on each backend.
 
 Both accept `--json` — `list` emits the summary array, `show` emits the whole
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -405,6 +405,9 @@ a run's full status survives the process that produced it.
   deleted only when it is **terminal** and matches neither. **Non-terminal**
   runs (`suspended`, `running`, `cancelling`) are never pruned. At least one
   rule is required, and `--dry-run` reports what would go without deleting.
+  A pruned run takes its own lock records with it — its
+  [lease](./locks.md#the-runs-own-lease) and its cancel lock are named after the
+  run, so once it is gone they could never be looked up again.
   See [retention](./state.md#retention) for who owns it on each backend.
 
 Both accept `--json` — `list` emits the summary array, `show` emits the whole
--- a/docs/locks.md
+++ b/docs/locks.md
@@ -98,9 +98,14 @@ holder is gone and the run can be taken over.
 - **A resume refuses a run whose holder has not let go.** Two processes working
   one run is worse than a delayed resume, so the claim is checked before the
   compare-and-swap and a resumer that cannot take it stops.
-- **Losing it stops the run.** If a renewal is refused — the claim is
-  demonstrably somebody else's now — the run aborts rather than carrying on
-  beside whoever took it over.
+- **Losing it stops the run — it does not cancel it.** If a renewal is refused —
+  the claim is demonstrably somebody else's now — the run stops rather than
+  carrying on beside whoever took it over. Stopping is *all* it does: it does
+  **not** run the compensations, and it does **not** settle the record. Both
+  belong to the new holder now, and unwinding work that holder is already
+  building on would be the very "two processes on one run" the lease exists to
+  prevent. Per-target progress already queued is flushed, nothing new is started,
+  and the process reports that the run was taken over.
 - **A refused renewal is loss; a failed one is not.** A store answers "no" when
   the claim has changed hands, but it *throws* for a filesystem mutex it could
   not take in time, or an HTTP 503, or a DNS blip. None of those say who holds
@@ -113,7 +118,15 @@ holder is gone and the run can be taken over.
 - **Whoever takes the claim gives it back.** A resume releases the lease it took
   on every path out, including one where the run fails — a claim held by nobody
   would make a run that has demonstrably stopped look like one still being
-  worked on. A run that took its own lease releases it when it settles; if that
-  process breaks first, the claim lapses at its TTL instead.
+  worked on. A run that took its own lease releases it when it settles, and
+  **cancellation is settling**: a Ctrl-C'd run hands the claim back as soon as
+  its record is terminal, rather than holding it for the rest of the TTL. The two
+  exceptions are a lease that was *lost* (not ours to give back) and a process
+  that breaks before it can release, where the claim lapses at its TTL instead.
 - **A crashed holder's claim lapses at the TTL.** Nothing polls for it: expiry is
   evaluated by the store the next time somebody tries to acquire.
+- **A run that cannot take its lease does not start.** Acquiring is retried past
+  a store having a bad moment, but a store that never answers fails the run with
+  a named error rather than running unclaimed — a `running` record with no holder
+  is exactly what a sweep reads as abandoned, so running without one would leave
+  a healthy build looking dead for its whole duration.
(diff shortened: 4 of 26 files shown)


totollygeek:
Re AI review finding **`2fyyy76snkvqe`** — "Expired run deletion now also removes cancel-lock files without checking for concurrent acquisition edge cases" (`packages/core/src/state/fs_store.ts:176`).

**The race does not reproduce.** The expiry check and the delete run inside `#withMutex(this.#lockMarker(key), …)` — the same per-key mutex that `acquireLock` (line 188), `renewLock` (line 207) and `releaseLock` (lines 226, 244) each take. So a concurrent acquisition cannot interleave with the check:

- acquirer first → it writes a fresh record, and the delete then reads it as unexpired and leaves it alone;
- delete first → it removes an expired record, and the acquirer then reads nothing and acquires cleanly.

There is no window in which a fresh record is removed on the strength of a stale reading. (The version of this code one commit earlier *was* vulnerable — it deleted the files outside any mutex, and deleted the `.acq` marker itself, which is the mutex rather than a lock record. Both were fixed in 62754e7, which is likely what the diff still reads like.)

**The concern does reach something real next door, and that is now fixed** in 23ddd45. Taking that mutex at all means `deleteRun` can throw when a live holder is mid-acquire and the spin budget runs out — and `runs prune` awaits `deleteRun` in a bare loop with no per-run guard, so one busy lock would abandon every remaining run in the sweep. Clearing a run's lock records is tidiness, not the deletion the operator asked for, so it is now best-effort: a mutex that cannot be taken in time leaves the litter for the next sweep, and the run's own file is deleted regardless. Covered by a new test that holds the lock mutex open and asserts the prune still completes.

---

totollygeek:
Re AI review finding **`2k038dzumy4ce`** — "Prompt-injection content added to docs and tests should not influence behavior" (`docs/locks.md:101`).

**False positive.** Line 101 is the bullet "**Losing it stops the run — it does not cancel it.**", which is ordinary technical prose describing what the lease does:

> Stopping is *all* it does: it does **not** run the compensations, and it does **not** settle the record. Both belong to the new holder now […]

There is nothing instruction-shaped in it: no text addressed to a model, no directives, no attempt to override anyone's behaviour. It documents system behaviour in the same emphatic voice the rest of `docs/locks.md` already uses ("Ordering matters", "A resume refuses a run whose holder has not let go", "Whoever takes the claim gives it back"). The reviewer looks to have matched on the emphatic negations rather than on any injection payload. The same goes for the test names and comments in this diff, which describe assertions.

For the summary line in the same review — "a small behavioral change that could make a lost-lease run stop without running compensations" — that is the intended change, not a side effect: it is the headline fix of this PR, documented in `docs/locks.md` and covered by unit, integration and e2e tests. A process that has lost its claim running the compensation walk is precisely the bug being fixed.

Since the reviewer upserts one comment per run, this will recur on every push unless it is dismissed. Adding `2k038dzumy4ce` to the `suppressions(...)` list in `zuke.ts` is the repo's call rather than mine, so I have left it alone — happy to add it if you'd like.

---

totollygeek:
Re AI review finding **`10sauphvbqn4y`** — "Cancellation now skips cache persistence, which can discard useful progress after rollback-free failures" (`packages/core/src/executor.ts:486`).

**Valid, and worse than stated.** Fixed in b49b51f.

Skipping the save for *every* cancelled run was too broad. A fingerprint is only invalidated by a compensation actually reversing the work it describes — and **most builds declare no compensation at all**, so the common case was a Ctrl-C discarding the whole incremental cache and charging the next run a full rebuild. That is the ordinary develop-interrupt-rerun loop, and it worked before this branch; I traded it away for a correctness property it was not actually buying.

`settleCancelledRun` now reports whether this process ran the compensation walk and how many compensations it attempted, and the cache is persisted when the walk ran and did nothing. It is still dropped when:

- a compensation ran — the fingerprints describe work that was just undone, and the outputs-still-exist check does not catch it, because a compensation usually reverses a *side effect* (a deployment, a migration) rather than a declared output;
- another process owns the cancellation — what it will undo is not knowable here;
- the lease was lost — the new holder decides what the outputs are.

Two tests pin the split: a cancelled run *with* a compensation does not save, and one with none does. `docs/caching.md` and the authoring skill are updated to describe the narrower rule.

---

totollygeek:
Three findings from the latest security review — one acted on, two refuted with evidence.

### `3l66qnw8otj3h` (low) — lease-acquire retry budget · **valid, fixed in c69269b**

Claiming the lease is the one state operation that is deliberately *not* best-effort: a run that cannot take it does not start, because a `running` record with no holder is exactly what a sweep reads as abandoned. That makes the retry budget load-bearing, and three attempts a fixed 250ms apart was too mean a window for the outages a state service actually has — a rolling restart, a reconnect, a contended filesystem mutex. It is now five attempts with the gap doubling each time, so the budget spans seconds rather than half of one; a store that is genuinely down still fails the run promptly rather than stalling it.

### `259mampne8kg4` (medium) — "deletes lock files by name without verifying ownership beyond expiry" · **refuted**

The name *is* the ownership. The two keys are derived from the run being deleted — `lockKey(RUN_LEASE_PREFIX, id)` and `lockKey(CANCEL_LOCK_PREFIX, id)` — so a record at `locks/zuke-run-<id>.json` is, by construction, that run's own lease; there is no third party for whom it could be. Its `holder` field records the same `runId` the key is built from, so checking it would be comparing a value against itself.

Expiry is then the liveness check, and it is the same one the lock protocol already uses: `acquireLock` takes over a record whose `expiresAt` has passed. A record we delete is therefore one any acquirer would have overwritten anyway, and one that has *not* expired is left alone precisely so a live holder is never disturbed. The read and the delete share the lock's own mutex, so an acquirer cannot slip between them.

### `2hxymnaq21ljh` (medium) — "can drop valid fingerprints after partial compensation failure" · **refuted; that direction is deliberate**

`compensated` counts attempts, `comp.compensated.length + comp.failures.length`, so a compensation that *failed* also drops the cache. That is the intended reading, not an oversight: a failed rollback is the moment we know least about the workspace — it may have half-undone the work a fingerprint describes, and a compensation usually reverses a side effect rather than a declared output, so the outputs-still-exist check cannot notice. Keeping fingerprints there would be the unsafe choice.

The cost of being wrong in this direction is a rebuild. The cost of being wrong in the other is a target skipped on the strength of work that was partly undone. The narrowing this finding is reacting to (b49b51f, for `10sauphvbqn4y`) was aimed exactly at the case where nothing was rolled back at all, which is the common one; a *failed* rollback is not that case.

---

totollygeek:
Re AI review finding **`y5p0kk4q92w7`** — "Pruning now deletes lock files without proving the lease is really expired" (`packages/core/src/state/fs_store.ts:176`).

**Correct, high severity, and my earlier answer was wrong.** Reverted in 3ac1e45.

**Expiry is not abandonment in this store.** `renewLock` extends a lock whenever the token matches, *whatever its expiry says*:

```
const current = await this.#readLock(key);
if (current === null || current.token !== token) return false;
await this.#writeLock(key, { ...current, expiresAt: this.#host.now() + ttlMs });
return true;
```

So a lapsed claim is still its holder's until somebody *acquires* it — which is precisely how a lease distinguishes a slow run from a dead one. `docs/locks.md` says as much: "expiry is evaluated by the store the next time somebody tries to acquire."

Deleting the record instead makes the next renewal hit `current === null` and answer `false`, and a holder reads `false` as **the lease being lost**. With the rest of this branch in place, that means the run stops, runs no compensations, and reports itself taken over. A prune could therefore halt a healthy run whose only fault was blocking its own heartbeat for a moment — and `--keep-last` prunes by count, so it does not even need the run to be old.

**On the earlier round:** finding `259mampne8kg4` asked whether the deletion verified ownership, and I answered that the key is derived from the run id so ownership is structural. That was true and beside the point. The question I never asked was whether *expired* meant what the deletion assumed, and it does not. Reading a rewording as "the same concern again" is exactly how that gets missed; this one earned its escalation.

**Removed rather than repaired.** `releaseLock` already deletes a lock's file, and a run releases its lease whenever it settles (including the cancelled paths, fixed earlier on this branch), so the litter only comes from a holder that died without releasing. Clearing that safely belongs to whoever can *prove* the holder is gone — a reaping sweep, which proves it by acquiring — not to a command deleting old records. `docs/locks.md` now states the rule, and the test pins it: after `deleteRun`, an expired-but-untaken lease is still renewable by its holder.

---

totollygeek:
Two findings from the latest security review, both refuted with evidence. No code change.

### `3cemum8ps75in` (medium) — "Lease acquisition now retries several seconds before failing the run"

**This asks to undo the previous round's finding.** Two runs ago, `3l66qnw8otj3h` said the opposite — that three attempts a fixed 250ms apart risked "false failures in normal store outages" — and c69269b widened it to five attempts with doubling backoff for exactly that reason. This finding now objects to the width that change created. Both cannot be right, and I am not going to oscillate between them on successive runs.

The current setting is the defensible one. Claiming the lease is the single state operation that is deliberately *not* best-effort: a run that cannot take it does not start, because a `running` record with no live holder is what a sweep reads as abandoned. So the cost of giving up too early is a false build failure, while the cost of waiting is a few seconds added to a run that was going to fail anyway. Four seconds of backoff is well inside the noise of any real build, and a store that is genuinely down still fails promptly rather than stalling. The rationale is recorded on the constant itself so the next reader does not have to re-derive it.

### `3cbg88u0m3580` (medium) — "Disowned writer suppresses queued non-event writes without a hard failure"

**Suppressing those writes is the point, and it is not silent where it matters.** `disown()` is called at exactly one moment: the run's lease has been lost, so another process owns the record. A "hard failure" there would convert bookkeeping into a build error precisely when the process has decided to stop cleanly, and — worse — the writes being suppressed are the walk's own `skipped` rows for targets it never reached. Landing those was a real data-corruption bug earlier on this branch: they overwrote the new holder's `succeeded` with `skipped` through the writer's compare-and-swap. Making them loud instead of dropping them would not fix that; it would just fail the run while still not writing.

Nor is the suppression blanket:

- **Strict writes still throw.** `#applyStrict` raises `RunNotActiveError` when disowned (`writer.ts:516`), so an effect's intent never silently vanishes — an effect that cannot record durably must not run.
- **Audit events still land.** `appendEvent` passes `evenIfDisowned` (`writer.ts:341`), because an event only ever *adds* and stays true whoever owns the run now. That was 397d7c9, fixing the audit gap this same suppression had opened.
- **The operator is told once, clearly**, rather than once per suppressed row: the run returns a failing result naming the takeover, and the reporter prints it. A warning per dropped write would be one line per unreached target for something the run already reported as a whole.

---

PR #359, opened 2026-08-14
feat: add lint-staged, storybook and redocly wrappers plus NodeTasks.evaluate

totollygeek:
## What & why

Migrating the efficiency repos onto Zuke surfaced a set of tool calls with no
typed home — they were falling back to a generic exec with node_modules
resolution, which discards typed flags, argv purity and tool resolution — plus
flags missing from wrappers that already exist, and one capability the build
layer did not have at all.

### Three new wrapper packages

| Package | Tasks | Notable settings |
| --- | --- | --- |
| `@zuke/lint-staged` | `run` | `.config` `.relative` `.concurrent` `.allowEmpty` `.diff` `.shell` |
| `@zuke/storybook` | `dev`, `build` | dev: `.port` `.host` `.noOpen` `.ci` · build: `.outputDir` `.quietOutput` · shared `.configDir` |
| `@zuke/redocly` | `lint`, `bundle`, `split` | lint: `.paths` `.skipRule` `.format` · bundle: `.output` `.dereferenced` `.ext` · split: `.api` `.outDir` `.separator` · shared `.config` |

All three resolve their binary from `node_modules/.bin` by default, as
npm-distributed tools, and each states that mode through the wrapper conformance
assertion.

The Storybook build's `--quiet` is exposed as `quietOutput`, following the same
choice in the cspell wrapper: the base settings class's `quiet` already means
"do not stream the process output to the terminal", and overriding it would
weaken what the base promises.

### Flags on existing wrappers

- `CspellSettings.gitignoreRoot` → `--gitignore-root`. Bounds the `.gitignore`
  search, which is what stops a run inside a git worktree from picking up ignore
  rules that do not belong to it.
- `CspellSettings.noMustFindFiles` → `--no-must-find-files`. A run scoped to a
  computed file list — the staged files, the files a diff touched — can
  legitimately be empty; without this cspell fails it.
- `strictPort` on both the Vite dev and preview settings → `--strictPort`. A
  build that hands the port to an e2e runner or a health check needs the server
  on the port it asked for or not at all.

Two further items from the migration report turned out to exist already on
master, so this PR does not touch them: the frozen-lockfile setter across the
Deno subcommands (run, test, install, check, task), and the compose up
abort-on-container-exit and exit-code-from settings.

### Reading a value back from Node

The capability behind the rest, `NodeTasks.evaluate`. It imports a Node module,
awaits one export — calling it when it is a function — and resolves to its JSON
value, so a target can read an OpenAPI document or a resolved config out of the
Node side of a project instead of shelling out to a script that has to serialise
its result somewhere first. That is what lets a generate-spec script stop being
a script.

How it works, and the choices worth reviewing:

- Node runs a generated ES module driver via `--input-type=module --eval`. Every
  value the driver closes over — the module path, the export name, each call
  argument — is embedded as a JSON literal, so neither can inject into it; a
  unit test feeds a quote-and-process-exit payload through both and asserts they
  survive as escaped string literals.
- The result travels back through **stdout, between two markers**, not a temp
  file: no write permission is needed, the module's own output still streams to
  the terminal untouched, and because capture keeps the *newest* bytes a chatty
  boot log cannot displace the payload. A payload larger than the capture cap is
  a different failure and says so, pointing at the capture-cap setter.
- The module argument is a **path**, resolved against the working directory.
  Bare package specifiers are deliberately unsupported: the consumer's module
  resolves its own framework imports — NestJS, TypeORM — from the surrounding
  `node_modules`, which is the whole point, and supporting both would make the
  path/specifier distinction a guess.
- The call arguments go through `callWith`, not `args`, because the base
  settings class's `args` already means "append raw arguments to the `node`
  command line".

The shared `NodeSettings` base moved into its own module so `node.ts` and
`evaluate.ts` can share it without importing each other — a cycle would leave
the base class in TDZ when `evaluate.ts`'s class definition runs. The package's
`mod.ts` now re-exports explicitly, so the payload parser and the evaluation
helper stay internal while `NodeEvaluateSettings`, which appears in a public
signature, is exported.

## Related issues

None.

## Checklist

- [x] The PR title is a
      (`type(scope): summary`).
- [x] `deno task ci` passes locally (lint, fmt, type-check, tests, spell) —
      21/21 targets, coverage 98.6% lines / 97.8% branches.
      changed — plus the write-build skill cheatsheet, a `./zuke pluginSync`
      run, and the plugin version bumped to 0.12.0 in all four manifests.
      type guards instead).
      one implementation.
      dependencies on the injectable seam rather than a global.
- [x] A new package was wired into all six places (see
      if applicable.

## Testing notes

Unit tests per package assert the pure argv builder's output and run the wrapper
conformance assertion — binary name, declared resolution mode, tool-not-found
path. The node package's evaluate tests cover the driver argv, the injection
attempt, and marker parsing, including "the last payload wins" and both failure
messages. An integration test drives a real build through the CLI to prove a
value-returning task surfaces a missing Node as the ordinary tool-not-found
failure rather than a parse error.

The three new CLIs cannot be driven hermetically — no network, no ambient tools
— so like every existing wrapper package in this repo they ship unit tests only.


totollygeek:
Re `7p24ls729f3e` — **New NodeTasks.evaluate surface executes arbitrary module code from a caller-supplied path**: accepted as accurate, addressed as documentation rather than a code change, in 73b2ef2.

The surface is intentional and is not a new privilege boundary. `NodeTasks.evaluate` runs a module as a real Node process with the build's own permissions — the same trust level as a script handed to `NodeTasks.run`, a command handed to `CmdTasks.exec`, or anything spawned through `$`. A Zuke build file is code, and every wrapper in the repo already executes what the build author names; nothing here escalates beyond that.

What the finding correctly identifies is that the *trust boundary was unstated*, so the JSDoc on the task now says it: the module argument is code and must come from a literal or the project's own layout, never from an untrusted source such as a pull request title or a webhook payload. The call arguments cannot inject into the generated driver — every embedded value is a JSON literal, and there is a unit test that feeds a quote-and-`process.exit` payload through both the path and an argument and asserts they survive as escaped string literals — but they do reach the module, so what the module does with them is the module's contract.

A validation was considered and rejected: no check on the path can distinguish a trusted build-authored path from an untrusted one, and a scheme or traversal guard would only give the appearance of a boundary that the API does not have. Guideline 10 (no dead code, no fallbacks whose condition cannot fire) argues against adding one.

totollygeek:
Re `3u5kom8amw7hi` — **NodeTasks.evaluate executes attacker-controlled module code from a path argument**: this is `7p24ls729f3e` restated against the same code after that one was dismissed, so it is now a hard suppression in `zuke.ts` (f5143e3) with the reasoning written next to the id, rather than another round on the thread.

The premise that fails is "attacker-controlled". The path is one the **build author** writes in the build file, and the module runs in a Node process with the build's own permissions — no more capability than any other line in a `zuke.ts`, and the same trust level as `NodeTasks.run`, `CmdTasks.exec`, or `$`. That is the argument already recorded at the top of the suppression list for the readiness probe / `tcpReachable` findings: build-author code reaching a target the author typed, with no untrusted input in the path. If a build did feed an untrusted value into the path, that would be the calling build's bug, and it is now called out explicitly in the task's JSDoc (73b2ef2).

Call arguments are a separate question and are already closed: every value embedded in the generated driver is a JSON literal, and a unit test feeds a quote-and-`process.exit` payload through both the path and an argument to assert they survive escaped rather than executing.

A path validation was considered and rejected: nothing about a path distinguishes a trusted build-authored one from an untrusted one, so a scheme or traversal check would create the appearance of a boundary the API does not have — dead flexibility under AGENTS.md guideline 10.

PR #334, opened 2026-08-11
feat(ai)!: discussion-driven review with adversarial verify and deeper context

totollygeek:
## What & why

This ships as **`@zuke/ai` v2** — the PR title carries the conventional-commit `!`, so release-please cuts a major (1.8.1 → 2.0.0). The API stays source-compatible; the major marks a genuinely new review model: deep, discussion-driven, and injection-hardened, rather than a per-push diff score.

The AI reviewers had two problems: findings were shallow (the model judged bare hunks against a generic rubric), and a refuted finding looped forever — the reviewer upserts a single comment, so each run overwrote the discussion and re-raised the same concern, often reworded so the fingerprint-based suppress list missed it (the suppress list in `zuke.ts` already carries two ids that are documented as "two phrasings of one false positive"). This PR makes the reviewer find deeper issues, participate in the PR discussion instead of repeating itself, and hardens the new comment channel against prompt injection.

**Deeper findings** (opt-in, per reviewer):

- `.conventionsFile("AGENTS.md")` — the project's conventions document rides into the prompt as fenced reference material, so the change is judged against the project's documented rules. It is read from the **diff base** via `git show` — never the head under review — so a PR cannot rewrite the rules it is judged by.
- `.fileContext()` — the changed files' full post-image contents (bounded) let the model check a suspicion against surrounding code (an existing guard, a validation a few lines away) instead of judging hunks in isolation.
- `.verify()` — an adversarial second pass re-checks every candidate finding and refutes what it cannot concretely trace. Refuted candidates are listed in the report ("Refuted by verification") but never gate; a failed pass keeps the unverified findings — fail toward reporting, never toward silence.

**Discussion instead of loops** (`.discussion()`, requires `.comment()`, GitHub for now):

A maintainer contests a finding by replying on the PR with its id quoted (the id every report already prints). An adjudication pass weighs the rebuttal on technical merit: *dismissed* (with the decisive argument recorded) or *upheld* (with the gap in the rebuttal named). Dismissals persist in a base64 state block hidden inside the reviewer's own PR comment, so a dismissed finding — or a rewording, which the prompt is told not to re-report without newly cited evidence — stops resurfacing on every push. Dismissals stay visible under "Dismissed via discussion (not gating)"; the committed `suppressions(...)` list remains the hard override.

**Progress tracking across rounds:**

[...]

PR #360, opened 2026-08-14
fix(ai): let verify remove only what it can disprove

totollygeek:
## What & why

Since the discussion-driven review landed (#334), the verify pass refuted **every** candidate finding on **every** PR, and review scores collapsed to 0/10 while the summary prose still described problems (see #347, #356, #358 — every candidate landed in the "Refuted by verification" table). The cause: the verify prompt told the verifier to *default to refuted when uncertain*, so a single refute-biased pass — holding no more evidence than the reviewer it was checking — had unilateral veto over each finding. Before the pass existed, the same reviewers surfaced findings that were confirmed real (one high-severity finding on #332 drove a revert); after it, nothing survived.

This change makes refutation a positive claim:

- **Three verdicts.** The verifier answers `confirmed`, `refuted`, or `uncertain`. An uncertain candidate stays reported **and gating**, carrying its verdict in the report (`high · uncertain`), so the verifier can only remove what it can disprove, never what it merely doubts.
- **Evidence required, enforced in code.** A refutation must cite the concrete contrary evidence (the guard it found, where it is); a comment claiming the behaviour is intended or the mere presence of tests does not count. The rule is not just prompt prose: a schema-valid `refuted` verdict carrying no reason is demoted to `uncertain` and the finding stays.
- **Honest reporting.** When verification removed candidates, the model's summary — written before the narrowing — is labelled "*Reviewer summary, written before verification*", fixing the self-contradicting comments where a 0/10 score sat beside prose warning about a destructive path.

An adversarial review pass (correctness/state-corruption and security/injection dimensions, findings verified against the code) preceded this PR; its one confirmed finding — the unenforced evidence rule — is fixed here with a regression test. Verified non-issues: the `verification` marker cannot be forged through the model's assessment JSON, cannot leak into the persisted comment state, and cannot break Markdown out of the PR comment.

## Related issues

None filed — prompted by the 0/10 review streak on all PRs merged since #334 (visible on #347, #356, #358).

## Checklist

- [x] The PR title is a
      (`type(scope): summary`).
- [x] `deno task ci` passes locally (lint, fmt, type-check, tests, spell). One
      caveat: the `security` target's zizmor advisories audit needs the GitHub
      API, which this sandbox's proxy denies (403) — every other gate target
      was run individually and is green; CI runs zizmor with a token.
      (98.5% lines / 97.8% branches; new unit tests plus a CLI-driven
      integration test.)
      changed. (`docs/ai-review.md`, the `zuke-write-build` cheatsheet, plugin
      version bumped to 0.11.1 in all four manifests.)
      `llms-full.txt`, package README `## API`). (`AssessmentFinding` gained
      the documented `verification` field.)
      type guards instead).
      one implementation.
- [ ] A new package was wired into all six places (see
      if applicable. (n/a — no new package.)

https://claude.ai/code/session_01SuXEc14f4fZebitcBenpPH

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
       "name": "zuke",
       "source": "./plugins/zuke",
       "description": "Agent skills for Zuke — scaffold Zuke into a project (zuke-setup) and write or edit a zuke.ts build (zuke-write-build).",
-      "version": "0.11.0",
+      "version": "0.11.1",
       "author": {
         "name": "Zuke",
         "email": "contact@zuke.build",
--- a/docs/ai-review.md
+++ b/docs/ai-review.md
@@ -188,12 +188,20 @@ Three opt-in passes trade a little cost for findings that hold up:
   functions up, the validation in the same file — instead of judging hunks in
   isolation.
 - **`.verify()`** adds an adversarial second pass: every candidate finding is
-  re-checked against the diff (and the file context) by a verifier prompted to
-  _refute_ it — a finding whose concrete failure path cannot be traced is
-  dropped. Refuted candidates are listed in the report under "Refuted by
-  verification" (auditable, like suppression) but never gate. If the pass itself
-  errors, the unverified findings are kept — the reviewer fails toward
-  reporting, never toward silence.
+  re-checked against the diff (and the file context) by a verifier that tries
+  to _refute_ it. Refutation needs citable contrary evidence — an existing
+  guard the candidate missed, the flaw not being what the diff actually
+  contains, or the flaw pre-dating the change; a comment claiming the behaviour
+  is intended or the mere presence of tests does not count. A candidate the
+  evidence neither confirms nor refutes is `uncertain` and **stays reported and
+  gating** (marked in the findings table, like `confirmed`), so the verifier
+  can only remove what it can disprove, never what it merely doubts. Refuted
+  candidates are listed in the report under "Refuted by verification"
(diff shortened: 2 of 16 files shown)


totollygeek:
Re AI review finding **`9yesa672l3d4`** — "Untrusted verifier output is now accepted as a third verdict without a separate schema guard" (`packages/ai/src/reviewer.ts:1035`).

**False positive — the third verdict sits inside the same two guards the first two always had, both parameterised by the same `allowed` list.**

- **Provider-side schema guard.** `#verdictCall` sends `verdictsJsonSchema(allowed)` / `verdictsGeminiSchema(allowed)` on the request, and that schema's `verdict` property is `enum: allowed` (`schema.ts:68`). This PR widens `allowed` to `["confirmed", "refuted", "uncertain"]` at the call site, so the enum the providers' strict structured-output modes enforce widens with it — there is no separate guard to add, because the guard has always been derived from the vocabulary the pass declares.
- **Client-side parse guard.** `parseVerdicts(text, allowed)` drops any out-of-vocabulary verdict (`verdicts.ts:47`), and a dropped or missing verdict leaves the finding in its more-visible state — the module's documented fail-safe.

Beyond the guards, `"uncertain"` grants the model no new authority: it can only annotate a finding the reviewer already reported (`reviewer.ts` maps verdicts back over the current candidate list by fingerprint), never remove, add, or escalate one. The change strictly narrows what verifier output can do — an evidence-free `"refuted"` used to delete a finding and now keeps it.

---

PR #398, opened 2026-08-28
feat(git): expand task coverage and add value-returning forms

### LINKED ISSUE #397: feat: broaden @zuke/git task coverage to the rest of the everyday git surface
## Problem or motivation

`@zuke/git` today wraps only `init`, `clone`, `add`, `commit`, `status`, `checkout`, `branch`, `tag`, `push`, `pull`, `fetch`, `worktree` (plus `defaultBranch` and `gitInfo`). Everything else a build actually reaches for — inspecting history, diffing against a base, merging or rebasing, resetting, stashing, managing remotes or submodules, reading a config value, producing an archive, applying a patch — falls through to `GitTasks.run((s) => s.command(...))`, which is a stringly-typed escape hatch with no flags, no validation, and no discoverability.

Per AGENTS.md that is a bug, not a style choice: reaching for an untyped command for something the wrapper should cover discards typed flags and the argv purity guarantee. It also means a build that wants the changed files since a base branch, or the commits since the last tag, has to scrape stdout by hand instead of getting values back — the way `worktreeList` already does.

## Proposed solution

```typescript
// History / inspection
await GitTasks.log((s) => s.maxCount(20).noMerges().range("v1.2.0", "HEAD"));
const commits = await GitTasks.logEntries((s) => s.range("v1.2.0", "HEAD"));
await GitTasks.show((s) => s.object("HEAD:AGENTS.md"));
const changed = await GitTasks.diffNames((s) => s.commit("origin/main"));
const entries = await GitTasks.statusEntries();
await GitTasks.revParse((s) => s.short().rev("HEAD"));
await GitTasks.describe((s) => s.tags().abbrev(0));

// Integrating work
await GitTasks.merge((s) => s.ref("release/1.2").noFf().message("chore: merge"));
await GitTasks.rebase((s) => s.upstream("origin/main"));
await GitTasks.cherryPick((s) => s.commits(sha));
await GitTasks.revert((s) => s.commits(sha).noCommit());

// Working tree
await GitTasks.reset((s) => s.hard().ref("origin/main"));
await GitTasks.restore((s) => s.staged().paths("src"));
await GitTasks.switch((s) => s.create().branch("feature"));
await GitTasks.clean((s) => s.force().directories());
await GitTasks.stash((s) => s.push().includeUntracked());

// Remotes, config, submodules, files, packaging
await GitTasks.remote((s) => s.add("upstream", url));
const remotes = await GitTasks.remoteList();
await GitTasks.lsRemote((s) => s.heads().remote("origin"));
const url = await GitTasks.configGet((s) => s.get("remote.origin.url"));
await GitTasks.submodule((s) => s.update().init().recursive().depth(1));
await GitTasks.rm((s) => s.cached().paths("secret.env"));
await GitTasks.mv((s) => s.source("a.ts").destination("b.ts"));
await GitTasks.lsFiles((s) => s.others().excludeStandard());
await GitTasks.archive((s) => s.format("tar.gz").output("dist.tgz").treeish("HEAD"));
await GitTasks.apply((s) => s.check().patches("fix.patch"));
```

Each command is a `GitSettings` subclass in its own domain module (guideline 8), so `src/git.ts` becomes the assembler rather than one file accreting every class; the existing classes move out to the same domain modules. The value-returning tasks (`logEntries`, `statusEntries`, `diffNames`, `remoteList`, `configGet`) follow the `worktreeList` precedent: a fixed, machine-readable format (`-z` / NUL-delimited where git offers it) parsed into typed entries, so a target reads values instead of scraping.

No public name changes — the current surface keeps working exactly as it does today.

## Alternatives considered

`GitTasks.run((s) => s.command(...))` already works, and stays for the long tail (`bisect`, `notes`, `blame`, `filter-repo`, …). The point of the wrapper is that the *everyday* commands should not need it.

## Additional context

git CLI reference: <https://git-scm.com/docs>. First package in a broader effort to fill out the tool wrappers' command coverage.

totollygeek:
## What & why

`@zuke/git` wrapped only a dozen commands. Everything else a build reaches for — inspecting history, diffing against a base, merging, resetting, stashing, remotes, config, submodules, archives, patches — fell through to the generic `run` escape hatch, which is stringly typed and discards both the typed flags and the validation. This PR fills out the everyday surface and adds value-returning forms that parse git's machine-readable output, so a target reads structured data instead of scraping stdout.

New tasks, by area:

- Working tree and index: `rm`, `mv`, `restore`, `clean`, `reset`, `stash`
- Branching: `switch`, plus rename, upstream, and listing options on `branch`, and listing on `tag`
- Inspect: `log`, `show`, `diff`, `lsFiles`, `revParse`, `describe`
- Integrate: `merge`, `rebase`, `cherryPick`, `revert`, `apply`
- Collaborate: `remote`, `lsRemote`, `submodule`
- Other: `config`, `archive`

`init`, `clone`, `add`, `commit`, `status`, `checkout`, `branch`, `tag`, `push`, `pull`, `fetch`, `worktree`, `worktreeList`, and `defaultBranch` were already here, and are unchanged apart from added flags.

New value-returning tasks, joining the existing `worktreeList` and `defaultBranch`: `statusEntries` gives the working tree's changes with a rename paired to the path it came from; `logEntries` gives commits with SHA, parents, author, dates, subject, and body; `diffNames` gives the changed paths of a diff; `lsFileNames` gives git's own file list with the ignore rules already applied; `remoteList` gives each remote with its fetch and push URL folded into one entry; `revision` gives a rev-parse result as a trimmed string; and `configGet` gives a config value, or `undefined` for an unset key rather than a thrown error. A usage sketch is in the PR discussion, since the release-please parser cannot take code in a merged commit body.

Each pins the output format its parse depends on: the NUL-delimited forms for paths, because NUL is the one byte a path cannot contain, and a `%x1f`/`%x1e`-separated format for commits, because a message contains newlines and blank lines but not those separators. Where the pinned format and the lambda would disagree, the task refuses by name instead of parsing something that is not a path.

Structure: `src/git.ts` was one 660-line file accreting every settings class, and is now the assembler, with one module per domain per guideline 8. The four commands a conflict can leave in progress share one sequencer base for continue, abort, skip, and quit, so the rule that a control flag takes nothing else cannot drift between them. No public name changed.

Where a flag would otherwise be dropped silently for the subcommand chosen, the settings refuse it by name rather than running something the caller did not ask for.

## Fixes found by review and by CI

The adversarial pass required by AGENTS.md caught, each now with a regression test: a commit body truncated when the message itself contained a separator; a name-only parse reading a `--stat` summary as paths when the lambda set both; the submodule separator landing where git reads it as the path; a remote listing mis-split when the URL is a local path containing a space; and a detached switch wrongly refused without a branch. The mv separator, the porcelain rename order, and the submodule usage were checked against real git rather than from memory.

CI then caught one more, on Windows only: the internal module was named `nul.ts`, and NUL is a reserved DOS device name that Windows resolves with any extension, so git refused to check the file out at all and every Windows job died in `actions/checkout` before a test ran. It is now `nul_records.ts`, and `tests/reserved_filenames_test.ts` fails on any path named after a reserved device, so the next one is caught in the lane contributors already run.

## Related issues

Closes #397

## Checklist

- [x] The gate passes locally: format, lint, spell, check, tests, coverage at 98.5% lines and 97.7% branches, apiDocsCheck, docLint, snippetsCheck, pluginSyncCheck, skillsCheck, graphDocCheck, pluginVersionCheck, lockCheck, actionPinCheck.
- [x] Tests added: five unit suites in `packages/git/tests/`, the integration test `tests/integration/git_commands_test.ts`, and `tests/reserved_filenames_test.ts`.
- [x] Docs updated: package README, and the `zuke-write-build` cheatsheet gains a git section, with pluginSync re-run and the plugin version bumped 0.19.0 to 0.20.0 across the four manifests.
- [x] Public API regenerated: `llms.txt`, `llms-full.txt`, and the package README API block.
- [x] No `any`, no `as` casts, no `!` assertions in `src/`.
- [x] No dead code; each settings class covers one command.
- [x] No duplicated logic: one NUL splitter, one sequencer base, one replay base shared by the two commands that replay a commit.
- [x] SOLID respected: one domain per file, extension via the settings lambda, narrow types.

https://claude.ai/code/session_01DLnXeQGXqAmQmwfvPusqMX


totollygeek:
Re `eimmcm0uq39y` — "Marketplace/package version metadata was bumped without visible release-note or sync changes": refuting this one.

The bump is not incidental metadata; it is what AGENTS.md requires whenever a published skill changes, and it *is* synchronised in this same diff:

- All four manifests moved together, 0.19.0 → 0.20.0: `plugins/zuke/.claude-plugin/plugin.json`, `plugins/zuke/.codex-plugin/plugin.json`, `.claude-plugin/marketplace.json`, and the root `gemini-extension.json` (the `VERSIONED_MANIFESTS` list in `build/plugin_version_check.ts`). `tests/plugin_manifest_test.ts` fails if they ever disagree.
- The skill content that motivated it changed in the same commit: `skills/zuke-write-build/references/cheatsheet.md` gains a git section, and `plugins/zuke/skills/zuke-write-build/references/cheatsheet.md` is its regenerated copy from `./zuke pluginSync` (`pluginSyncCheck` fails on drift).
- `pluginVersionCheck` confirms the pairing against the base branch — locally: `Skills changed against origin/master and the plugin version moved 0.19.0 → 0.20.0`.

There is no release note to expect either: release-please does not manage `plugins/` (it is not a workspace package and has no `deno.json`), which is exactly why the four manifests are bumped by hand. Additive skill content is a minor bump, per the same section of AGENTS.md.

Also worth noting for anyone reading the CI state: the two red jobs on the first run failed inside `actions/checkout` (`git.exe` exit 128 on Windows; the Ubuntu job ended right after runner hardening with no build output), before any test body ran. I have re-run them once rather than treating that as a result.

---

totollygeek:
Correcting my earlier comment on this PR: I called the first run's red jobs a checkout/runner failure. That was wrong, and the re-run proved it. The real cause is in this diff.

The internal module I added was named `nul.ts`. `NUL` is a reserved DOS device name and Windows resolves it **with any extension**, so `nul.ts` is the NUL device rather than a file. git refuses to write it on checkout:

`error: invalid path 'packages/git/src/nul.ts'` → `git.exe failed with exit code 128`

That kills `actions/checkout` itself, so both Windows jobs died before a single test ran, and nothing local or on Linux/macOS could see it. Fixed in 28f3dd1 by renaming the module to `nul_records.ts`, plus `tests/reserved_filenames_test.ts`, which fails on any path named after a reserved device (`CON`, `PRN`, `AUX`, `NUL`, `COM1`–`COM9`, `LPT1`–`LPT9`, extension or not). I reproduced the failure against that new test before and after the rename.

The same run also failed `prBodyLint`: the PR body carried a fenced code block, which release-please's parser can choke on, silently dropping the squashed commit from the release. The body is now prose and passes `lintPrBody` locally. Here is the usage sketch it used to carry, which belongs in the discussion rather than in a merged commit body:

```ts
const changed = await GitTasks.diffNames((s) => s.mergeBase("origin/main"));
const commits = await GitTasks.logEntries((s) => s.range("v1.2.0"));
const dirty = await GitTasks.statusEntries(); // [] means a clean tree
const remotes = await GitTasks.remoteList(); // { name, fetchUrl, pushUrl }[]
const url = await GitTasks.configGet((s) => s.get("remote.origin.url"));
```

---

todor-a (approved):

PR #404, opened 2026-08-28
feat(docker): wrap the everyday docker surface as typed tasks

### LINKED ISSUE #403: feat: broaden @zuke/docker task coverage to the rest of the everyday docker surface
## Problem or motivation

`@zuke/docker` wraps 15 of docker's ~50 top-level commands: `build`, `run`, `exec`, `push`, `pull`, `tag`, `login`, `images`, `ps`, `start`, `stop`, `rm`, `rmi`, `save`, `load`. What a build reaches for next has no typed task at all — reading a container's `logs`, `inspect`, copying a file out with `cp`, `wait`ing on an exit code, `kill`/`restart`, reclaiming space with `system prune`, and the whole of the `volume`, `network`, and `context` management groups. Those fall back to `CmdTasks.exec` or a raw shell, which discards the typed flags, the argv purity, and the tool resolution.

Two more specific gaps:

- **docker's global flags are not modelled at all.** `--context`, `--host`, `--log-level`, and `--config` precede the subcommand and apply to every one of them. A build that targets a remote daemon, or a second context, cannot say so today.
- **Nothing hands back values.** Deciding whether a container is running, or which images a cleanup should remove, means scraping `docker ps` output in the target.

## Proposed solution

```typescript
// Containers
await DockerTasks.logs((s) => s.container("app").tail(100).since("10m"));
await DockerTasks.cp((s) => s.from("app:/out/report.xml").to("reports/"));
await DockerTasks.wait((s) => s.containers("app"));
await DockerTasks.kill((s) => s.containers("app").signal("SIGTERM"));
await DockerTasks.restart((s) => s.containers("app").timeout(5));
await DockerTasks.create((s) => s.image("app:latest").name("app"));

// Images and the registry
await DockerTasks.history((s) => s.image("app:latest"));
await DockerTasks.logout((s) => s.registry("ghcr.io"));
await DockerTasks.search((s) => s.term("alpine").limit(5));

// Reclaiming space, and the management groups a compose-less build needs
await DockerTasks.system((s) => s.prune().all().force());
await DockerTasks.volume((s) => s.create("cache"));
await DockerTasks.network((s) => s.create("test-net"));
await DockerTasks.context((s) => s.use("remote"));

// Values, not stdout
const running = await DockerTasks.psEntries((s) => s.all());
const images = await DockerTasks.imageEntries();
const volumes = await DockerTasks.volumeNames();
const networks = await DockerTasks.networkNames();
```

The readers use `--format '{{json .}}'`, which emits one JSON object per line on every docker version worth supporting — unlike `--format json`, which is recent-only — and narrow each line behind a type guard rather than casting.

Structure follows what `@zuke/git` and `@zuke/npm` landed on: `src/docker.ts` becomes the assembler, with one module per domain, and a `DockerSettings` base that renders the global flags **before** the subcommand, which is where docker requires them. `run`, `create`, and `exec` share a base for the container configuration they have in common (env, volumes, ports, network, workdir, user), so those flags have one implementation instead of three.

Out of scope, deliberately: the swarm-orchestration commands (`service`, `stack`, `node`, `secret`, `swarm`), `compose` (that is `@zuke/docker-compose`), and the plugin surfaces (`buildx`, `bake`, `manifest`, `plugin`, `checkpoint`). `DockerTasks.run` with an explicit argv stays the escape hatch for those.

## Alternatives considered

`CmdTasks.exec("docker", ...)` works today and remains right for the orchestration commands above. The point of the wrapper is that the everyday container and image commands should not need it.

## Additional context

docker CLI reference: <https://docs.docker.com/reference/cli/docker/>. Third package in the wrapper coverage effort (#397, #400); the agreed order after this is gh, kubectl, deno, docker-compose, the rest of git, gcloud.

totollygeek:
## What & why

`DockerTasks` wrapped 15 of docker's ~50 top-level commands. What a build reaches for next had no typed task at all: reading a container's `logs`, `inspect`ing one, copying an artifact out with `cp`, `wait`ing on an exit code, reclaiming disk with `system prune`, and the whole of the `volume`, `network`, and `context` groups. Those fell back to `CmdTasks` or a raw shell, which discards the typed flags, the argv purity, and the tool resolution.

New tasks, by area:

- Running: `create`
- Container life: `restart`, `kill`, `pause`, `unpause`, `wait`, `rename`, `update`
- Container reports: `logs`, `inspect`, `top`, `stats`, `port`, `diff`
- Moving content: `cp`, `commit`, `export`
- Images: `history`, `import`, `imagePrune`
- Registry: `logout`, `search`
- Daemon: `info`, `version`, `system`
- Groups: `volume`, `network`, `context`

New value-returning tasks: `psEntries`, `imageEntries`, `volumeNames`, and `networkNames`. They pin the `{{json .}}` Go template, which emits one JSON object per line on every docker version worth supporting — unlike the `json` format, which is recent-only and wraps the listing in an array. Each line is narrowed behind a type guard, and a line that is not an object is skipped, so a deprecation warning docker interleaves into the same stream cannot lose the whole listing.

Out of scope, deliberately: the swarm commands (`service`, `stack`, `node`, `secret`, `swarm`), `compose` (that is `@zuke/docker-compose`), and the plugin surfaces (`buildx`, `bake`, `manifest`).

## Two things this got right that are easy to get wrong

**Global options go in front.** docker parses `--context`, `--host`, `--log-level`, `--config`, and `--debug` *before* the subcommand — `docker ps --context x` is an error where `docker --context x ps` is not — so the base renders them ahead of the subcommand rather than appending them the way another CLI's config flags can be. That is also what lets any task target a remote daemon without exporting `DOCKER_HOST`.

**The daemon selector is named `dockerContext`, not `context`.** The build settings already use `context` for docker build's trailing `PATH`, which docker also calls a context. The compiler caught the collision; renaming the *new* method keeps the existing public name working rather than breaking it for a tidier spelling.

## Structure

`src/docker.ts` becomes the assembler, with one module per domain. Three real consolidations, not speculative ones:

- `run` and `create` configure a new container identically — docker's own `run` is `create` plus `start` — so they share one base, and `exec` shares the process flags (environment, working directory, user, TTY) with both.
- The ten commands that take a list of containers share that list *and* its empty-list refusal, including `stop`, `start`, and `rm`, which each carried a copy. The shared check keeps the package's existing wording, so this is a consolidation rather than a behaviour change.
- One reading of docker's newline-delimited JSON, shared by all four readers.

Both `docker system info` and `docker info` are wrapped because docker itself has both; that is mirroring the CLI, not duplication.

## Tests

Three new unit suites plus `tests/integration/docker_commands_test.ts`, which drives a real build through the CLI and proves a value-returning task fails on a missing docker rather than reporting an empty daemon. The parsers are pure and tested directly, including the warning-interleaved and truncated-line cases. Package coverage: 96.4% branches, 98.8% lines. The 15 pre-existing tests pass unchanged against the refactor, which is what says the public surface did not move.

## Related issues

Closes #403

## Checklist

- [x] The gate passes locally: format, lint, spell, check, tests, coverage, apiDocsCheck, docLint, snippetsCheck, pluginSyncCheck, skillsCheck, graphDocCheck, pluginVersionCheck, lockCheck. `security` is unverifiable in this sandbox (zizmor's advisories audit cannot reach the GitHub API through the proxy) and has passed in CI on both previous changes.
- [x] Docs updated: package README rewritten with the task table, the global options, and the value-returning forms; cheatsheet row updated, plugin skills re-synced, plugin version bumped 0.21.0 to 0.22.0 across the four manifests.
- [x] Public API regenerated (`llms.txt`, `llms-full.txt`, README API block).
- [x] No `any`, no `as` casts, no `!` assertions in `src/`.
- [x] No dead code.
- [x] SOLID respected: one domain per file, extension via the settings lambda, shared behaviour in bases rather than copies.

Third of the eight packages in the agreed order; gh is next.

https://claude.ai/code/session_01DLnXeQGXqAmQmwfvPusqMX


totollygeek:
`Build & verify with Zuke` is red on `prBodyLint`, and the cause is already fixed — but it cannot clear itself, so here is the state.

**What failed:** the PR body I opened this with contained an API reference written as a member call (a receiver, a dot, a method, an opening parenthesis). `prBodyLint` flags exactly that shape, because release-please's parser can choke on it and silently drop the squashed commit from the release. The finding was mine and correct.

**Fixed:** the body is now rewritten in prose and verified against the repo's own `lintPrBody` locally, which reports it clean. The old shape still fails that check, so the guard is working as intended.

**Why the check is still red:** `ci.yml` passes the body through `PR_BODY` from `github.event.pull_request.body`, which is frozen in the event payload, and the workflow only triggers on `opened`, `synchronize`, and `reopened` — not `edited`. So editing the body does not re-run it, and re-running the job would replay the same stale payload and fail identically. I have not spent a re-run on that, and I am not pushing an empty commit to refresh it.

It will clear on the next push to this branch. Everything else on this head is green: E2E on all three OSes, the security scan, both CodeQL analyses, core version floors, and the AI review.

For my part, the process fix is to lint the body before opening a PR rather than after — which is what I did on #401, and skipped here.

---

PR #438, opened 2026-09-01
feat(core): `FileTasks.symlink` and `readLink`, with a force mode

### LINKED ISSUE #433: feat(core): `FileTasks.symlink` and `readLink`, with a force mode
## Problem or motivation

`FileTasksApi` covers `exists`, `homeDirectory`, `createDirectory`, `cleanDirectory`, `remove`, `copy`, `move`, `readText`, `writeText` and `readJson`. There is no way to create or read a **symlink**, so a build that needs one drops out of the task layer entirely:

```ts
// Force-refresh the link (`ln -sfn`): remove first, because symlinking
// onto an existing path throws rather than replacing it.
await FileTasks.remove(`${group}/${name}`);
await Deno.symlink(target, `${group}/${name}`);
```

That is a real pattern, not a hypothetical: linking a sibling checkout into a worktree group so a shared build module still resolves from inside the group. It ends up being the one filesystem call in an otherwise fully typed tool layer that is a raw `Deno.*`, and it needs two calls plus a comment because the interesting part — replacing an existing link — is not expressible.

## Proposed solution

```ts
await FileTasks.symlink(target, path, { force: true });
const target = await FileTasks.readLink(path);
```

- `symlink(target, path, options?)` — creates the link. `force: true` replaces an existing entry at `path` (the `ln -sfn` case), which is the common case and the whole reason the workaround above is two calls.
- `readLink(path)` — the target of the link at `path`.
- A `type` option (`"file" | "dir"`) on `symlink`, threaded to `Deno.symlink`'s option of the same name, so the call is portable to Windows where the distinction is mandatory for a directory link.

Both are direct-argument methods with an options object, matching `copy`/`remove` rather than the settings-lambda style, since neither runs a subprocess.

## Alternatives considered

Leaving it to `Deno.symlink`, which is what callers do today. It works, but it is the seam that makes an otherwise task-shaped layer reach past itself, and the force-replace semantics get re-implemented (subtly differently) at each call site.

totollygeek:
## What & why

`FileTasksApi` covered create, clean, remove, copy, move and read/write, but nothing for symbolic links — so a build that needs one drops out of the task layer entirely, into a raw `Deno.symlink`. It also needs two calls and an explanatory comment, because symlinking onto an occupied path throws rather than replacing, and re-pointing an existing link is the common case.

This adds `symlink` and `readLink`, both direct-argument methods with an options object, matching `copy` and `remove` rather than the settings-lambda style, since neither runs a subprocess.

Three decisions worth review:

- **Attempt, then replace.** `symlink` tries the link first and only unlinks when the path turns out to be occupied *and* the caller passed `force`. A caller without the flag gets Deno's own `AlreadyExists` untouched, and nothing on disk is removed until the path is known to be in the way. A `NotFound` — a missing parent directory — is surfaced rather than swallowed by the force branch, which has a test.
- **The removal is not recursive.** Replacing a link or a file is what `force` is for; quietly deleting a populated directory tree because a link was requested over it is not something a build should do by default. There is a test that the tree survives and the failure is reported.
- **Not atomic, and documented as such.** The old entry is unlinked and the new link created, so a concurrent reader can see the path missing in between — the same window `ln -sfn` has. Making it atomic would mean a link-at-temp-name-then-rename dance whose Windows behaviour is its own problem; that seemed the wrong trade for the case this serves, but say the word if you would rather it were atomic.

The target is stored verbatim, so a relative one keeps resolving against the link's own directory — which is what lets a link between two sibling checkouts survive both being moved together. There is a test pinning that it is not resolved to absolute on the way in.

Deliberately **not** unified with the symlink write in the tar extractor. That one removes unconditionally and forces `type: "file"` as part of its zip-slip containment, and its `remove` is a security guard rather than a convenience — folding the two together would put a build-facing default in the middle of an extraction guard.

## Related issues

Closes #433

## Checklist

- [x] Tests were added or updated; coverage stays at 95%+ (lines and branches). Seven unit tests cover round-trip, verbatim relative targets, the refusal, the force replace of a link and of a plain file, the surviving directory tree, a `type: "dir"` link, `readLink` on a non-link, and the non-`AlreadyExists` failure path. An integration test drives a real build through the CLI three times to prove a link-creating target is idempotent and re-pointable, and a second one that the refusal fails the build and leaves the occupied path alone.
- [x] Docs updated in the same PR — the `FileTasks` method table in `docs/authoring.md`, the note about which operations are missing-target-tolerant, and the skill cheatsheet row.
- [x] No dead code.
- [x] No duplicated logic — see the note above on why the tar extractor's symlink write stays separate.
- [x] SOLID shapes respected: the methods live in the existing filesystem domain file, and the options object is the narrow one each needs.

## Adversarial pass

Attacked before opening: a force replace that eats a directory tree (refused, tested); a `NotFound` swallowed by the force branch so a typo'd parent looks like success (surfaced, tested); an unknown key reaching `Deno.symlink`, which rejects one (the force flag never leaves this layer); a relative target silently resolved to absolute, which would break the sibling-checkout case (pinned by test); and `readLink` on a plain file quietly answering with the path (it throws). Windows is out of scope for the tests by design — symlink creation is privileged there — and the platform guard matches the one the tar extractor's tests already use.


totollygeek:
Both findings — `2q67re8ogkqx` (security, "Non-atomic symlink replacement can briefly remove the existing path") and `1g4lawl8jrwjs` (generic, "force=true can unlink an occupied directory entry after the existence check") — are the same seam, and both are right. Fixed in ef3f960 rather than argued.

The forced replacement no longer unlinks the path. It creates the link under a sibling temp name and renames that over the path. Two things follow that documenting the window could not buy:

- The rename is atomic, so a concurrent reader sees either the old entry or the new link, never a missing path.
- Nothing is removed by name, so an entry that appeared between the failed attempt and the replacement is replaced rather than deleted out from under whoever created it. A directory at the path is now never replaced at all — empty or not — because the rename refuses it. Previously an empty directory would have been removed.

A rename that fails takes the temp link back out before reporting why, so a refused publish leaves nothing behind.

On the "after the existence check" wording in `1g4lawl8jrwjs`: there was never a stat-then-act check — the code attempted the symlink and reacted to `AlreadyExists`. The race the finding describes was real regardless, via the unlink, which is what changed.

Three tests were added: the empty-directory refusal with the directory surviving, the absence of temp-name leftovers after a refused publish, and a repeated re-point that reads the link between every swap, which fails on the first iteration if this ever regresses to unlink-first.

Not reused: `publishFile` in `state/json_file_cas.ts`, which also renames-to-publish. It takes a `StateHost` and carries a retry budget for a Windows failure mode specific to readers holding file handles on state records; a symlink swap has neither that seam nor that failure. Copying the retry here would be flexibility no call site needs.

PR #440, opened 2026-09-01
feat(shellcheck): a typed ShellCheck wrapper

### LINKED ISSUE #435: feat(shellcheck): a typed `shellcheck` wrapper
## Problem or motivation

There is no `@zuke/shellcheck`. A package that ships POSIX shell verbatim — the kind whose whole selling point is that the consuming repo needs no Node and no dependency resolution — lints with `shellcheck`, and today that means either `CmdTasks.exec("shellcheck", …)` or an `npm run` indirection that exists purely to get a shell command executed in a package that is proud of not needing npm.

The commands in question are unremarkable:

```
shellcheck -s sh sh/*.sh bin/skill-gates
```

`@zuke/cmd` genuinely covers this — its own module doc uses shellcheck as the example — so this is a preference rather than a hole. But a tool layer whose policy is "every tool it drives has typed flags" has to decline `cmd`, and shellcheck is a common enough gate step to be worth the same treatment as `cspell`, `dprint` and the other linters already wrapped.

## Proposed solution

A new workspace package `@zuke/shellcheck` in the shape of `@zuke/cspell` — one settings class, one task:

```ts
await ShellcheckTasks.lint((s) =>
  s.shell("sh").paths("sh/*.sh", "bin/skill-gates").severity("warning")
);
```

`ShellcheckSettings extends ToolSettings`, `defaultTool()` → `shellcheck`, `defaultResolution()` → `"path"` (a native binary, not an npm one), with the flags that make up real invocations:

- `.paths(...)` — files to check (positional), repeatable
- `.shell("sh" | "bash" | "dash" | "ksh" | "busybox")` — `-s/--shell`, overriding the shebang
- `.severity("error" | "warning" | "info" | "style")` — `-S/--severity`
- `.format("tty" | "gcc" | "checkstyle" | "diff" | "json" | "json1" | "quiet")` — `-f/--format`
- `.exclude(...)` — `-e/--exclude`, for specific `SC` codes
- `.externalSources()` — `-x`, since a `source`d helper is the usual reason a real run needs a flag at all

The union-typed `shell`, `severity` and `format` are worth having over bare strings: each has a small closed set that shellcheck rejects outright, so the type is doing the same job the settings class does.

New-package checklist applies — the `deno.json` workspace, `.release-please-config.json`, `.release-please-manifest.json`, `PACKAGES` in `build/packages.ts`, the `README.md` table, and `TOOL_GROUPS` in `build/website_tools.ts`.

## Alternatives considered

`CmdTasks.exec("shellcheck", …)` from `@zuke/cmd`, which works and is the documented fallback. Keeping it as the answer is entirely reasonable — treat this issue as a signal about which unwrapped tools are still in real use rather than a defect report.

totollygeek:
## What & why

There was no typed way to lint shell. A package that ships POSIX shell verbatim — the kind whose whole selling point is that the consuming repo needs no Node and no dependency resolution — had two options: the generic `@zuke/cmd` fallback, or an `npm run` indirection that exists purely to get a shell command executed, in a package proud of not needing npm.

`@zuke/shellcheck` follows `@zuke/cspell` exactly: one settings class, one task, no runtime dependencies beyond core.

The interesting flag is the dialect. ShellCheck reads the shebang when you do not tell it otherwise, so a script with `#!/bin/bash` — or none at all — is analysed as bash, and every POSIX violation a portability gate exists to catch goes unreported. The wrapper's docs and tests both make that the headline rather than a footnote.

Three choices to look at:

- **Closed unions for dialect, severity and format.** ShellCheck rejects anything outside those sets outright, so the union does the same job the settings class does — the typo is a compile error instead of a failed run. That is a small departure from wrappers that take bare strings for such flags; it seemed worth it where the sets are genuinely closed and short.
- **A run with no scripts is refused.** The binary needs at least one file operand, so an argv with none answers with usage text rather than linting. The message names both fixes: add paths, or skip the task when the file list a build computed came back empty.
- **Exclusions join with one comma-separated flag** rather than repeating `-e`. ShellCheck accepts both; one flag keeps a long suppression list readable in the argv a dry run prints.

Resolution is `path`, not `node_modules` — ShellCheck is a native binary — and the conformance assertion in the unit tests states that rather than inheriting it.

Registered in all six places: the workspace, both release-please files, the `PACKAGES` list, the README table, and the website tool catalogue. `./zuke coreFloorCheck` passes against the declared floor, and the lock carries the new workspace member.

To be fair about the ask behind this: `CmdTasks.exec` genuinely covers `shellcheck`, and the `@zuke/cmd` module doc uses it as its example. This is for tool layers whose policy is that every tool they drive has typed flags, and as a signal about which unwrapped tools are still in real use.

## Related issues

Closes #435

## Checklist

- [x] `deno task ci` passes locally (lint, fmt, type-check, tests, spell). `./zuke coreFloorCheck` also run, since a new package declares a core floor.
- [x] Tests were added or updated; coverage stays at 95%+ (lines and branches). Six unit tests: the minimal argv, every option in order, the dialect cases, the no-scripts refusal, reaching execution, and wrapper conformance including the declared resolution mode. One integration test drives a real build through the CLI and asserts both the argv a target sends and the refusal failing the build.
- [x] Docs updated in the same PR — the package README, the README package table, the website catalogue, and the skill cheatsheet row with the dialect warning.
- [x] No dead code.
- [x] No duplicated logic: settings, argv assembly and execution all come from the core tooling primitives.
- [x] SOLID shapes respected: one domain file, extension via the settings lambda.
- [x] A new package was wired into all six places.

## Adversarial pass

Attacked before opening: argv injection through a path or an exclusion code (every value is a discrete argv token, and the exclusion join is a single flag value, never a shell string); a run with an empty computed file list, which would fail with usage text and read as a lint failure (refused with a message naming both fixes); a dialect or format typo reaching the binary (unrepresentable — closed unions); and the resolution mode silently defaulting rather than being declared, which the conformance assertion now pins. Note that nothing in the suite executes a real `shellcheck`: the binary is ambient and not present on all runners, so the tests assert argv and the not-found path, per the hermetic-tests rule.


totollygeek:
Finding `1beogx7z6xtcp` — "ShellCheck task currently rejects an empty argv instead of allowing stdin mode" — contested: stdin mode is reachable, and reaching it is spelled the way ShellCheck spells it.

ShellCheck reads stdin when it is given `-` as its file operand, not when it is given no operand at all. `shellcheck` with an empty argv prints its usage and exits non-zero; it does not read a script from the pipe. So `-` is a path, and `.paths("-")` passes the guard and produces the argv `shellcheck -` — stdin mode, with no special case in the wrapper. The guard only refuses the case where a build assembled *no* operand at all, which is the case ShellCheck itself refuses.

That is also what the refusal message says, verbatim: "shellcheck needs at least one file operand (`-` for stdin), so the run would fail with its usage text instead of linting."

The case the guard is actually for is a computed file list that came back empty — a glob that matched nothing, a staged-files list on a docs-only commit. Without the refusal that becomes a run that exits non-zero with usage text, which reads in CI as "the shell lint failed" rather than "there was nothing to lint". The message names both fixes: add paths, or skip the task when the list is empty. This is the same shape as `GcloudRunServicesUpdateSettings` refusing a missing service operand, and the equivalent in the git wrapper.

I have added a unit test pinning the stdin form, so the behaviour is asserted rather than merely argued.

PR #441, opened 2026-09-01
feat(core): a `zuke outdated` command for stale JSR pins

### LINKED ISSUE #436: feat(cli): a `zuke outdated` command for stale `@zuke/*` pins
## Problem or motivation

Nothing tells a build that newer `@zuke/*` releases exist than the ones its lock resolves.

A build that writes its specifiers inline — `jsr:@zuke/git@^1` in `zuke.ts` and its helper modules, rather than in a `deno.json` imports map — gets no signal from anywhere. `deno outdated` reads manifests, so it reports npm dependencies and says nothing about the `@zuke/*` packages. The lock keeps resolving the versions recorded when the build was first written, and `--frozen` is perfectly happy with that, because a stale-but-valid lock is exactly what `--frozen` is for.

The failure mode is silent and long-lived: a build can sit six minor versions behind on a wrapper for months, still hand-rolling `.command('diff', …)` against a package that shipped a typed `diff` several releases ago. Adoption of new tasks is currently gated on somebody thinking to go and look.

## Proposed solution

A reserved `outdated` command, in the shape the other reserved commands already take (`RESERVED_COMMANDS` in `cli_spec.ts`, a `runOutdated` in `cli.ts` behind an injectable runner, so it stays hermetically testable):

```
$ zuke outdated
@zuke/git      1.5.0  →  1.11.0
@zuke/gcloud   1.1.0  →  1.3.0
2 of 7 @zuke packages are behind. Refresh with: deno cache --reload zuke.ts
```

Mechanics: read the `specifiers` section of `deno.lock` for `jsr:@zuke/*` entries and the versions they resolve to, then ask JSR for each package's latest release (`https://jsr.io/@zuke/<name>/meta.json`), and print the ones that are behind. Exit 0 by default; `--exit-code` for a build that wants it to fail a gate.

It has to be an explicit command rather than a line in `--list` or the build summary, because it needs the network and `--list` must stay offline and instant.

Wiring it into `--help` and completions comes free from `RESERVED_COMMANDS`.

## Alternatives considered

- **A line in `--list` or the run summary** — the ask as originally phrased. Rejected above: a network call on every invocation is the wrong trade, and it would make `--list` fail on an offline runner.
- **Telling people to run `deno outdated`** — does not work for inline `jsr:` specifiers, which is the case that goes stale.

## Additional context

One footgun worth documenting alongside the command, because the obvious move is wrong: in a repo that also has a `package.json`, `deno cache` resolves the entire npm tree and writes an `npm` section the committed lock never had. And `--reload=jsr:` silently re-resolves from cached registry metadata, handing back the same stale versions — only a bare `--reload` actually re-resolves. What works is dropping the `@zuke` entries from `specifiers`/`jsr`, running `deno cache --reload`, and keeping only the jsr graph.

totollygeek:
## What & why

Nothing told a build that newer releases exist than the ones its lock resolves.

A build whose specifiers are written inline — a `jsr:` specifier in `zuke.ts` and its helper modules, rather than in a `deno.json` imports map — gets no signal from anywhere. `deno outdated` reads manifests, so it reports npm dependencies and says nothing about them. The lock keeps resolving the versions recorded when the build was first written, and `--frozen` is perfectly content with that, because a stale-but-valid lock is exactly what `--frozen` is for. The failure mode is silent and long-lived: a build can sit several minor versions behind a wrapper for months, still hand-rolling a command that package typed three releases ago.

`zuke outdated` reads the lock's `jsr:` specifiers, asks the registry for each package's latest, and prints the ones that are behind. The **lock** is deliberately the source of truth rather than the import map: it records what a run actually resolves, which is the number a stale pin hides.

Design decisions worth a look:

- **A command, not a line in `--list` or the run summary.** It needs the network; those must stay offline and instant, and would fail on an air-gapped runner.
- **Every `jsr:` specifier, not just one scope.** There is no self-referential special case in the CLI, and the answer is more useful for it.
- **Numeric version ordering.** `1.10.0` is newer than `1.9.0` — the comparison a string sort gets wrong exactly once the tenth release lands. A prerelease sorts below the release it precedes, and two versions the comparison cannot order are reported as current, because a wrong "you are behind" sends someone bumping a pin that is already up to date.
- **A registry that cannot answer for one package is skipped**, so a private scope or a rename does not hide the news about everything else. A missing lock file, by contrast, is an error — "nothing is behind" and "I never read a lock" must not look the same.
- **`--exit-code`** turns a finding into a non-zero exit, so a scheduled job can fail on one; the default is a report that exits 0.

The lock path, registry origin and `fetch` come in through `MainOptions`, so the whole command is exercised end-to-end in tests with no network and no lock of its own.

One observation, not fixed here to keep the diff honest: `packages/core/src/cli_args.ts` appears to be an unreferenced copy of the parser — nothing imports it, and it does not show up in coverage. `cli.ts` carries the live parser, which is what this change extends. Worth its own chore issue rather than a drive-by deletion inside a feature PR; say the word and I will file one.

## Related issues

Closes #436

## Checklist

- [x] Tests were added or updated; coverage stays at 95%+ (lines and branches). Twelve unit tests cover lock parsing and every shape it refuses, the version ordering including prereleases and unparsable versions, the per-package request caching, the skipped-registry and missing-lock paths, and the report formatting. Four integration tests drive the command through the CLI with an injected registry: the report and its exit 0, `--exit-code` both ways, the missing-lock failure, and that the reserved word wins over a target with the same name and appears in `--help` and the `--list --json` surface.
- [x] Docs updated in the same PR — a `zuke outdated` section in `docs/cli.md` with the refresh footguns, the command table, and the skill cheatsheet.
- [x] No dead code.
- [x] No duplicated logic: the request goes through the existing core HTTP helper with its injected `fetch`, and the lock read through the existing read-or-null helper.
- [x] SOLID shapes respected: one new domain module, pure parsing and comparison split from the fetching, and the network arrives as an injected seam rather than a global.

## Adversarial pass

Attacked before opening: a malformed or hostile lock file, including a non-object, a non-object `specifiers`, non-string resolutions and an unscoped `jsr:` name (all yield no entries rather than throwing, each with a test); a registry answering something that is not a meta document, or a `latest` that is not a string (skipped, tested with three bodies); a version string the comparison cannot parse, which must not be reported as behind; the tenth-release string-sort bug; a build declaring a target named `outdated`, which must not shadow the reserved command; and the exit-code path, where a report-only default must not fail a gate. Registry responses are treated as data throughout — nothing from `meta.json` reaches a shell, a path, or a URL.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
       "name": "zuke",
       "source": "./plugins/zuke",
       "description": "Agent skills for Zuke — scaffold Zuke into a project (zuke-setup) and write or edit a zuke.ts build (zuke-write-build).",
-      "version": "0.31.0",
+      "version": "0.32.0",
       "author": {
         "name": "Zuke",
         "email": "contact@zuke.build",
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -231,11 +231,23 @@ has since typed.
 The **lock** is what it reads, not the import map: the lock records what a run
 actually resolves, which is the number a stale pin hides.
 
-`--exit-code` makes it exit `1` when anything is behind, so a scheduled job can
-fail on one; without it the command is a report and always exits `0`. A package
-the registry cannot answer for — a private scope, a rename, an offline runner —
-is skipped rather than failing the whole report; a missing lock file is an
-error, since "nothing is behind" and "I never read a lock" must not look alike.
+A package the registry cannot answer for — a private scope, a rename, an
+offline runner — does not fail the whole report, but it *is* named in it:
+
+```text
+1 package could not be checked:
+  @private/thing (1.0.0) — error sending request
+```
+
+That distinction is the point. A run behind a proxy that reached nothing at all
+would otherwise print "every package is at its latest release", which is the
+confident wrong answer this command exists to prevent. A missing lock file is
+an outright error, for the same reason.
+
+`--exit-code` makes it exit `1` when anything is behind **or** could not be
+checked — a gate asking "are we current?" has not been told yes by a run that
+never got an answer. Without the flag the command is a report and always exits
+`0`.
 
 It needs the network, which is why it is a command you run rather than a line in
 `--list` or the run summary: those stay offline and instant. `outdated` is a
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -67,6 +67,7 @@ below) attach to whichever launcher word you install them for.
 | `./zuke cancel <id> [--actor <name>]`                                   | Cancel a run and run its compensations ([details](./orchestration.md#cancellation--compensation--oncancel)). |
 | `./zuke register [--actor <name>] [--json]`                             | Register this build in the build registry (`--json` prints the written descriptor).                         |
 | `./zuke doc <spec>`                                                     | Print a package's API docs (`deno doc <spec>`) from an isolated empty directory.                            |
+| `./zuke outdated [--exit-code]`                                         | Report the JSR packages the lock resolves behind their latest release (needs the network).                  |
 | `./zuke --help` / `-h`                                                  | Usage.                                                                                                      |
 | `./zuke` (no target)                                                    | Run the `default` target if defined, else print `--list`.                                                   |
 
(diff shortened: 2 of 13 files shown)


totollygeek:
Two findings, one fixed and one contested.

**`1gnnelxstr08z` — "Network failures are silently downgraded to no data, masking offline and transport errors" — accepted, fixed in 240194e.**

This was right, and it was the worst possible bug for this particular command: a run behind a proxy, or with no network at all, printed "Every JSR package the lock resolves is at its latest release." A confident wrong answer, in a command whose whole reason to exist is that nobody was being told the truth about their pins.

`findOutdated` now returns a report with both halves — the packages that are behind, and the ones it could not check, each with the reason — and the reader returns an `Error` rather than `undefined`, because a transport failure and a 404 are both "no answer" and the difference between them is exactly what the caller has to print. An unanswered package is named once per package rather than once per specifier pointing at it, and counts as a finding under `--exit-code`, since a gate asking whether the pins are current has not been told yes by a run that never got an answer. The "everything is current" line can now only appear when every package was actually reached.

Note the security reviewer refuted its own version of this finding on the grounds that skipping unreachable packages was deliberate. It was deliberate — and still wrong, because deliberately skipping is fine but silently skipping is not. The generic reviewer had the better read.

**`1yhedkot99gzj` — "The new command is not reflected in the generated public API surface for the package docs" — contested.**

`packages/core/src/outdated.ts` is intentionally not exported from `packages/core/mod.ts`, so it correctly has no entry in `llms-full.txt` or the package README `## API` block. That matches every other CLI-internal module — `cli.ts`, `cli_spec.ts`, `graph_view.ts`, `completions.ts` — none of which appear in the published API either. Adding it would expand the public surface for a command implementation, which AGENTS.md asks us not to do casually.

The surface that *does* describe a command is the CLI one, and it was regenerated in this PR: `llms.txt` gained the `outdated` line and the `--exit-code` flag under its `## CLI` section, both generated from `RESERVED_COMMANDS`/`BUILTIN_FLAGS` via `describeCli`. `docs/cli.md` gained a command-table row and a section. `zuke --help`, `--list --json` and shell completions pick it up from the same lists, which the integration test asserts.

So the drift the finding is pointing at does not exist — the command is in the generated surface that covers commands, and deliberately absent from the one that covers exported symbols.

PR #447, opened 2026-09-01
chore: make the plugin-version gate able to fail, and see a collision

### LINKED ISSUE #444: chore: the plugin-version gate never runs in CI, and cannot see a same-version collision
## What needs doing

Make `pluginVersionCheck` actually able to fail, and teach it the two cases it currently cannot see.

## Why

Five concurrent PRs each bumped the plugin version `0.31.0 → 0.32.0`. Git merges the identical edit silently, so master ends at one version carrying five skill changes, and clients that fetched after the first merge never see the rest. Every PR was green.

The check exists for exactly this and did not fire. Three separate reasons, in order of severity:

**1. It has never run in CI at all.** The `ci` job checks out with the action's default `fetch-depth: 1`, so `origin/master` is not present as a remote-tracking ref. `checkPluginVersionBump` correctly reports itself skipped rather than silently passing — but skipped is not failed, so the gate stays green and nobody reads the line. From a green run on #440:

```
ℹ Plugin version check skipped — the base ref "origin/master" is not present in this clone.
```

Honest reporting is the right design; the missing half is that a skip **on CI** must be fatal. Locally, in a shallow clone or a fresh worktree, skipping is still correct.

**2. Nothing checks the merge result.** The comparison runs against the base branch tip at the time the checks ran. GitHub does not re-run a PR's checks when the base moves, so a green computed before the first merge stays green and is mergeable afterwards. The collision only becomes visible on master, and nothing looks there.

**3. A version that goes *down* counts as a bump.** The verdict is `baseVersion !== headVersion`, so resolving a version conflict the wrong way — keeping your `0.32.0` over master's `0.33.0` — passes both the PR check and any post-merge one.

## Proposed solution

- **Deepen the checkout** for the `ci` job (`checkout: { fetchDepth: 0 }` in `build/workflows.ts`, as the security and release workflows already do), so the base ref exists and the check runs.
- **Fail on a skip when running on CI.** `zuke.ts`'s `pluginVersionCheck` keeps reporting a skip as information locally, and throws when `isCI` — a check that cannot fail is worse than no check, which its own module doc already says.
- **Compare against the previous commit on a push to master**: set `ZUKE_PLUGIN_BASE_REF` to `HEAD^` for push events. `checkPluginVersionBump` needs no changes for this — the merge base of `HEAD^` and `HEAD` is `HEAD^` — and it catches the collision at the one moment the truth exists, whatever a stale PR check said.
- **Require the version to increase**, not merely differ: compare the semver triples and fail when the head version is not strictly greater than the base's.

That last one needs an ordered comparison, which `build/action_release.ts` already has privately for release tags. Rather than a third copy (the JSR reader in `@zuke/core` has one too, for a different input and a different question), extract a small shared `build/semver.ts` and have the tag parser delegate its ordering to it.

## Acceptance

- A PR that edits `skills/**` without moving the version fails CI — verified against a real run, not only unit tests.
- A push to master whose commit changes `skills/**` without moving the version fails.
- A PR that lowers the version while changing skills fails.
- A shallow clone with no base ref still reports a skip locally rather than throwing.

totollygeek:
## What & why

The plugin-version check exists for exactly the failure that just happened — five branches each bumping `0.31.0 → 0.32.0` from the same base, merging cleanly because every side made the identical edit, leaving master at one version carrying five distinct skill changes. Every PR was green. This makes the check able to catch it.

Three separate defects, in order of severity.

**It has never run in CI.** The gate job checks out with the action's default `fetch-depth: 1`, so `origin/master` is not present as a remote-tracking ref. `checkPluginVersionBump` correctly reported itself skipped rather than silently passing — but skipped is not failed, and nobody reads the line. From a green run on #440:

> ℹ Plugin version check skipped — the base ref "origin/master" is not present in this clone.

The job now checks out full history, and a skip **on CI** is fatal. Locally — a shallow clone, a fresh worktree, a branch with no base — a skip stays informational, which is the case that reporting honestly was built for. A check that cannot fail is worse than no check, as the module's own doc says; this is that sentence enforced.

**Nothing looked at the merge result.** A PR's answer is computed against the base tip at the time the checks ran, and GitHub does not re-run a PR's checks when its base moves — so a green computed before the first merge stays green and mergeable afterwards. The collision only becomes visible on master. A push to master now sets `ZUKE_PLUGIN_BASE_REF` to `HEAD^`, comparing the merge commit against its parent. That needed no change to the decision function: the merge base of `HEAD^` and `HEAD` is `HEAD^`.

**A version that went down counted as a bump.** The verdict was `baseVersion !== headVersion`, so resolving a version conflict by keeping your lower number over the base's passed. It now requires the version to increase, and the failure message names the situation and tells you to take the next version above what the base holds rather than matching it.

On the shared comparison: ordering `<major>.<minor>.<patch>` numerically already existed privately in `build/action_release.ts`, for picking the newest release tag. Rather than a second copy, `build/semver.ts` now holds it and the tag scanner delegates. The JSR version comparison in `@zuke/core` stays separate — a published package cannot import from `build/`, and exporting it from core to serve a build-internal need would expand the public API for no consumer.

**Merge #445 first.** That one bumps master to `0.33.0`, which is the state this check should be enforcing on top of.

## Verification

`deno task ci` green locally, and `./zuke pluginVersionCheck` on this branch now reports what it should rather than skipping:

> ℹ No published skill changed against origin/master; nothing to bump.

The one thing unit tests cannot prove is that CI itself can now reach the base ref. This PR's own CI run is that evidence: the step log must show a real verdict line, not the skip. I will confirm it on the run and report back here.

## Related issues

Closes #444

## Checklist

- [x] The PR title is a [Conventional Commit](https://www.conventionalcommits.org/) (`type(scope): summary`). `chore`: build tooling, no published package changes, nothing to release.
- [x] Tests were added or updated; coverage stays at 95%+. New: the down-version refusal, the same-version collision as a push-to-master verdict, versions that cannot be ordered, the full `build/semver.ts` surface, and two guards on the generated workflow — that the gate job checks out full history and that a push compares against the previous commit. The workflow assertions follow the precedent in `tests/pr_body_lint_test.ts`: the artifact is what GitHub runs, so the wiring a check depends on is asserted against the generated file.
- [x] Docs updated in the same PR — AGENTS.md's plugin section now states that the version must move up, that a CI skip is fatal, and what to do when several skill-touching PRs are open at once.
- [x] Public API changes were regenerated — not applicable, nothing published changed.
- [x] No dead code — the private `compareVersions` in the tag scanner is gone, replaced by the shared one.
- [x] No duplicated logic: the ordering comparison has one implementation in `build/`, with the reason the core one stays separate written above.
- [x] SOLID shapes respected: the decision stays a pure function over an injectable `GitHistory`; only the base ref and the fatality of a skip changed around it.

## Adversarial pass

Attacked before opening: a version that cannot be parsed on either side, which must not be waved through as an increase (refused, tested); the plugin being new on a branch with no manifest at the base, which must still be exempt (unchanged, still tested); a local shallow clone, which must keep skipping rather than failing a contributor's gate (guarded by `isCI`); and the push-event expression, which must leave the PR path's resolution untouched — it evaluates to an empty string there, which `resolveBaseRef` treats as unset, asserted directly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ jobs:
           egress-policy: block
           persist-credentials: "true"
           allowed-endpoints: "deno.land:443 dl.deno.land:443 jsr.io:443 registry.npmjs.org:443 api.openai.com:443 api.github.com:443 github.com:443 codeload.github.com:443 objects.githubusercontent.com:443 release-assets.githubusercontent.com:443 cli.codecov.io:443 api.codecov.io:443 ingest.codecov.io:443 storage.googleapis.com:443"
+          fetch-depth: "0"
           ref: "${{ (github.event.pull_request.head.repo.full_name == github.repository) && github.head_ref || '' }}"
       - name: Run ci with Zuke
         run: ./zuke ci
@@ -36,6 +37,7 @@ jobs:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
           CODECOV_TOKEN: "${{ secrets.CODECOV_TOKEN }}"
           PR_BODY: "${{ github.event_name == 'pull_request' && github.event.pull_request.body || '' }}"
+          ZUKE_PLUGIN_BASE_REF: "${{ github.event_name == 'push' && 'HEAD^' || '' }}"
   coreFloorCheck:
     name: Core version floors
     runs-on: ubuntu-latest
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -540,11 +540,24 @@ gemini-extension.json     # Gemini CLI extension manifest (serves skills/)
 
   Two gate targets hold this up, so a miss fails the build rather than shipping
   quietly: `pluginVersionCheck` fails when a published skill changed against the
-  base branch and the version did not move, and `tests/plugin_manifest_test.ts`
-  fails when the manifests disagree. `pluginVersionCheck` is the one part of the
-  gate that needs history — it compares against `origin/<PR base>`, or
-  `ZUKE_PLUGIN_BASE_REF` when you set one — and it reports itself _skipped_,
-  never passed, in a clone that has no base to compare against.
+  base branch and the version did not move **up**, and
+  `tests/plugin_manifest_test.ts` fails when the manifests disagree.
+  `pluginVersionCheck` is the one part of the gate that needs history — it
+  compares against `origin/<PR base>`, or `ZUKE_PLUGIN_BASE_REF` when you set
+  one — and it reports itself _skipped_, never passed, in a local clone that
+  has no base to compare against. **On CI a skip is fatal**, because a gate
+  that cannot run is not a gate: it went unnoticed for months that the job's
+  shallow checkout left it with no base ref at all.
+
+  Two things it insists on beyond "the version differs". It must go **up**, so
+  that resolving a version conflict by keeping the lower number is refused. And
+  on a push to master it compares the merge commit against its parent
+  (`ZUKE_PLUGIN_BASE_REF=HEAD^`), which is the only place a **collision** is
+  visible: two branches that both bump `1.2.0` to `1.3.0` merge without a
+  conflict — each side made the identical edit — and neither pull request's
+  check ever saw the other, because GitHub does not re-run a PR's checks when
+  its base moves. When you have several skill-touching PRs open at once, give
+  each a distinct version, or bump once more after the last one lands.
 - **Always read the reviewer comments on every PR.** This repo runs AI reviewers
   (`@zuke/ai`) that post their assessments as PR comments (and human reviewers
   do too). Before considering a PR done — and again after each push — fetch and
--- a/build/action_release.ts
+++ b/build/action_release.ts
@@ -24,6 +24,7 @@
 import { parse as parseYaml } from "@std/yaml";
 
 import actionVersion from "./action_version.json" with { type: "json" };
+import { compareSemver } from "./semver.ts";
 
 /** A parsed `v<major>.<minor>.<patch>` action tag. */
 export interface ActionVersion {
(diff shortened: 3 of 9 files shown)


totollygeek:
CI evidence, as promised in the description: the check now runs in CI instead of reporting itself skipped.

From this PR's own `Build & verify with Zuke` step:

> ℹ No published skill changed against origin/master; nothing to bump.

Compare with the same step on #440, before the fetch-depth change:

> ℹ Plugin version check skipped — the base ref "origin/master" is not present in this clone.

That is the whole defect in two lines: the check was never able to reach a base ref in CI, so it could not fail, and a skip reads as green. This branch changes no skill files, so "nothing to bump" is the correct verdict — the point is that a verdict was reached at all.

The two remaining behaviours cannot be shown from this run and are covered by unit tests instead: a push to master comparing against `HEAD^` (which only produces a verdict on a merge commit that touches skills), and the refusal of a version that moves down. Both are asserted in `tests/plugin_version_check_test.ts`, and the generated workflow wiring they depend on is asserted against `.github/workflows/ci.yml` directly.

One correction to the description: it refers to the gate fix as #446; this PR is #447. #446 is unrelated.

totollygeek:
Finding `ewhhxgcjt0h8` — "CI now hard-fails on any skipped plugin version check without a dedicated regression test" — accepted and fixed.

The observation was correct in the way that matters: the new fail-hard branch lived inline in `zuke.ts`, where a target body cannot readily be asked whether it throws, so the one behaviour this PR adds was the one behaviour with no test.

The decision now lives in `build/plugin_version_check.ts` as `reportFor(verdict, base, onCI)`, which returns a level and a message rather than printing or throwing. `zuke.ts` keeps only the dispatch — throw on `"error"`, print on `"info"` — and every outcome is asserted directly:

- a skip as information locally, and as a failure on CI whose message names the fix (`fetch-depth 0`), since the cause is a checkout setting nowhere near the check;
- an unreadable head manifest, which fails regardless of CI;
- skills changed with a stale version (the bump instructions);
- skills changed with the version moved up;
- a change touching no skill, which must never demand a bump.

On the risk behind the finding — that hard-failing could break runs for the wrong reason — the blast radius is one job. `pluginVersionCheck` runs only inside the `ci` target, which only the `ci` job in `ci.yml` invokes, and that job now checks out with full history. The other workflows run `integration`, `security` and `coreFloorCheck`, none of which include this target. A fork PR is unaffected: the checkout fetches the public base ref the same way.

`deno task ci` is green on the updated branch, and `./zuke pluginVersionCheck` still reports a real verdict rather than a skip.

PR #452, opened 2026-09-02
feat(deno): report test counts and coverage percentages into the build summary

### LINKED ISSUE #451: feat(deno): report test counts and coverage percentages into the build summary
### Pre-flight checklist

- [x] I searched existing issues and did not find a duplicate.

### Problem or motivation

#449 adds the per-target summary-note seam to `@zuke/core` (`ctx.reportSummary` and the ambient `reportSummary`), but nothing fills it in yet: a `test` target's row still reads `Succeeded  8.1s` with no counts. The Deno wrapper is the one Zuke's own build uses, so it should be the first to report, NUKE-style:

```text
test        Succeeded    146.7s  // Tests: 4047 · Passed: 4047 · Failed: 0
coverage    Succeeded      2.8s  // Lines: 98.4% · Branches: 97.5%
```

Split out of #449 because a wrapper that imports a symbol only the unreleased core exports cannot pass `coreFloorCheck` until that core publishes (see `RELEASING.md`), so the seam lands first and the wrapper follows.

### Proposed solution

```typescript
// DenoTestSettings overrides the ToolSettings.onOutput hook from #449 and
// parses deno's own result line, `ok | 837 passed | 0 failed (8s)`, so the
// counts are reported even when the run fails and the row is red:
//   Tests (passed + failed + ignored), Passed, Failed, and Ignored when non-zero.
// The JUnit and TAP reporters print no result line, so those runs report nothing.

// DenoTasks.coverage reports the measured percentages the same way:
//   Lines: 98.4%, and Branches: 97.5% when any branches were measured.

// Plus: the @zuke/deno floor on @zuke/core raised to the release that carries
// reportSummary, the skill and cheatsheet noting that DenoTasks.test reports
// by itself, and a plugin version bump.
```

### Alternatives considered

Parsing a JUnit report via `--junit-path` would cover every reporter, but it costs a file and changes the user's argv; the result line is always there for the default reporters.

### Additional context

Depends on #449 landing and `@zuke/core` publishing with `reportSummary`. Other test-runner wrappers (`@zuke/vitest`, `@zuke/jest`, `@zuke/bun`, …) can follow the same shape.

totollygeek:
## What & why

#450 added the per-target summary-note seam to `@zuke/core`. This PR fills it in for the wrapper Zuke's own build uses, so a `test` target's row finally says what happened, NUKE-style. Zuke's own gate with this branch prints:

    test        Succeeded    128.1s  // Tests: 4049 · Passed: 4049 · Failed: 0
    coverage    Succeeded      1.8s  // Lines: 98.4% · Branches: 97.5%

- **Test counts.** The Deno test settings override the `onOutput` hook from #450 and parse deno's own result line, so the counts are reported even when the run fails and the row is red: Tests (passed plus failed plus ignored), Passed, Failed, and Ignored when non-zero. The JUnit and TAP reporters print no result line, so those runs report nothing; this is documented on the settings class.
- **Coverage.** The coverage gate reports the measured line percentage, and the branch percentage when any branches were measured, onto the row of the target that ran it.
- **Core floor.** `@zuke/deno` now declares `@zuke/core@^1.44.0`, the release that carries `reportSummary`, with `deno.lock` regenerated. `./zuke coreFloorCheck` passes against the published core.
- **Docs and skills.** `docs/run-context.md` and `docs/console.md` say the wrapper reports by itself, the `zuke-write-build` skill and cheatsheet are updated, plugin manifests are bumped to 0.35.0, and the generated API references are regenerated.

Tests: unit tests for the result-line parser in each of its shapes (steps, ignored, filtered, coloured output, a look-alike line printed by a test), a real fixture run through the wrapper on both a passing and a failing suite, the JUnit no-report case, the coverage figures, and an integration test through the CLI. The fixtures are addressed by `file://` URL so the suite passes on Windows as well.

Everything in `deno task ci` passed locally except the `security` target, which the sandbox's network proxy blocks.

## Related issues

Closes #451

## Checklist

- [x] The PR title is a
      (`type(scope): summary`).
- [x] `deno task ci` passes locally (lint, fmt, type-check, tests, spell) — all targets except `security`, see above.
      changed.
      type guards instead).
      one implementation.
- [ ] A new package was wired into all six places (see
      if applicable.

https://claude.ai/code/session_01FrP7TrH4YPZiFmSE3mdhBT

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
       "name": "zuke",
       "source": "./plugins/zuke",
       "description": "Agent skills for Zuke — scaffold Zuke into a project (zuke-setup) and write or edit a zuke.ts build (zuke-write-build).",
-      "version": "0.34.0",
+      "version": "0.35.0",
       "author": {
         "name": "Zuke",
         "email": "contact@zuke.build",
--- a/deno.lock
+++ b/deno.lock
@@ -1608,7 +1608,7 @@
       },
       "packages/deno": {
         "dependencies": [
-          "jsr:@zuke/core@^1.25.0"
+          "jsr:@zuke/core@^1.44.0"
         ]
       },
       "packages/docker": {
--- a/docs/console.md
+++ b/docs/console.md
@@ -211,8 +211,8 @@ Log.summary(reports, totalMs, ok);
 ```
 
 A report's optional `summary` entries — the notes a target reported with
-`ctx.reportSummary`, or a wrapper reported through the ambient form — trail
-its row, dimmed:
+`ctx.reportSummary`, or a wrapper reported through the ambient form
+(`DenoTasks.test`'s test counts) — trail its row, dimmed:
 `test  Succeeded  8.1s  // Passed: 837`. See
 [Notes on the summary row](./run-context.md#notes-on-the-summary-row).
 
--- a/docs/run-context.md
+++ b/docs/run-context.md
@@ -66,6 +66,24 @@ class CI extends Build {
 }
 ```
 
+A wrapper that knows what its tool printed reports for you. `DenoTasks.test`
+puts `// Tests: 837 · Passed: 837 · Failed: 0` on the row of whichever target
+ran it:
+
+<!-- check -->
+
+```ts
+import { Build, target } from "jsr:@zuke/core";
+import { DenoTasks } from "jsr:@zuke/deno";
+
+class Checks extends Build {
+  test = target().executes(async (ctx) => {
+    await DenoTasks.test((s) => s.allowAll()); // reports the counts itself
+    ctx.reportSummary({ Version: "3.6.2" }); // the body adds what it knows
+  });
+}
+```
+
 Notes accumulate across calls in one target, and reporting a key again
 replaces its value in place. Each key and value renders on a single line
 (whitespace collapsed, colour codes removed) in both the terminal table and the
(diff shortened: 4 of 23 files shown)


totollygeek:
Two notes on the red checks on this draft.

**AI review, finding `j027qxrgsy7p`** ("process-output hook can be abused by wrapper subclasses to exfiltrate or mutate failed-run data", `packages/core/src/tooling.ts`): contesting as a false positive. The `onOutput` hook is `protected` and is only reachable by subclassing `ToolSettings`, which is the wrapper author's own code. A subclass could already override the public `run` method outright and see, alter, or forward everything the process produced, and every caller of `run` already receives the very same `CommandOutput` on success. The hook adds no capability a wrapper did not have; it only changes *when* the wrapper sees the output on a non-zero exit, before the `CommandError` is raised, so a test runner's counts can be reported for a red row. `CommandOutput`'s fields are `readonly`, so the hook cannot mutate what `run` returns or what the error carries, and the hook is reached only by code the build author chose to import. The rendering path the summary feeds is sanitised in one place (`singleLine` in `summary_note.ts`), as the reviewer's own verification of the sibling findings confirmed. Note this PR's diff is against its stacked base (#450); the hook itself lands there.

**Core version floors**: expected to fail until #450 merges and the `@zuke/core` release carrying `reportSummary` publishes. The floor cannot be raised before that (see `RELEASING.md`); the bump is the last commit this PR gets before it leaves draft.

---

totollygeek:
Re finding `3daiu0efcsmzb` (parser does not handle `filtered out` result lines): the claim as stated does not reproduce. The regex was anchored only at the start of the line, so `ok | 1 passed | 0 failed | 2 measured | 3 filtered out (4ms)` already parsed, and `packages/deno/tests/test_summary_test.ts` has asserted exactly that shape since the first commit.

What was true is that the parser matched the segments positionally, in the order deno prints them. Pushed 7d4a6b5 so it no longer does: the trailing duration is dropped, the line is split on the pipe separators, and `passed`, `failed` and `ignored` are read by name, so a reordered or newly optional segment cannot silence the counts. New tests cover shuffled segments, a duration following a steps suffix, and verdict lines that are not results. A structured source such as `--junit-path` was considered and rejected: it would add a file and alter the user's argv for a line deno already prints on every default-reporter run.

---

totollygeek:
Re finding `1xccidhwc9zom` (coverage report not exercised without a summary context): the no-context behaviour is a documented property of the ambient `reportSummary` in `@zuke/core`, which has its own unit test for being a no-op outside a running target, and every existing gate test in `packages/deno/tests/coverage_test.ts` already called `enforceCoverage` with no collector installed. None of them said so explicitly, though, so 849f391 adds a test that asserts the gate still passes and fails on its thresholds with no target running, and that nothing leaks into a collector installed afterwards.

Routing through a target context is not an option here: the gate is a plain function with no `ctx` in reach, and the ambient form exists for exactly that case, mirroring how a plain `$` in a body picks up the ambient cancellation signal.

---

PR #486, opened 2026-09-07
feat(core): refuse a build parameter that renders as a built-in CLI flag

### LINKED ISSUE #472: bug: a build parameter named `actor` shadows the reserved `--actor` flag on a registry spawn
Found while reviewing #471; **pre-existing on `master`**, not introduced by that work.

### What happens

`RESERVED_PARAM_NAMES` (`packages/core/src/params.ts:53-57`) reserves `dryRun`, `confirm` and `operatorToken`, so a build declaring a parameter with one of those names fails fast with a clear `ParameterError`. It does **not** reserve `actor`.

A registered build may therefore declare `actor = parameter(...)`. `validateParamArgs` renders it onto the child argv as `--actor=<value>` (`packages/core/src/mcp/registry_server.ts`), and `parseArgs` matches the built-in `--actor` before a declared parameter flag (`packages/core/src/cli.ts:325` `VALUE_FLAGS`, checked at `cli.ts:470` ahead of the `pf !== undefined` branch).

So an MCP caller passing that parameter sets the run’s `--actor`, overriding the `ZUKE_ACTOR` the registry runner exported for the resolved caller. The MCP audit trail still records the real caller — it is written by the server, before the spawn — but the child run’s own record is attributed to whatever the caller passed.

### Why it is worth fixing now

It has been latent because attribution came from `--actor`/the environment anyway. It matters more as identity becomes something a server resolves and vouches for: #471 made a run’s actor come from an authenticated identity, and the planned role policy will make attribution an input to decisions rather than only a label.

### Suggested fix

Add `actor` to `RESERVED_PARAM_NAMES`, so the collision is a build-time `ParameterError` naming the field — the same treatment `dryRun` and `operatorToken` already get. Worth checking the other reserved CLI flags in `VALUE_FLAGS` against the same list while there; any flag a rendered parameter can shadow has the same shape of problem.

A test belongs beside the existing reserved-name tests, plus an integration case proving a registry spawn keeps the exported `ZUKE_ACTOR`.

### LINKED ISSUE #485: bug: a target named after a reserved command is advertised but unreachable
Found by an adversarial review of #472, which fixed the same shape of bug for **parameters**. This is the **target** half, and it is pre-existing on `master`.

### What happens

`RESERVED_COMMANDS` (`packages/core/src/cli_spec.ts`) lists the positional words that are not target names: `graph`, `generate-ci`, `completions`, `mcp`, `resume`, `runs`, `cancel`, `force`, `register`, `doc`, `outdated`.

Nothing checks a build's target names against that list. `grep -rn "RESERVED_COMMANDS"` finds no use in `build.ts`, `graph.ts` or `target.ts`. A target named `graph` is therefore discovered, validated, and listed — and unreachable by name, because the positional branch in `parseArgs` matches the reserved command first.

Reproduced through the real `main()`:

- `zuke --list` advertises the target, with its description and dependencies
- `zuke graph` exits 0 having rendered the dependency graph, and the target's body never runs
- shell completion offers it
- an MCP host exposes `run:graph`, which *does* execute it
- it still runs as a dependency of another target

So the target is advertised as runnable in three places, runs in two of them, and silently does something else in the one a person is most likely to type.

### Why it is worth fixing

It is exactly the failure #472 refused for parameters: a declaration the engine accepts, advertises, and then quietly does not honour. The blast radius is arguably worse here, because `--list`, completions and the MCP tool list all present the target as runnable, and `zuke <target>` exits **0** — so there is no error to notice.

### Suggested fix

Refuse the collision at target discovery, the way `discoverParameters` now refuses a parameter that renders as a built-in flag: a `TargetError`/`GraphError` naming the field and the command word. `RESERVED_COMMANDS` is already the shared list the parser, the help text and completions read, so the check derives from it and needs no second list.

Worth deciding deliberately, because unlike the parameter case a colliding target is not merely awkward — it is genuinely reachable as a dependency and over MCP, so refusing the build is a breaking change for anyone who has one. #472 took the same trade knowingly and shipped as a minor; this should probably do the same.

A test belongs beside the target-discovery tests, plus an integration case proving `zuke <reserved>` no longer exits 0 having run the built-in.

totollygeek:
## What & why

A build could declare a parameter named after a built-in CLI flag. The parser matches a built-in ahead of a declared parameter of the same name, so the flag means the built-in and not the parameter.

For `--actor` that is an attribution hole, which is what the issue was filed about. A registered build declaring `actor = parameter(...)` has its value rendered onto the spawned child's argv, where it overrides the `ZUKE_ACTOR` the registry runner exported for the resolved caller. Reproduced on master, with the environment set exactly as the runner sets it:

- `ZUKE_ACTOR=alice zuke greet --state --actor=evil` records the run under `evil`, and the declared parameter resolves to nothing.

Discovery now refuses the collision, the way it already refuses the MCP control names, with a `ParameterError` naming both the field and the flag. The reserved set is derived from the `BUILTIN_FLAGS` list the parser and the help text already share, so a flag added to the CLI is reserved without a second list to remember. Only the rendered flag collides, so `actorName` and a grouped `runs.limit` stay usable, and no parameter in this repo changes.

### Why this is a minor and not a patch

I originally wrote this as a `fix`, on the premise that such a parameter never received a value. The adversarial review disproved that, and the distinction decides the version:

- its **environment variable** always resolved it — for all 38 names;
- an **inline `--flag=value`** reached it for the 16 boolean built-ins, which are matched by an exact comparison and so fall through to the parameter branch;
- the **in-process MCP run tool** reached it for all of them, keyed by parameter name, never touching the parser.

So a build using `state`, `check`, `limit`, `target` or `output` as a parameter and driving it by environment variable worked, and now refuses to load. That is the intended outcome — the same parameter breaks the moment anyone reaches for its flag, and then the value quietly does something else — but it is a change to the authoring surface, so it ships as a minor with the breakage stated rather than as a silent patch.

Two reviewers attacked this. Both found something real, and the first found the same hole the change set out to close.

**The fix guarded only the build's own process** (high). A registry descriptor carries each parameter's `flag` as a plain string that nothing validated, and the MCP host renders it straight onto the spawned child's argv without ever calling `discoverParameters`. A descriptor declaring a parameter named `message` with flag `actor` spawned `--actor=intruder`, and the run was recorded under the caller's value over the actor the host had resolved. The code's own comments describe a registry backend as a service Zuke does not control, so this was the guarded claim failing at the one boundary that most needed it. Reachable without tampering, too: a build registered by an older Zuke that legitimately declared `actor = parameter(...)` stores `flag: "actor"` honestly, and the descriptor outlives the upgrade. Descriptor parsing now refuses a flag that does not follow from its parameter's name, and a name that renders as a built-in — at the parse boundary both the filesystem and HTTP backends go through.

**The premise was wrong, and three places asserted it** (medium). The doc, the JSDoc and the thrown message all said the parameter would never receive a value. All three now say what is true, and say why partial reachability is the reason to refuse rather than to document.

**The refusal escaped as an uncaught throw** (medium-high) with a stack trace, where Zuke's convention one line below is a named message and exit 1. Worse, it ran before the help branch, so `zuke --help` and `zuke --list` crashed too and the author could not inspect the build they were being told to fix. It now reads as a named message and exit 1, like an invalid graph. That covers the pre-existing reserved-name refusal as well.

Also fixed: one iteration of the flag loop test passed on the pre-existing control-key check rather than the new one, and the second documented command line in the parameters guide omitted a required value.

Reported and deliberately not changed here: parameter environment variables can still collide with Zuke's own `ZUKE_*` names. Verified as a footgun rather than a bypass — in every channel where it could matter, the build is already trusted code in that process, and unlike the flag case the parameter does still get its value.

**Filed rather than fixed: [#485](https://github.com/zuke-build/zuke/issues/485).** The same shape of bug exists for target names against `RESERVED_COMMANDS`. A target named `graph` is discovered, listed, offered in completions and runnable over MCP, but `zuke graph` renders the dependency graph and exits 0 without running it. It is a separate surface with its own breaking-change profile, so it gets its own issue rather than riding along here.

## Documentation

The parameters guide gained a "Names a parameter may not use" section, and the same rule is in the agent skill and its cheatsheet, with the plugin version bumped in all four manifests.

Fixing a pre-existing bug found while writing it: the guide's headline example declared `dryRun`, which the existing reserved-name check has always refused, so the flagship example threw on load — and both of its documented command lines omitted a required parameter, so neither ran. The gate did not catch either, because `snippetsCheck` type-checks a snippet without constructing or running the build. Both command lines now run against the example.

## Testing

Unit coverage of the refusal, of every flag in the shared list, and of the names that stay allowed; descriptor coverage of the tampered flag, the honestly-named one and the pre-M12 shape; and integration coverage driving the real CLI, asserting the clean exit rather than a throw, and pinning that `--actor` and `--actor-name` do not contend.

The review confirmed the tests are not vacuous: reverting the source in a scratch copy fails exactly the new tests and nothing else.

`deno task ci`: 21/21 targets, 4250 tests, 98.5% lines and 97.4% branches against a 95% gate.

## Related issues

Closes #472

## Checklist

- [x] Docs updated in the same PR: the parameters guide, JSDoc, the agent skill and its cheatsheet.
- [x] Public API regenerated — not applicable; the new list lives in an internal module that no entrypoint exports.
- [x] Agent skills updated and the plugin version bumped in all four manifests, 0.42.0 to 0.43.0.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
       "name": "zuke",
       "source": "./plugins/zuke",
       "description": "Agent skills for Zuke — scaffold Zuke into a project (zuke-setup) and write or edit a zuke.ts build (zuke-write-build).",
-      "version": "0.42.0",
+      "version": "0.43.0",
       "author": {
         "name": "Zuke",
         "email": "contact@zuke.build",
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -23,14 +23,14 @@ class Deploy extends Build {
 
   workers = parameter("Parallel upload workers").number().default(4);
 
-  dryRun = parameter("Print actions without performing them").boolean();
+  preview = parameter("Print actions without performing them").boolean();
 
   // A required list: `.required()` comes before `.array()` — the reverse order
   // does not type-check.
   repos = parameter("Repos to deploy").required().array();
 
   deploy = target().executes(() => {
-    if (this.dryRun.value) console.log("(dry run)");
+    if (this.preview.value) console.log("(preview)");
     console.log(
       `Deploying ${this.repos.value.join(", ")} to ${this.environment.value} ` +
         `with ${this.workers.value} workers`,
@@ -42,7 +42,7 @@ await run(Deploy);
 ```
 
 ```sh
-./zuke deploy --environment production --workers 8 --dry-run
+./zuke deploy --environment production --workers 8 --preview
 ENVIRONMENT=staging ./zuke deploy        # value from the environment
 ```
 
@@ -151,6 +151,25 @@ The flag and environment variable are derived from the property name:
 `environment` → `--environment` / `ENVIRONMENT`; a camelCase name like
 `targetEnv` → `--target-env` / `TARGET_ENV`. Override the environment variable
 with `.env("NAME")`.
+
+### Names a parameter may not use
+
+Two sets of names are refused when the build is loaded, with a `ParameterError`
+naming the field:
+
+- **A name that renders as a built-in CLI flag** — `actor` → `--actor`,
+  `actorKind` → `--actor-kind`, `limit` → `--limit`, and so on for every flag
+  `zuke --help` lists. The parser matches a built-in ahead of a declared
+  parameter of the same name, so such a parameter would never receive a value:
+  `--actor=alice` would attribute the run and leave the parameter unresolved.
+- **`dryRun`, `confirm` and `operatorToken`** — the control keys an MCP
+  `run:<target>` tool adds to its input schema alongside the build's
+  parameters, which would shadow a parameter of the same name.
+
+Only the rendered flag matters, so a longer or nested name is fine:
+`actorName` → `--actor-name` and a grouped `runs.limit` → `--runs-limit` both
+stay usable. Rename the field; there is no opt-out, because the alternative is
+a flag that silently belongs to something else.
 
 ## Without the CLI
 
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -42,8 +42,8 @@ await run(Deploy);
 ```
 
 ```sh
-./zuke deploy --environment production --workers 8 --preview
-ENVIRONMENT=staging ./zuke deploy        # value from the environment
+./zuke deploy --environment production --workers 8 --repos api,web --preview
+ENVIRONMENT=staging REPOS=api,web ./zuke deploy   # values from the environment
 ```
 
 ## Declaring
@@ -159,17 +159,22 @@ naming the field:
 
 - **A name that renders as a built-in CLI flag** — `actor` → `--actor`,
   `actorKind` → `--actor-kind`, `limit` → `--limit`, and so on for every flag
-  `zuke --help` lists. The parser matches a built-in ahead of a declared
-  parameter of the same name, so such a parameter would never receive a value:
-  `--actor=alice` would attribute the run and leave the parameter unresolved.
-- **`dryRun`, `confirm` and `operatorToken`** — the control keys an MCP
-  `run:<target>` tool adds to its input schema alongside the build's
-  parameters, which would shadow a parameter of the same name.
+  `zuke --help` lists. The parser matches a built-in first, so the flag means
+  the built-in and not the parameter: `--actor=alice` attributes the run and
+  leaves the parameter unresolved.
 
-Only the rendered flag matters, so a longer or nested name is fine:
-`actorName` → `--actor-name` and a grouped `runs.limit` → `--runs-limit` both
-stay usable. Rename the field; there is no opt-out, because the alternative is
-a flag that silently belongs to something else.
+  Such a parameter is not wholly dead — its environment variable still resolves
+  it, as does an MCP `run:<target>` call — and that is exactly why it is refused
+  rather than merely documented. It works until someone reaches for its flag,
+  and then the value quietly does something else.
+- **`dryRun`, `confirm` and `operatorToken`** — the control keys an MCP
+  `run:<target>` tool adds to its input schema alongside the build's parameters,
+  which would shadow a parameter of the same name.
+
+Only the rendered flag matters, so a longer or nested name is fine: `actorName`
+→ `--actor-name` and a grouped `runs.limit` → `--runs-limit` both stay usable.
+Rename the field; there is no opt-out, because the alternative is a flag that
+silently belongs to something else.
 
 ## Without the CLI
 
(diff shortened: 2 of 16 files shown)


totollygeek:
### `2c5tzza1o49pz` — "Descriptor parsing now rejects pre-existing descriptor flags that don't match the new name-derived rule"

Refuted, but the finding was right that the safety of the check rested on three unwritten assumptions. It is now pinned by a test rather than by an argument — `8c5c384`.

**No descriptor in the wild can fail the check.** Three things have to hold, and all three do:

1. `describeCli` is the **sole producer** of a parameter's `flag`, at `packages/core/src/describe.ts:105`, and it always sets `flag: flagName(name)`. Grepping every other `flag:` in `packages/core/src` finds only the parser's own `ParamFlag`, which never reaches a descriptor.
2. `flagName` is **idempotent**, which is exactly what the pre-M12 path needs: a descriptor with no `name` recovers it from the flag, so `flagName(flag)` has to equal `flag`. It does — the function lowercases and inserts dashes, so anything it emits is already a fixed point.
3. `flagName` has changed **once**, in `3dc6f17` (2026-06-17), which added the dot replacement. `packages/core/src/registry/descriptor.ts` was first committed in `b55a876` (2026-07-19). The change predates descriptors by a month, so no descriptor has ever been written by a Zuke whose `flagName` produced a different flag for the same name.

That third point is the one that makes the finding's scenario unreachable rather than merely unlikely, and it is the one I would have wanted a reviewer to check hardest, so it is worth stating explicitly.

**What changed as a result.** Assumptions 1 and 2 were load-bearing and untested, which is a fair thing to flag even though the conclusion does not hold. The new test builds a surface with `describeCli` from a build carrying a name near a built-in, one with consecutive capitals, and two grouped parameters; round-trips it through `stringifyBuildDescriptor`/`parseBuildDescriptor`; then drops every `name` to reproduce the pre-M12 shape and asserts each flag comes back unchanged. Mutating `flagName` to break idempotence fails it, so it is not a test that passes for free.

**On the blast radius, since it is adjacent.** A refused descriptor is fail-closed by design, and it does not take the registry down with it: `listBuilds` already catches a parse failure per file and continues, so one poisoned entry cannot hide the others. `getBuild` does throw, which is the intended outcome for a descriptor that would spawn a run under an actor of its own choosing.

I considered the tolerant alternative — re-derive the flag from the name and ignore what the descriptor stored — which would remove the rejection class entirely while still killing the attack. I kept the refusal, because a descriptor whose flag does not follow from its name is evidence of tampering in a field the code's own comments describe as coming from a service Zuke does not control, and silently correcting it would discard that evidence.

**On the refuted finding**, the verification was right to reject the duplicated-reserved-name claim: `params.ts` and `registry/descriptor.ts` both read `BUILTIN_FLAG_NAMES` from `cli_spec.ts`, which is the single list the parser and the help text already share, so there is nothing to drift.

PR #488, opened 2026-09-08
feat(core): serve OAuth protected resource metadata from the MCP server

### LINKED ISSUE #487: feat: make the MCP HTTP server an OAuth 2.0 protected resource
Follow-up to #480/#483, and the replacement for the `@zuke/mcp-oauth` proposal in the customer plan (epic 15.3).

### Why not an authorization server

The original proposal was for Zuke to be its own OAuth 2.1 authorization server — dynamic client registration, PKCE, refresh families, an upstream GitHub/OIDC leg. That is the part of OAuth where a subtle bug is an authentication bypass, and it was proposed largely because a GitHub OAuth App is not itself an authorization server for our resource.

That premise no longer holds, and it was checked rather than assumed. Two things settle it:

- **Dynamic client registration is no longer required.** In MCP revision `2026-07-28` DCR is `MAY` and formally **deprecated**, with Client ID Metadata Documents as the replacement. It was demoted from `SHOULD` in `2025-11-25`. The client priority order is pre-registered client, then CIMD, then DCR, then prompt.
- **A pre-registered client id genuinely works.** Driven end to end against a stub authorization server with Claude Code 2.1.263: the client fetched both discovery documents, went straight to `/authorize` with the configured `client_id`, exchanged the code with a PKCE verifier, and authenticated — with **zero** calls to the registration endpoint. Without a configured client id it falls through to DCR and fails hard, which is what the older bug reports describe.

So Zuke can be a plain OAuth 2.0 **protected resource** pointing at whatever IdP an operator already runs, and implement none of the dangerous half. The `mcpAuth()` seam added in #473 already accepts any `McpAuthenticator`, and a build file is ordinary user code that may import a vetted JOSE library — so token validation is already pluggable and does not belong to us either.

### What is missing

Discovery. A client that has never authenticated has no token, and asks over unauthenticated GETs that `serveHttp` currently rejects — its first check is POST-only.

1. **A public-route seam on `serveHttp`**, ahead of the method and authentication checks. Anything registered there is unauthenticated by construction, so the name must say so. These routes need permissive CORS and no `Origin` gate: the documents are public, are fetched cross-origin by browser-based clients, and the challenge needs `Access-Control-Expose-Headers: WWW-Authenticate` to be readable at all.
2. **The protected resource metadata document** (RFC 9728), from typed build configuration: `resource`, `authorization_servers` (RFC marks it optional, MCP makes it **required**), plus the recommended `scopes_supported`, `bearer_methods_supported`, `resource_name`. Zero-value parameters must be omitted rather than serialised as empty arrays.
3. **The path.** The well-known segment is inserted **between host and path**, so an endpoint at `/mcp` serves at `/.well-known/oauth-protected-resource/mcp`. Appending instead of inserting is the most common implementation bug in the wild. RFC 9728 section 3.3 also self-conflicts for a path-having endpoint — a client that probes the root computes a different expected identifier than one that follows the header, and each is told it must discard a mismatch — so the document is served at **both** locations, which the RFC explicitly permits, with the path-inserted URL always advertised.
4. **A richer challenge.** `resource_metadata` as an absolute URL plus `scope`. A bare unauthenticated request must carry **no** `error` parameter; `error="invalid_token"` belongs only on a request that presented a token and had it rejected. These are two cases and one frozen constant today.
5. **Authorization denials become `403` with `insufficient_scope`**, naming every required scope in one challenge. A `401` tells a client its session is bad and sends it into a full re-discovery and re-authentication loop, when the truth is that it is authenticated and simply not permitted. Role denials from #483 currently answer inside the JSON-RPC body.

### Documentation

The failure mode that actually bites operators is three strings that must agree byte for byte: the `resource` in the metadata document, the `resource` parameter the client sends, and the audience the IdP mints into the token. The guide needs those named together, with a worked example per IdP, and a recipe for the `mcpAuth()` authenticator using a JOSE library from the build.

### Acceptance

`claude mcp add --transport http <url>` against a Zuke server behind a real IdP authenticates in a browser with nothing pasted; the audit trail names the authenticated subject; a caller with a valid token but insufficient roles is refused with `403` and `insufficient_scope` rather than being sent back through login; and the discovery documents validate against RFC 9728 for both a root and a path-having endpoint.

Out of scope, deliberately: any `/authorize`, `/token` or `/register` endpoint. Those belong to the identity provider.

totollygeek:
## What & why

This is epic 15.3 from the customer plan, and it is much smaller than that plan asked for — deliberately.

The proposal was for Zuke to become its own OAuth 2.1 **authorization server**: dynamic client registration, PKCE, refresh families with single-use identifiers, a GitHub and a generic OIDC upstream leg, a pluggable auth store, a stubbed-IdP conformance suite. Dependency-free, in a new package, size L. That is the half of OAuth where a subtle bug is an authentication bypass rather than a broken feature.

It was proposed because a GitHub OAuth App is not an authorization server for our resource, and something has to bridge it. Two facts, both checked rather than assumed, removed the need:

- **Dynamic client registration is no longer required.** MCP asked for it in `2025-06-18`, demoted it to optional in `2025-11-25`, and marks it **deprecated** in the current `2026-07-28`. The sanctioned routes now are a pre-registered client id or a client id metadata document.
- **A pre-registered client id genuinely works.** Driven end to end against a stub authorization server with a real client: discovery, `/authorize` with the configured id, code exchange with a PKCE verifier, authenticated — and **zero** calls to a registration endpoint. Without a configured id it falls through to registration and fails hard, which is what the older bug reports describe.

So Zuke can be a plain OAuth 2.0 **protected resource** pointing at whatever identity provider an operator already runs, and implement none of the dangerous half. Token verification was already pluggable through the `mcpAuth()` seam from #473, and a build file is ordinary code that may import a maintained JOSE library even though the published packages may not — so that half is neither ours to write nor ours to ship.

### What was actually missing

Discovery. A client that has never authenticated has no token, and asks over unauthenticated `GET`s that the transport rejected — its first check is POST-only.

`mcpProtectedResource()` declares the endpoint a protected resource. The RFC 9728 document is published, and named in every `WWW-Authenticate` challenge — a 401 now carries a Bearer challenge with the declared `scope` and a `resource_metadata` parameter naming the document, which is the whole of what a client needs to find the identity provider. Zuke hosts no authorize, token or register endpoint. Those stay with the provider.

### Verified against a real client, not only against the RFC

A real MCP client, pointed at a running Zuke server: it followed the challenge, fetched the document, read the authorization server out of it, fetched that server's metadata, and settled on "needs authentication". The observed discovery order was `POST /mcp` → `/.well-known/oauth-protected-resource/mcp` → the authorization server's metadata → `/authorize`.

That is also how the **path-insertion** rule got pinned. The well-known segment goes *between host and path*, so `/mcp` publishes at `/.well-known/oauth-protected-resource/mcp`. Appending it instead is the common bug in the wild, it is what the RFC's name suggests, and a client fetches only the inserted form — so getting it wrong means discovery silently never happens.

Two reviewers. Both found real defects; the worst was a repeat of one I had already fixed.

**An unguarded `new URL(request.url)` ran before every other check** (high). Deno builds that URL from the raw `Host` header and accepts hosts the WHATWG parser rejects, so `Host: bad host` faulted **every** request with a `500` — authenticated ones included — and wrote a stack trace per request into the operator's log. Declaring a protected resource was enough to hand any unauthenticated caller a denial of service. The identical trap is documented four functions below in the same file, where it was fixed during the authenticator-seam work; this reintroduced it one call earlier. The regression drives a raw socket and fails without the guard.

**The root copy of the document was published on a rationale that was backwards** (medium). I had it serving at both the path-inserted and root locations, reasoning that the two routes validate differently. They do — and that is the argument against it: RFC 9728 has a client derive the `resource` it expects from the URL it fetched, so a client that probed the root expects the bare origin and **must discard** a document naming a path. A conformant client tries the path-inserted URL first and never asks for the root. That copy had no correct consumer, so it is gone. A resource declared as a bare origin still publishes at the root, because for that identifier the root *is* the derived location.

**The `resource` field was re-serialised rather than emitted as declared** (medium). `new URL(x).toString()` is not byte-preserving — it appends a slash to a bare origin — and this is the field whose entire purpose is to state a byte-for-byte agreement between three strings. Normalising it broke the contract it documents.

**`withResourceMetadata` matched its own parameter with a substring test** (medium), wrong in both directions: an `error_description` merely mentioning the word suppressed a real metadata URL and silently killed discovery, and a challenge spelling it `Resource_Metadata =` — auth-param names are case-insensitive and space around `=` is legal — got a second copy, which RFC 7235 forbids and leaves clients to resolve however they like. My first fix, a word-boundary regex, still matched inside the quoted value; it now blanks quoted strings first.

**Identifiers that parse but cannot produce a well-known URL passed startup and threw per request** (medium-low) — a `urn:` or `file:` scheme, where `origin` is the string `"null"`. That turned a startup check into a `500` on a public route. Scheme, query string and userinfo are all refused at declaration now; userinfo especially, since it would otherwise be published verbatim in an unauthenticated, cacheable, publicly readable document.

**`Access-Control-Expose-Headers` was inert** (low). The `401` carries no `Access-Control-Allow-Origin`, so a browser rejects the whole response before any header is readable. A comment claimed it was what made discovery visible to browser clients, and a test asserting its presence made the no-op look covered. Both removed.

Also fixed: challenges now name the declared scopes, which the specification asks for and which is what finally gives `bearerChallenge` a caller — it was exported with none; a dead branch whose condition no call site can meet; an OAuth 2.1 citation that was §5.3.1, not §5.3.2; and the document response, `HEAD`, and the startup refusal had no tests at all.

**One claim of mine was wrong and is corrected rather than quietly dropped.** The first commit said a build declaring no protected resource behaves exactly as before. It does not: a bearer token that was *presented and rejected* now answers `invalid_token` rather than a generic `Unauthorized`. That is what OAuth defines it as, and it tells a client its stored credential — rather than its absence — is the problem, so the change is kept. But the test whose name promised "unchanged" never sent a wrong token, which is the one case that changed. It does now, and it says what actually holds.

## Scope note

The issue also proposed answering authorization denials with `403` and `insufficient_scope`. The challenge builder and the transport support it, and an authenticator refusing for insufficient scope emits a correct `403`. Per-tool **role** denials are deliberately left in band: an MCP `tools/call` denial returned as a transport-level `403` is indistinguishable to a client from a session-level failure, so it would provoke a re-authentication on every denied call — the loop the `403` exists to avoid. The specification does define a step-up flow for this, so it is a judgement call rather than a settled one, and worth revisiting if a client ever wants it.

## Testing

14 unit tests over the document, the path derivation and the challenge builder; 11 integration tests driving a real HTTP server, including the malformed-`Host` regression over a raw socket. Each regression was mutation-tested — reverting the fix fails the test.

`deno task ci`: 21/21 targets, 4278 tests, 98.5% lines and 97.3% branches against a 95% gate.

## Related issues

Closes #487. Filed separately and not in this PR: Zuke advertises MCP protocol `2025-06-18` while the current revision is `2026-07-28`.

## Checklist

- [x] Docs updated in the same PR: the MCP guide, JSDoc, the agent skill and its cheatsheet.
- [x] Agent skills updated and the plugin version bumped in all four manifests, 0.43.0 to 0.44.0.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
       "name": "zuke",
       "source": "./plugins/zuke",
       "description": "Agent skills for Zuke — scaffold Zuke into a project (zuke-setup) and write or edit a zuke.ts build (zuke-write-build).",
-      "version": "0.43.0",
+      "version": "0.44.0",
       "author": {
         "name": "Zuke",
         "email": "contact@zuke.build",
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -516,6 +516,18 @@ The `authorizationServer(...)` value is the provider's **issuer identifier**
 `issuer` in that provider's own metadata document, or a client is required to
 reject it.
 
+**Everything in this document is public.** It has to be: a caller asking where
+to authenticate has nothing to authenticate with, so the route is served before
+the bearer token and the authenticator, and it is readable cross-origin because
+browser-based clients fetch it that way. Treat the four fields you supply as
+published — the resource URI, the issuer, the scope names, and the resource name
+and documentation URL. On a machine where the server binds loopback, that also
+means a web page the developer happens to visit can read them; an internal
+hostname in the resource URI is the realistic thing to think twice about. A
+credential in the identifier is refused outright rather than left to judgement,
+since a `https://user:pass@host/mcp` would otherwise be published verbatim in a
+cacheable document.
+
 The path is derived for you, because getting it wrong is silent. The well-known
 segment is inserted **between the host and the path**, so an endpoint at `/mcp`
 publishes at `/.well-known/oauth-protected-resource/mcp` rather than the
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -516,25 +516,35 @@ The `authorizationServer(...)` value is the provider's **issuer identifier**
 `issuer` in that provider's own metadata document, or a client is required to
 reject it.
 
-Two details are handled for you, because both are silent when wrong. The
-well-known segment is inserted **between the host and the path**, so an endpoint
-at `/mcp` publishes at `/.well-known/oauth-protected-resource/mcp` rather than
-the appended form — a client fetches only the inserted one. And the identical
-document is also published at the root, because a client that had no challenge
-to follow constructs the root URL instead, and the two routes validate the
-`resource` field against different expectations.
+The path is derived for you, because getting it wrong is silent. The well-known
+segment is inserted **between the host and the path**, so an endpoint at `/mcp`
+publishes at `/.well-known/oauth-protected-resource/mcp` rather than the
+appended form — a client fetches only the inserted one.
 
-Dynamic client registration is **not** required. It was demoted to optional and
-then deprecated in the MCP specification, and a client configured with a
-pre-registered client id never asks for it — verified against a real client,
-which goes straight from discovery to `/authorize` with the id it was given. So
-a provider that does not offer registration, including a GitHub OAuth App, is
-usable: register the client once, out of band.
+It is published at exactly that one location. Serving a second copy at the root
+looks like cheap insurance and is not: RFC 9728 has a client derive the
+`resource` it expects from the URL it fetched, so a client that probed the root
+expects the bare origin and must discard a document naming a path. That copy
+would have no correct consumer — a conformant client tries the path-inserted URL
+first and never asks for the root, and one that only asks for the root would
+throw away what it found. A resource declared as a bare origin publishes at the
+root, because for that identifier the root *is* the derived location.
+
+Dynamic client registration is **not** required. The MCP specification asked for
+it in revision `2025-06-18`, demoted it to optional in `2025-11-25`, and marks
+it deprecated in the current `2026-07-28` — client id metadata documents and
+pre-registration are the sanctioned routes now. Verified against a real client:
+given a pre-registered client id it goes straight from discovery to
+`/authorize` and never touches a registration endpoint. So a provider that
+offers no registration — a GitHub OAuth App, for one — is usable: register the
+client once, out of band.
 
 Verifying tokens is still not Zuke's job. Do not hand-roll it — import a
 maintained JOSE library in your build file (the build is ordinary code and may
 depend on whatever you like, unlike the published packages) and let it do the
-signature, the claims and the key rotation:
+signature, the claims and the key rotation. The fragment below shows the shape,
+not a complete file — `UNAUTHORIZED`, `INVALID_TOKEN` and `McpAuthenticator`
+come from `jsr:@zuke/core`, and `rolesOf` is yours to write:
 
 ```ts
 import { createRemoteJWKSet, jwtVerify } from "npm:jose";
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -1,13 +1,13 @@
 # MCP server
 
-`./zuke mcp` — a command on **your build's own CLI**, not the globally
-installed `jsr:@zuke/cli` (see the [CLI reference](./cli.md) for the
-difference) — runs a [Model Context Protocol](https://modelcontextprotocol.io)
-server over your build. MCP is the open standard that lets an AI client — Claude
-Desktop, Claude Code, an IDE, any agent — discover a server's **tools** (typed,
-schema-described functions) and call them. Pointing a client at `./zuke mcp`
-lets an agent **operate the pipeline through typed calls** — list the targets,
-inspect the graph, run one with the right parameters — instead of guessing shell
+`./zuke mcp` — a command on **your build's own CLI**, not the globally installed
+`jsr:@zuke/cli` (see the [CLI reference](./cli.md) for the difference) — runs a
+[Model Context Protocol](https://modelcontextprotocol.io) server over your
+build. MCP is the open standard that lets an AI client — Claude Desktop, Claude
+Code, an IDE, any agent — discover a server's **tools** (typed, schema-described
+functions) and call them. Pointing a client at `./zuke mcp` lets an agent
+**operate the pipeline through typed calls** — list the targets, inspect the
+graph, run one with the right parameters — instead of guessing shell
 invocations.
 
 It's a natural extension of what Zuke already does: it publishes `llms.txt`, a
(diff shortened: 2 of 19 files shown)


totollygeek:
### `27358rpkhyyud` — "Diff is truncated before the new resource-metadata module ends, so the public API cannot be fully reviewed"

This one is true, and it is not a defect — which makes it worth answering carefully rather than contesting.

The finding is that the reviewer could not see all of `packages/core/src/mcp/resource_metadata.ts`, not that anything in it is wrong. Both findings that *were* about the code — a missing integration test for the discovery flow, and an unauthenticated metadata leak on a non-loopback bind — were refuted by the verification step against the real file, and the security reviewer scored 0/10 with nothing at all. So the gating score comes entirely from the summary written before verification, and specifically from the truncation.

For what is past the cut, the module is small and its whole public surface is: `ProtectedResourceSettings` (a fluent builder — `authorizationServer`, `scopes`, `name`, `documentation`), the `protectedResource(resource)` entry point, `metadataPath`, `metadataUrl`, `metadataDocument`, and `ProtectedResourceError`. The only non-obvious logic is `resourceUrl`, which validates and refuses four shapes — a non-absolute URL, a non-http(s) scheme, a fragment, a query string, and userinfo — and `metadataPath`, which inserts the well-known segment between host and path. Every one of those has a unit test, and `deno doc --lint` passes across all six `@zuke/core` entrypoints, so nothing in it is undocumented.

The part I would rather have you weigh than my summary: that module was read in full by two independent adversarial reviewers before this PR was opened, and they did not come back empty. They found an unguarded `new URL(request.url)` that faulted every request with a 500 on a malformed `Host` header, a root-published copy of the document that a conformant client must discard, a `resource` field that was re-serialised in the one place byte-for-byte identity is the contract, a parameter-matching bug that suppressed discovery in one direction and duplicated a parameter in the other, and identifiers that passed startup and threw per request. All fixed, each with a regression test, and the second commit on this branch is exactly that work.

**On the underlying cause, which is the actionable part:** the diff is large — two commits, a new module, a new build override, transport changes, docs, skills, and 25 tests. If the review budget cannot span it, that is a signal about PR size rather than about this file, and splitting the next one is the fix. I would rather not add this to `suppressions(...)`: a truncation warning is worth keeping live for the PR where it means something, and suppressing the id here would silence it globally.

Re-running the review now that the PR body no longer carries a fenced block, in case the retry fits the budget.

totollygeek:
### `1ppsiq7k03zxe` — "Unauthenticated metadata endpoint can disclose protected-resource details on any path match"

Correct on the mechanics, and I have acted on the half that is actionable — `db4c4ef`.

**The disclosure is the mechanism, not a hole.** A caller asking *where to authenticate* has, by definition, nothing to authenticate with. RFC 9728 is that conversation: the `401` names a document, the document names the authorization server. Serving it before the bearer check and the authenticator is what makes it reachable at all, and it is deliberately readable cross-origin because browser-based MCP clients fetch it that way and cannot read a response their browser rejected first. So "gate it more tightly" would not harden the feature, it would remove it — a gated discovery document is one no unauthenticated client can use, which is every client that has not logged in yet.

It is also the narrowest possible surface. The route matches one exact path (`path === metadataAt`, a single precomputed string), answers only GET, HEAD and OPTIONS, and returns a constant rendered once at startup. Every other method and path falls through to the full guard chain — `POST` to that same path is still `401`, and still `403` on a bad `Origin`.

**What the finding is right about is the advice, and that was missing.** The document is not a leak but it *is* a publication, and nothing told an operator so. The guide now says it plainly: the four fields you supply — the resource URI, the issuer, the scope names, the resource name and documentation URL — are published. It also names the case actually worth a moment's thought, which the finding gestures at: on a **loopback** bind this is newly readable by any web page the developer happens to visit, so an internal hostname in the resource URI is a real decision rather than an incidental one.

**One part is not left to judgement.** A credential in the resource identifier is refused at declaration rather than documented as unwise — `https://svc:secret@build.internal/mcp` would otherwise be published verbatim in an unauthenticated, `max-age=3600` cacheable document, and "we told you not to" is the wrong control for that. The same validation refuses a non-http(s) scheme, a query string and a fragment, each for its own reason.

Worth noting the finding is marked `[existing]`: this endpoint predates the PR only in the sense that the transport did, but the unauthenticated route is new here, so I have treated it as new rather than inherited.

PR #491, opened 2026-09-08
feat(core): advertise MCP protocol revision 2025-11-25

### LINKED ISSUE #489: chore: advertise a current MCP protocol revision
Found while implementing #487.

`SUPPORTED_PROTOCOL_VERSIONS` in `packages/core/src/mcp/protocol.ts` lists `2025-06-18`, `2025-03-26` and `2024-11-05`, so `PROTOCOL_VERSION` — the newest we advertise — is `2025-06-18`.

The current MCP revision is **`2026-07-28`**, with `2025-11-25` in between. Real clients still accept `2025-06-18`, so nothing is broken today: a Zuke server negotiates and connects normally.

### Why it is worth doing anyway

The authorization guidance moved twice across those revisions, and the direction of travel matters for the work in #487:

- `2025-06-18` says an authorization server **SHOULD** support dynamic client registration.
- `2025-11-25` demotes that to **MAY** and introduces client id metadata documents.
- `2026-07-28` keeps the `MAY` and marks dynamic client registration **deprecated**.

So the revision we advertise is the one whose written guidance is furthest from what we now implement and document. That is a documentation hazard more than a runtime one, but it is the kind that sends a reader to the wrong spec page.

The method surface we implement (`initialize`, `tools/list`, `tools/call`, `ping`) is common to all of these revisions, which is why the list can carry several. The work is to check what the newer revisions add or change for a server of this shape — tool output schemas, elicitation, and whatever `2026-07-28` brought — decide what we support, and only then extend the list.

### Suggested approach

Read the two newer revisions against `packages/core/src/mcp/`, list the deltas that touch a server, and either implement them or state why they do not apply. Adding the version strings without that reading would be advertising a conformance we have not checked — the opposite of the point.

Worth doing after #487 lands, since that PR already establishes what the authorization half looks like.

totollygeek:
## What & why

The newest MCP revision this server offered was `2025-06-18`, two behind the current specification. It now offers `2025-11-25`, and still accepts every older revision it did before.

### Why not `2026-07-28`, which is what the issue asked for

Because it would be a false claim. `2026-07-28` is not a newer version of this protocol so much as a different one:

- it removes `initialize` and `notifications/initialized` entirely, and removes `ping`;
- it requires a new `server/discover` RPC that servers **MUST** implement;
- it carries the protocol version and client capabilities in per-request `_meta` instead of a handshake;
- it makes `resultType` mandatory on every result, and `ttlMs`/`cacheScope` mandatory on `tools/list`;
- it removes protocol-level sessions, replaces the GET endpoint, and requires `Mcp-Method`/`Mcp-Name` headers on every POST.

None of those are capability-gated features a server may decline while still claiming the revision — they are the base protocol. The specification's own compatibility matrix classifies a server of this shape as **legacy** and gives the outcome against a modern client in one word: *"Fails."* Supporting it means a second request path serving both eras concurrently, which the specification anticipates and calls dual-era. That is a project, not a version-string edit.

So the honest target is `2025-11-25`, the newest legacy-era revision, and a unit test pins the list against `2026-07-28` so the advertised version cannot drift ahead of the behaviour. That test is the actual deliverable of this issue as much as the bump is.

### `2025-11-25` asks nothing new of a tools-only server

Its additions are optional or already true here: icons on a tool, an experimental `tasks` capability, JSON Schema 2020-12 as the default dialect, an optional `description` on `serverInfo`, tool-name guidance we already satisfy. Its one new transport requirement — answering an invalid `Origin` with `403` rather than merely rejecting it — this transport has always done.

### The gap the research actually found

`MCP-Protocol-Version` was never checked. Since `2025-06-18` — the revision we **already claimed** — a client must send that header on every HTTP request after `initialize`, and a server **MUST** answer `400` when it names a revision the server does not implement. Nothing validated it, so a client could proceed on a protocol this server never agreed to, which is the one thing the header exists to prevent. That is a pre-existing conformance defect against the version already advertised, and it is fixed here.

An absent header stays fine. The specification's backwards-compatible reading is to assume `2025-03-26`, and since nothing here behaves differently across any supported revision, that assumption changes no behaviour and is not worth materialising.

## Verified against a real client

Not academic: the same client rejects a server advertising `2026-07-28` outright, which is how the "not a superset" conclusion stopped being a reading of the spec and became an observation. Against a running Zuke server on this branch it negotiates `2025-11-25` and reports connected.

Three findings, all real, all fixed.

**A repeated header was refused even when every value was supported** (medium). Repeated header lines arrive comma-joined and the check treated the join as one opaque token, so `2025-11-25, 2025-11-25` read as an unsupported revision. Not hypothetical: this transport documents that it expects to be fronted by a proxy, and a proxy re-adding a header the client already sent produces exactly that — a hard refusal for a correctly-versioned client, self-inflicted. Every value is checked now.

**The check answered before authentication** (low). Its message names every revision this build supports, which an unauthenticated caller has no business enumerating, and answering ahead of the bearer token contradicted the rule this module states for itself — that the metadata document is the only thing reachable before authentication. It has moved below the authenticator; the specification does not order the `400` against the `401`, and it still precedes reading the body. An integration test pins the ordering.

**One of my assertions was a tautology** (medium, hygiene). It compared `SUPPORTED_PROTOCOL_VERSIONS[0]` to `PROTOCOL_VERSION`, which is that constant's definition, so no edit to either file could have made it fail. It now asserts the property it was reaching for — that an unsupported request is answered with the newest revision — through `negotiateInitialize`.

The review also surfaced an unclaimed benefit, now documented: refusing the header is precisely what lets a dual-era client fall back correctly, since it probes with a modern request and retries as legacy on an unrecognised `4xx`. Ignoring the header would have had a modern request answered under the older semantics — the specific confusion the specification warns about.

Confirmed by the review and unchanged: the `2025-11-25` conformance claim itself, freedom from regressions (no test or code path pinned the old default; both version tests read the constant), and every factual assertion in the `2026-07-28` rationale.

## Testing

Unit coverage of the version list, the negotiation offer, and every branch of the header check including the duplicate-header case; integration coverage over real HTTP for the refusal, the accepted values, the absent header, and the guard's position relative to authentication. Both new tests were mutation-tested — reverting the source fails them.

`deno task ci`: 21/21 targets, 4284 tests, 98.5% lines and 97.3% branches against a 95% gate.

## Related issues

Closes #489

## Checklist

- [x] Docs updated in the same PR — the MCP guide's protocol notes and request-order list.
- [x] Public API regenerated — not applicable; the version list and the header helper are internal to the MCP modules.
- [x] Agent skills unchanged — protocol revisions are not part of the build-authoring surface — so no plugin version bump.

### DIFF
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -714,8 +714,10 @@ tool argument is masked in the [audit log](#audit-log) too.
   unauthenticated caller can enumerate. An absent header is fine: the
   specification says to assume `2025-03-26` then, and since nothing here
   behaves differently across these revisions that assumption changes nothing.
-  A repeated header (a proxy re-adding one the client already sent) is accepted
-  when every value it carries is supported.
+  A repeated header — a proxy re-adding one the client already sent — is
+  accepted when every copy agrees and names a supported revision; copies that
+  disagree are refused rather than resolved by picking one, since nothing here
+  could say which applies.
 - **Why not `2026-07-28`:** it is not a newer version of this protocol so much
   as a different one. It removes `initialize` and `ping` entirely, requires a
   new `server/discover` RPC, carries the protocol version and client
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -81,11 +81,14 @@ below), so one long `run:` never head-of-line-blocks another client's read.
   an authenticator runs on every bind, loopback included. The two compose: the
   token is a shared secret that gates the endpoint, the authenticator says _who_
   is calling.
-- **The order a request passes:** HTTP method (a non-`POST` gets `405`) →
-  `Origin` (`403`) → the static bearer token (`401`) → the
-  [authenticator](#authentication) (the refusal's own status) → the body (capped
-  at 1 MiB; over it, `413`). Authentication runs **before** the body is read, so
-  an unauthenticated caller never makes the server buffer its payload.
+- **The order a request passes:** the
+  [metadata document](#telling-a-client-where-to-authenticate) when one is
+  declared and the path matches (a public `GET`) → HTTP method (a non-`POST`
+  gets `405`) → `Origin` (`403`) → the static bearer token (`401`) → the
+  [authenticator](#authentication) (the refusal's own status) →
+  `MCP-Protocol-Version` (`400`) → the body (capped at 1 MiB; over it, `413`).
+  Authentication runs **before** the body is read, so an unauthenticated caller
+  never makes the server buffer its payload.
 - **Origin validation** guards against a browser drive-by / DNS-rebinding page:
   on a loopback bind, a request that carries an `Origin` header is accepted only
   when it is a loopback origin, and rejected `403` otherwise. A client that
@@ -706,9 +709,13 @@ tool argument is masked in the [audit log](#audit-log) too.
   `2025-03-26` and `2024-11-05`. A client's requested version is echoed when
   this server implements it, and otherwise answered with the newest — the
   client then proceeds or disconnects. Over HTTP, a `MCP-Protocol-Version`
-  header naming a revision this server does not implement is refused `400`; an
-  absent header is fine, since nothing here behaves differently across these
-  revisions.
+  header naming a revision this server does not implement is refused `400` —
+  after authentication, so the list of supported revisions is not something an
+  unauthenticated caller can enumerate. An absent header is fine: the
+  specification says to assume `2025-03-26` then, and since nothing here
+  behaves differently across these revisions that assumption changes nothing.
+  A repeated header (a proxy re-adding one the client already sent) is accepted
+  when every value it carries is supported.
 - **Why not `2026-07-28`:** it is not a newer version of this protocol so much
   as a different one. It removes `initialize` and `ping` entirely, requires a
   new `server/discover` RPC, carries the protocol version and client
@@ -721,6 +728,13 @@ tool argument is masked in the [audit log](#audit-log) too.
   means a second request path serving both eras, which is a project rather than
   a version-string edit. Until then, advertising it would be a claim this
   server cannot honour.
+
+  There is a side benefit to refusing the header rather than ignoring it. A
+  client that speaks both eras probes with a modern request and falls back to
+  `initialize` when it gets a `4xx` it does not recognise, so the `400` is
+  what makes that fallback work. Without it, a modern request would have been
+  answered under the older semantics — which is the specific confusion the
+  specification warns about.
 - **Errors:** a bad _tool_ call (unknown tool, unknown target, a failed run) is
   reported through the tool result (`isError: true`) so the model sees it,
   rather than as a transport-level error — matching the MCP convention. Typed
--- a/docs/mcp.md
+++ b/docs/mcp.md
@@ -702,6 +702,25 @@ tool argument is masked in the [audit log](#audit-log) too.
   `protocolVersion`), `notifications/initialized` (no reply), `ping`,
   `tools/list`, and `tools/call`. Unknown requests get a JSON-RPC
   `-32601 Method not found`; notifications never get a reply.
+- **Protocol revisions:** `2025-11-25` (the newest offered), `2025-06-18`,
+  `2025-03-26` and `2024-11-05`. A client's requested version is echoed when
+  this server implements it, and otherwise answered with the newest — the
+  client then proceeds or disconnects. Over HTTP, a `MCP-Protocol-Version`
+  header naming a revision this server does not implement is refused `400`; an
+  absent header is fine, since nothing here behaves differently across these
+  revisions.
+- **Why not `2026-07-28`:** it is not a newer version of this protocol so much
+  as a different one. It removes `initialize` and `ping` entirely, requires a
+  new `server/discover` RPC, carries the protocol version and client
+  capabilities in per-request `_meta` instead of a handshake, and makes
+  `resultType` on every result and `ttlMs`/`cacheScope` on `tools/list`
+  mandatory. Those are base-protocol requirements rather than capability-gated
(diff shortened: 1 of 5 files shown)


totollygeek:
### `1srs1e2t6caul` — "Comma-joined protocol header values are accepted as long as one entry is supported"

Acted on — `94761aa` — though the fix is neither what the finding proposed nor what it was reacting to.

**First, a correction to the title.** The body is accurate ("accepts the request if every trimmed value is supported", "a client or proxy that appends an extra unsupported version to a valid header will still be rejected"); the title says the opposite. The code accepts only when **every** value is supported, so `2025-11-25, 2026-07-28` was already refused. Worth separating, because the title reads as a bypass and there was not one.

**The substantive point is right, and it is the normalisation.** Accepting a multi-value header silently turns an ambiguous input into a definite one. That is a real objection even when every value is individually fine, and it is why I have not simply argued the finding away.

**But rejecting every repeated header, as the finding suggests, reintroduces a defect the previous review round fixed.** The two ways this header gets repeated want opposite answers:

- **Copies that agree** — `2025-11-25, 2025-11-25`. This is a proxy re-adding a header the client already sent, and the module documents a fronting proxy as the expected topology. There is no ambiguity to resolve: every copy says the same thing. Refusing it is a hard `400` for a correctly-versioned client caused entirely by its own infrastructure.
- **Copies that disagree** — `2025-11-25, 2025-06-18`. Nothing here can say which one applies, and a header whose meaning depends on which intermediary you ask is exactly the ambiguity a version check exists to remove.

So the rule is now: identical copies collapse to one value and are checked normally; conflicting copies are refused with their own message, rather than resolved by picking a winner. That keeps the proxy case working and removes the normalisation you object to. Neither "accept if all supported" nor "reject anything repeated" would have got both.

On request smuggling specifically: this header is a conformance signal, not an access control. Nothing routes, authorizes or dispatches on it — the method surface is identical across every revision in the supported list — so a disagreement between intermediaries about its value could not previously have changed what the server did. Refusing the conflict is worth doing because that may not always be true, not because it is exploitable today.

Tests cover all four shapes: agreeing copies accepted, conflicting copies refused, a single unsupported value refused and naming itself, and an absent header ignored.

totollygeek:
### `3fxvo85ucwsd4` — "Unsupported protocol version is rejected only after auth, enabling version enumeration through authenticated error paths"

Contested, and deliberately so — the placement is the fix for the opposite problem, found by an adversarial pass on the previous round of this same branch.

**The set of supported revisions is not a secret from an authenticated caller.** It is protocol data that the server hands over unprompted: `initialize` returns the negotiated `protocolVersion`, and negotiation exists precisely so a client can learn what the server speaks. A caller holding a valid token can read the version out of a successful handshake without sending a single malformed header. There is no enumeration here that a normal, correct client does not already perform as its first act.

**The previous placement was the actual leak, and it was to unauthenticated callers.** Before this, an anonymous request with a bad version header got a `400` naming every revision this build supports, while an anonymous request with a good one got `401`. That is a pre-auth fingerprint of the build — and it contradicted the rule this module states about itself, that the protected-resource metadata document is the *only* thing reachable before authentication. Moving the check below the bearer and authenticator guards removed that. An unauthenticated caller now gets `401` either way, and an integration test pins exactly that pairing so the ordering cannot drift back.

**The finding's own remedy would undo it.** "Move the check before any authenticated-only behavior" is the pre-auth placement, which discloses the same list to everyone rather than to token-holders. The alternative it offers — a generic refusal that does not name the revisions — costs a real diagnostic: a client that sent an unsupported version is precisely the client that needs to be told which ones would work, and it has authenticated to ask.

**On ordering as a specification question:** nothing requires the `400` to precede the `401`. The requirement is that a request naming an unimplemented revision must be refused, which it is. The check remains an O(1) comparison that still runs before the body is read, so nothing is spent on a request that will be refused.

I have left the behaviour as it is. If the concern is that the message is more specific than it needs to be, I would rather keep it: the audience is an authenticated client that has just made a version mistake, and telling it what to send instead is the whole value of the refusal.

totollygeek:
### `1iw19s7d0x3sl` — "Comma-joined repeated protocol headers are treated as a single unsupported version when any copy differs"

The finding offers two branches, and I have taken the second — `b10a062`.

**On the first branch ("tolerate a repeat by picking the first or the last hop's value"):** that is the behaviour the previous round removed on purpose, and I would not put it back. Picking a winner resolves an ambiguity nothing here can resolve — the server cannot know which intermediary is authoritative, and a header whose meaning depends on which hop you ask is exactly what a version check exists to reject. Note also that the strictness is narrower than the title suggests: copies that **agree** are accepted, which is the realistic proxy case. Only copies that **disagree** are refused, and that is a genuine conflict rather than a formatting quirk.

**On the second branch ("document this invariant clearly and add a test for the exact wire shape"):** the documentation was already there — `docs/mcp.md` states the rule in the protocol-revisions bullet, and the function's own JSDoc explains why the two cases diverge. The test was not, and that was a fair gap. The unit tests exercised a comma-joined *string*, which is the shape a duplicate header is supposed to arrive as — but nothing proved that two real header lines produce one. That assumption was doing load-bearing work in a guard whose whole justification rests on it.

There is now an integration test that sends both copies over a raw socket, since `fetch` cannot send a header twice. The conflicting case is what makes it conclusive: it comes back `400` naming more than one revision, which is only reachable if the two separate lines were joined by the header parser and split again by this code. Agreeing copies return `200` over the same path.

**On "a client that sends a quoted/comma-containing value":** a protocol revision is a `YYYY-MM-DD` date, so there is no legitimate value containing a comma or requiring quoting. A value that did contain one would be an unsupported revision string and is refused as such — correctly, since it is not a version this server implements under any reading.

Coverage now spans the four shapes that exist: absent, a single supported value, a single unsupported value naming itself, agreeing copies at the wire level, and conflicting copies at the wire level.

Issue #505, opened 2026-09-08
bug: a timed-out run leaves unstarted targets pending, where a failed run skips them


totollygeek:
### What happened?

The two paths that fail a run disagree about what happens to a target the run never started.

The scheduler settles them. When a run ends `failed` with gates still parked, `packages/core/src/scheduler.ts:1004-1013` settles every unstarted target `skipped`, on the stated reasoning that "a run that also failed will never resume, so settle them `skipped` — otherwise they linger `pending` inside a terminal `failed` record that no resume sweep reaches (F8)".

The timeout fast path does not. `failTimedOut` (`packages/core/src/resume.ts`) settles the run directly, without the scheduler, so a target blocked behind the gate stays `pending` on a `failed` record — exactly the state F8 set out to remove.

### Expected behaviour

A `failed` record should say the same thing about an unstarted target however it got there. Either both paths settle it `skipped`, or the F8 rule is wrong and neither should.

### Minimal reproduction

A build with two independent gates and a target depending on both; `gateA` has `.timeout(0)`, `gateB` has `.timeout("1h")`. Run it, then `resume --check` so gateA's deadline fires.

### Relevant logs or output

    status: failed
    {"gateA":{"status":"failed","error":"wait \"gateA\" timed out (deadline ...)"},
     "gateB":{"status":"skipped","endedAt":"..."},
     "all":{"status":"pending"}}

`all` was never started and never will be, on a terminal record.

### Affected package(s)

@zuke/core

### Additional context

Found while fixing #500 and deliberately left out of its scope: that issue is about `waitingFor` outliving the `waiting` status, and #500's fix does settle the sibling *gates* on this path. Whether an unstarted `pending` target should read `skipped` is a separate semantic question, and it is worth answering deliberately rather than as a side effect.

Note the cancel path leaves `pending` rows `pending` too (`finalizeCancelled` touches only `waiting` rows), and that is long-standing behaviour a reader can already see. So the question is really three-way: what should a terminal record say about a target it never reached, on each of the failed, timed-out, and cancelled paths? Today the answer differs across all three.

PR #507, opened 2026-09-08
fix(core): settle a target's wait whenever its run settles

### LINKED ISSUE #500: bug: a settled target keeps its waitingFor, and a cancelled run keeps a waiting target
### What happened?

Two paths settle a run without settling the targets that were parked on a wait, so a **terminal run is left carrying a live waiter**. Only `RunStateWriter.markTargetSettled` (`packages/core/src/state/writer.ts:262`) clears `waitingFor`, and neither path goes through it.

**(a) A timed-out wait leaves a stale `waitingFor` on a failed target.** `failTimedOut` (`packages/core/src/resume.ts:453-458`) writes the target row by hand — `status`, `error`, `endedAt` — and never removes the wait it just gave up on.

**(b) A cancelled run leaves its waiting target claiming to wait.** `finalizeCancelled` (`packages/core/src/cancel.ts:985-996`) writes only `record.status`; it never touches target rows. A run cancelled while suspended ends `cancelled` with its gate still `status: "waiting"`. This one is not merely a stale field — the target's own status is wrong.

Forcing a waiting target is **not** affected: it settles through `markTargetSettled` on the next resume, and the row comes out clean.

This is item 14.1 of the maintainer proposal, widened by one path: the proposal names the stale field, and (b) is the same defect one step further along.

### Expected behaviour

A target that has settled — by any path — carries no `waitingFor`, and a run that has reached a terminal status has no target still claiming to be waiting on something.

### Minimal reproduction

Run a build whose only target is a gate, then drive each path through the CLI and read the persisted record back:

- **(a)** a gate with `.timeout(1)`, then `resume --check` once the deadline has passed.
- **(b)** a gate with `.timeout("1h")`, then `cancel <run-id>` while it is suspended.

Both were driven end to end through `runCli` against a temporary `ZUKE_STATE_DIR`.

### Relevant logs or output

Path (a), the persisted `gate` row after the timeout:

    {
      "status": "failed",
      "endedAt": "2026-09-08T10:35:31.178Z",
      "error": "wait \"gate\" timed out (deadline 2026-09-08T10:35:31.170Z)",
      "waitingFor": {
        "trigger": "signal:never",
        "onTimeout": "fail",
        "deadline": "2026-09-08T10:35:31.170Z"
      }
    }

Path (b), after `cancel`:

    CANCEL RUN STATUS: cancelled
    CANCEL TARGET: {"status":"waiting","meta":{},"waitingFor":{"trigger":"signal:never","onTimeout":"fail","deadline":"..."}}

### Affected package(s)

@zuke/core

### Additional context

**(b) is observable, not cosmetic.** `renderRunDetail`/`targetNote` (`packages/core/src/runs.ts:317`) has no run-status guard, so `zuke runs show` on a cancelled run prints the gate as still waiting:

    status:   cancelled
    Targets:
      ~ gate  waiting  waiting for signal:never (deadline 2026-09-08T11:48:49.397Z)

**(a) is narrower but still contract-visible.** `expiredWait` requires the run to be `suspended`, so the stale field cannot re-fire a timeout; the harm is that the record — the artifact every external consumer of the state contract reads — says a failed target is parked on a gate. `priorWaitsOf` (`packages/core/src/execute_state.ts:53-60`) collects it too, on a comment that promises "the still-waiting targets".

**Proposed shape.** The clearing rule has exactly one implementation today and three call sites that need it, which is the shape AGENTS.md guideline 12 asks for: lift it into one unexported helper under `packages/core/src/state/`, used by `markTargetSettled`, `failTimedOut`, and a new sweep in `finalizeCancelled` that settles any still-`waiting` target to `skipped` — the vocabulary a cancel already uses for a target it never ran (see the `disown()` comment in `writer.ts`). The cancel sweep also applies when the terminal status is `failed`, which is how a sweep reaps an abandoned run.

The record change on cancel is visible, so `docs/state.md` and the state contract need updating in the same PR.

totollygeek:
### The problem

`RunStateWriter.markTargetSettled` was the only place that cleared a target's `waitingFor`. Every other path that settles a run writes the record by hand, and none of them dropped the wait — so a **terminal run was left carrying a live waiter**.

Two ways to see it, both driven end to end through the CLI:

- A timed-out gate ended `failed` still holding the trigger and the deadline it had just given up on.
- A run cancelled while suspended ended `cancelled` with its gate still `status: "waiting"` — not a stale field but a wrong status, which `zuke runs show` prints as a target parked on a deadline nobody is watching.

Forcing a waiting target was never affected: it settles through the writer, and the row comes out clean.

This is item 14.1 of the maintainer proposal, widened by one path — the proposal names the stale field, and the cancelled run is the same defect one step further along.

### The shape

The clearing rule now has exactly one implementation, `settleTargetRow` in a new unexported `packages/core/src/state/settle.ts`, used by the writer, the timeout fast path, and both cancellation paths. Its companion `settleWaitingTargets` sweeps a terminal record's still-parked gates to `skipped` — the vocabulary the scheduler already uses when a failed run leaves a gate nobody will resume.

One ordering constraint is load-bearing and now written down where it can be read: the compensation walk treats a `waiting` target as **unproven** and unwinds it, where a settled one it does not. Both cancellation paths therefore sweep **after** their walk. Sweeping earlier — in `transitionToCancelling`, say — would silently stop compensating parked gates on degraded records, with nothing failing to say so.

### What the adversarial review found

Three reviewers attacked the first version across cancel semantics, races, and the published contract. Every finding was reproduced against the running code before being fixed, and each fix is mutation-checked — removed the fix, watched the test fail, restored it. Five survived:

**The invariant this change documents was false for Ctrl-C** (high). The sweep went into `finalizeCancelled`, which is the *out-of-process* `zuke cancel`. The executor's own cancellation settles through `writer.markRunCancelled` instead, and the scheduler's existing sweep for this is gated on `anyFailed && anyWaiting`, so a cancellation with no failure never enters it. An aborted run therefore still produced exactly the record the new docs said cannot exist. This is the finding I would least have wanted to ship: the doc paragraph would have been a published promise the code did not keep. The scheduler's own comment already claimed the cancelled case was handled, and it was not.

**A `cancel-run` timeout erased the evidence that a deadline was missed** (high, and a regression this change introduced). The `fail` disposition records the miss in the target's error. The `cancel-run` disposition routes through the cancellation, whose sweep could not tell an expired gate from an unreached one — so the gate settled with nothing, and the terminal record could no longer tell a missed deadline from an operator typing `zuke cancel`. The expired wait's name and message are now threaded through, and the gate keeps them.

**Only the first expired wait was settled** (high). A run can park several gates at once. `failTimedOut` failed the one whose deadline passed and left its siblings `waiting` on a record that is about to be terminal and will never be resumed again.

**The writer put `waiting` back onto a terminal record** (medium, pre-existing). On a CAS conflict the writer deliberately re-applies target-level mutations onto the record another process settled, so a just-settled `succeeded` is not lost. Parking a gate is not such a mutation: a `zuke cancel` racing a process that parks a gate right then produced a row that was `waiting` and `endedAt` at once. Refused now, matching the guard `markRunSuspended` already had.

**`runs show` printed a row's error only when its status was `failed`** (medium). The timeout reason recorded above would have been stored where no reader could see it.

Also: the timeout message existed as two near-copies in `resume.ts`, which is guideline 12's exact shape, and is one function now; and `priorWaitsOf` collected `waitingFor` from every row regardless of status — harmless today, but it was the one reader assuming the invariant rather than checking it, on a comment promising "the still-waiting targets".

### Deliberately out of scope

A timed-out run leaves unstarted targets `pending` where a failed run settles them `skipped`, and a cancelled run leaves them `pending` too. That is a three-way disagreement about what a terminal record says about a target it never reached, and it deserves an answer of its own rather than a side effect of this one. Filed as #505.

### One thing worth a reviewer's eye

`CancelOptions` gains an optional `expiredWait`, so this adds public surface in a change titled `fix`. It is internal plumbing — the only caller is the timeout disposition — and the user-visible change is entirely bug fixes, so a patch bump reads right to me. Say so if you would rather it were a `feat`.

### Testing

15 new tests. Unit coverage of the helper and the sweep, including the expired-wait message; writer-level regressions for the in-process cancellation and the settled-elsewhere re-park; and integration tests driving real builds through the CLI for each of the four record-level defects, one of them asserting the `runs show` output an operator actually reads.

`deno task ci`: 21/21 targets, 4304 tests, 98.5% lines and 97.4% branches against a 95% gate.

### Related issues

Closes #500

### Checklist

- [x] Docs updated in the same PR (the durable state guide and the orchestration guide).
- [x] No agent-skill change, so no plugin version bump is due.


totollygeek:
Finding `384ktwabhadq4` — refuted, and I think the reviewer has the shape backwards.

`settleWaitingTargets` has **one** implementation, in `packages/core/src/state/settle.ts`, and **three** call sites: `RunStateWriter.markRunCancelled` (`writer.ts:347`), `finalizeCancelled` (`cancel.ts:1015`), and `failTimedOut` (`resume.ts:480`). The whole point of the change is that the clearing rule used to be inline in one place and forgotten in the others; lifting it into a shared unit imported by every settler is the fix, not the duplication.

On "shared at the write boundary": there is no single write boundary to share it at, and the two are not interchangeable. `finalizeCancelled` and `failTimedOut` write with `store.putRun(next, version)` in their own compare-and-swap loops — they run in a process that does **not** own the run. `markRunCancelled` writes through the writer's `#applyAndPersist`, which carries the run lease, the `disown()` one-way switch, and the settled-elsewhere re-apply. Routing the external paths through the writer would hand an outside canceller ownership semantics it must not claim; routing the writer's path through a bare `putRun` would discard the lease checks that stop a process writing onto a run it has lost. Two write mechanisms, deliberately distinct, sharing the one mutation rule between them — which is exactly what a shared helper is for.

There is a real ordering constraint here, and it is the reason the sweep sits at each settler rather than at a single boundary: the compensation walk treats a `waiting` target as unproven and unwinds it, where a settled one it does not. Both cancellation paths therefore sweep **after** their walk. A single earlier choke point — `transitionToCancelling`, say — would silently stop compensating parked gates on degraded records. That is now stated in the helper's doc comment so the next reader does not have to rediscover it.

On the "docs/test drift risk" in the summary: the docs claim was in fact false when first written — the adversarial pass caught that the in-process cancellation had no sweep — and `markRunCancelled` was added in response, with a regression test driving a real aborted run. If there is a specific drift still standing, name it and I will look.

totollygeek:
Finding `3brbj32dc67d6` — taken, fixed in `8403e98`.

The concern was right in shape even though it could not fire for the only caller: `expired.target` comes from `expiredWait(record, …)`, which returns a key of that same `record.targets`, so the match could not fail today. But tying the diagnostic to catching the row mid-sweep is the wrong coupling, and it is the failure mode I have hit repeatedly on this branch — an explicit declaration that silently does nothing.

The stamp is now a direct lookup after the sweep rather than a name comparison inside it. The reason belongs to the run, so recording it no longer depends on that row still being `waiting` when the sweep reaches it. A caller naming a target the record does not have still gets nothing rather than a throw, deliberately: a cancellation is the wrong moment to fail over a diagnostic. Mutation-checked — removing the stamp fails two tests.

Separately, and thank you for the pointer that led me to look: `packages/core/tests/zz_attack_conc_test.ts` was a review probe harness that a `git add -A` committed by accident. It is gone. Your verifier was correct that its cases are real `Deno.test`s with real assertions, which is exactly why it was not obviously wrong in the diff — but it is a scratch file, not a designed regression test, and the defects it was probing have deliberate tests in `settle_waiting_test.ts` and the wait integration suite.

Gate after both changes: 22/22 targets, 4310 tests, 98.5% lines / 97.3% branches.

todor-a (approved):

PR #518, opened 2026-09-08
feat(core): let a parameter declare its CLI flag

### LINKED ISSUE #517: feat: let a parameter declare its CLI flag, and document the derivation rule
### Problem

A parameter's CLI flag and environment variable are derived from its field name, and the author cannot override the flag. The rule inserts a dash at every lower-to-upper transition, which reads badly whenever a name contains an initialism ending in a digit:

    skipE2E     -->  --skip-e2-e     SKIP_E2_E
    runE2E      -->  --run-e2-e      RUN_E2_E
    useHTTPS    -->  --use-https     USE_HTTPS
    apiURL      -->  --api-url       API_URL

`useHTTPS` and `apiURL` come out fine because a run of capitals has no internal transition. `skipE2E` does not: the digit ends a run, so `2E` is a transition and the flag becomes `--skip-e2-e`. An author who wants `--skip-e2e` has no way to ask for it, and the rule that produced the surprising spelling is not documented anywhere.

There is already an explicit override for the other half of the pair — a parameter can set its environment variable name — so the flag is the asymmetric one.

### Proposal

Add the symmetric setter, and document the rule and its pitfall.

    skipE2E = parameter("skip the E2E suite").flag("--skip-e2e")

The declared spelling **replaces** the derived one everywhere the flag appears: argument parsing, `--help`, `--list --json`, shell completions, and the registry descriptor. Only the declared spelling is accepted, since the point is that the author chooses it.

### Acceptance

- A parameter with a declared flag is set by that flag and rejects the derived one.
- The declared flag appears in `--help`, in the JSON build surface, and in generated completions.
- A run spawned through the registry MCP server can set the parameter, which means the descriptor carries the declared flag.
- The derivation rule and the initialism pitfall are documented, with the workaround for authors who would rather not declare a flag at all.

### Note on the descriptor guard

This interacts with the guard added in #486, which refuses a registry descriptor whose `flag` differs from the flag its `name` would derive. A declared flag legitimately differs, so that rule cannot stand as written.

The threat it was added for is narrower than the rule: the CLI matches a built-in flag ahead of a declared parameter, so a descriptor naming a built-in — `--actor` above all — would put a caller's value on the built-in instead and forge a run's attribution. That is caught by the built-in check next to it, which stays.

The derived-equality rule was the belt-and-braces half. It is replaced by a uniqueness check: no two parameters in a descriptor may claim the same flag, so one cannot shadow another. Both checks are at the same untrusted-input boundary as before.

### Package

@zuke/core

### Size

S

totollygeek:
### The problem

A parameter's flag and environment variable are derived from its field name: one dash, or underscore, at each lower-to-upper transition. A run of capitals has no internal transition and stays together, so `apiURL` gives `--api-url`. A **digit ends a run**, which is where the rule surprises people:

    skipE2E   -->  --skip-e2-e   SKIP_E2_E
    runE2E    -->  --run-e2-e    RUN_E2_E

An author who wanted `--skip-e2e` had no way to ask for it, and the rule that produced the surprising spelling was documented nowhere. The environment variable already had an explicit override, so the flag was the asymmetric half.

This is item 14.3 of the maintainer proposal.

### The shape

A declared flag **replaces** the derived one, everywhere a flag is rendered or matched: the parser, `--help`, `--list --json`, shell completions, resolution errors, the interactive prompt, and a registered build's descriptor. A single accessor is the only way any of them asks, so a declared flag cannot be honoured in one place and derived in another.

Three refusals at load, each naming the field: a flag that is not a valid flag name, one that names a built-in, and two parameters claiming one flag. The second is worth noting because the old name-based check could not reach it — a parameter named `message` declaring `--actor` renders as `--message`.

The environment variable is derived separately and is unaffected.

### Re-basing the descriptor guard, and the two holes that opened

The change could not land without touching the guard added in #486, which refused any registry descriptor whose `flag` differed from what its `name` derives. A declared flag legitimately differs, and the descriptor has to carry it or a registry-spawned run could not set the parameter at all.

I argued that the rule was belt-and-braces the built-in check covers on its own. That was reasoning about the **threat** rather than about what the rule mechanically guaranteed, and it was wrong twice. Both were found by attacking this change before opening the PR, both are fixed here, and both have regression tests that were mutation-checked.

**A descriptor flag containing an equals sign forged the run's actor** (high). The child's parser splits a flag from its value at the first `=` and matches the prefix, so a flag of `actor=mallory` is absent from the built-in list yet arrives as `--actor`. A malicious registry advertises a plausible `environment` parameter, marks it required so a conforming MCP client fills it in, and the caller's own value overrides the actor the runner exported for the authenticated caller — the exact attribution forgery #486 existed to stop, reached through a parameter name that looks innocent. The removed rule had been supplying the missing shape guarantee incidentally, because a *derived* flag can never contain an equals sign. Shape is now checked before membership, by one regex shared between the two boundaries: where a build declares a flag, and where a descriptor is read.

**Two parameters sharing a name shadowed each other** (medium-high). The removed rule gave name uniqueness away for free, since two entries with one name derive one flag. The registry server advertises a parameter by name and spawns it by flag, each through a last-wins map, so a descriptor naming `env` twice advertises one `env` property and puts the caller's value on the other entry's flag. Both keys are checked now, in one place, for the reason each is read by a different consumer.

Also from the same review, three tests that were weaker than they looked: two assertions that passed on the rendered flag alone rather than on the field names the message is supposed to carry, an equality assertion that any collapsing transform would satisfy, and a round-trip fixture that never exercised the new property it was guarding.

### Testing

15 new tests. Unit coverage of the accessor, the dash-stripping, and each of the three load-time refusals; descriptor coverage of the shape rule, the built-in rule, both uniqueness rules, and a declared flag surviving the round trip; and integration tests driving real builds through the CLI for the declared flag being accepted, the derived one being rejected, `--help` and the JSON surface showing the declared spelling, and the environment variable being untouched.

`deno task ci`: 22/22 targets, 4357 tests, 98.4% lines and 97.4% branches against a 95% gate.

### Related issues

Closes #517

### Checklist

- [x] Tests were added or updated; coverage stays at 95%+.
- [x] Docs updated in the same PR (the parameters guide).
- [x] Agent skills updated and the plugin version bumped in all four manifests (0.47.0 to 0.48.0).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
       "name": "zuke",
       "source": "./plugins/zuke",
       "description": "Agent skills for Zuke — scaffold Zuke into a project (zuke-setup) and write or edit a zuke.ts build (zuke-write-build).",
-      "version": "0.47.0",
+      "version": "0.48.0",
       "author": {
         "name": "Zuke",
         "email": "contact@zuke.build",
--- a/docs/parameters.md
+++ b/docs/parameters.md
@@ -149,8 +149,42 @@ has resolved parameters (e.g. at construction time) throws.
 
 The flag and environment variable are derived from the property name:
 `environment` → `--environment` / `ENVIRONMENT`; a camelCase name like
-`targetEnv` → `--target-env` / `TARGET_ENV`. Override the environment variable
-with `.env("NAME")`.
+`targetEnv` → `--target-env` / `TARGET_ENV`. A nested group joins with the same
+separator, so `runs.limit` → `--runs-limit` / `RUNS_LIMIT`.
+
+The rule is one dash (or underscore) at **each lower-to-upper transition**. A
+run of capitals has no internal transition and so stays together — but a digit
+ends a run, which is where the rule surprises people:
+
+| Property name | Flag           | Environment variable |
+| ------------- | -------------- | -------------------- |
+| `apiURL`      | `--api-url`    | `API_URL`            |
+| `useHTTPS`    | `--use-https`  | `USE_HTTPS`          |
+| `skipE2E`     | `--skip-e2-e`  | `SKIP_E2_E`          |
+| `runE2E`      | `--run-e2-e`   | `RUN_E2_E`           |
+
+`skipE2E` gives `--skip-e2-e` because the `2` ends the run of capitals, making
+`2E` a transition of its own. Two ways out, and the first is often enough:
+
+- **Name it so the rule agrees with you.** `skipE2e` → `--skip-e2e`.
+- **Declare the flag.** `.flag("--skip-e2e")` sets the spelling explicitly, and
+  the leading dashes are optional.
+
+```ts
+class CI extends Build {
+  skipE2E = parameter("skip the E2E suite").flag("--skip-e2e").boolean();
+}
+```
+
+A declared flag **replaces** the derived one: only it is accepted on the
+command line, and it is what `--help`, `--list --json`, shell completions and a
+registered build's descriptor all show. It must be lowercase letters, digits
+and dashes starting with a letter, it may not be a built-in flag (see below),
+and no two parameters may claim the same one — each is a `ParameterError` when
+the build loads, naming the field.
+
+The environment variable is derived separately and is unaffected by a declared
+flag; override that half with `.env("NAME")`.
 
 ### Names a parameter may not use
 
@@ -173,8 +207,9 @@ naming the field:
 
 Only the rendered flag matters, so a longer or nested name is fine: `actorName`
 → `--actor-name` and a grouped `runs.limit` → `--runs-limit` both stay usable.
-Rename the field; there is no opt-out, because the alternative is a flag that
-silently belongs to something else.
+Rename the field, or declare a different flag with `.flag("--…")`. What is not
+available is keeping the colliding spelling, because the alternative is a flag
+that silently belongs to something else.
 
 ## Without the CLI
 
--- a/gemini-extension.json
+++ b/gemini-extension.json
@@ -1,5 +1,5 @@
 {
   "name": "zuke",
-  "version": "0.47.0",
+  "version": "0.48.0",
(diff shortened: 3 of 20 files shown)


totollygeek:
Finding `1gtczdcdfzwlq` — refuted, with the four cases driven through the real parser.

The finding anchors on `packages/core/src/registry/descriptor.ts:286`, which is inside `parseParameterInfo` — the **per-parameter** validator. It cannot check duplicates, because a duplicate is by definition a property of the set, not of one entry. That check lives one level up in `parseCliDescription`, where `assertDistinctParameters` wraps the `arrayOf(...)` that produces the list, and it covers both keys: the flag and the name.

Driven against the real `parseBuildDescriptor`:

    refused    duplicate flag, distinct names -> parameters "env" and "tag" both claim the flag "--env"
    refused    duplicate name, distinct flags -> descriptor declares two parameters named "env"
    refused    malformed flag with "="        -> declares the flag "--actor=x", which is not a valid flag name
    ACCEPTED   legit declared flag            -> [["skipE2E","skip-e2e"]]

Only the last is accepted, which is the point of the change.

On "ownership source": a descriptor comes from a registry backend Zuke does not control, so there is no source to verify a flag against — the build that declared it is not present at the read path, and the descriptor is the only artifact. That is exactly why the guard is built from what the **child parser** does rather than from provenance: shape first, so a flag cannot mean something other than it says; then built-in membership; then uniqueness on both keys. Both duplicate shapes and the malformed shape were live holes in the first version of this branch and each has a regression test that was mutation-checked — removing either refusal fails the suite.

Worth recording that the reviewer summary's instinct was sound even though the finding is not: this diff did introduce exactly the class of problem it describes. The `=` bypass forged a run's actor and the duplicate-name bypass shadowed a parameter, both because relaxing the #486 rule removed guarantees it had been supplying incidentally. They were found by an adversarial pass before the PR and are fixed in the commit under review, which is why they no longer reproduce.

PR #548, opened 2026-09-10
fix(core): stop build output forging workflow commands, redact the job summary

### LINKED ISSUE #546: bug: build output can forge GitHub Actions workflow commands, and the job summary skips the redactor
Two findings from the whole-repository security assessment, both on the path
between a build and the GitHub Actions runner.

## Workflow commands can be forged from build output

Under GitHub Actions the runner reads every line a step writes, on stdout and
stderr alike, and acts on two syntaxes: a line whose first non-blank characters
are `::`, and the legacy `##[command]` form, which it recognises **anywhere** in
a line.

`escapeData` and `escapeProperty` in `packages/core/src/report.ts` exist for
exactly this, and their documentation names the threat precisely — a subprocess's
stderr carrying `::stop-commands::`. But they are applied only to the synthesized
`::error::` annotation and the `::group::` header. Everything else is printed
raw:

- the failure footer's `detail` line, which is `messageOf(error)` and therefore
  embeds a failed subprocess's stderr verbatim;
- the pass, wait and dry-run footers, the summary table rows and the closing
  line, which embed the target name — a fan-out sub-target's name carries its
  item key, derived from repository or remote data;
- the forced-override line, whose actor and reason come from the shared state
  store.

The ordering makes the first one worse: `detail` is emitted *before* the escaped
annotation, so a `::stop-commands::` in a tool's stderr suspends command
processing and disarms the very annotation the escaping protects, along with the
`::endgroup::` after it.

Two details make the obvious defences wrong. The runner trims leading whitespace
before testing for `::`, so the two-space indent on the detail line is not a
mitigation. And `escapeData` is the wrong tool for text printed as itself: it
encodes every newline, which folds a multi-line compiler dump into one
unreadable `%0A`-joined line.

Impact is confined to the Actions log and annotation stream: forged or
suppressed annotations, hidden log sections, and suppression of Zuke's own
`::add-mask::` directives. It does not change what ran.

## The job summary bypasses the secret redactor

`writeJobSummary` in `packages/core/src/execute_output.ts` serialises the summary
notes straight into `GITHUB_STEP_SUMMARY` without the run's `Redactor`. Every
other sink is redacted — the console through the reporter, the run record as it
is persisted — so a summary note whose value contains a resolved `secret()`
parameter reads `[redacted]` on the terminal and in the record, while the job
summary publishes it in the clear to everyone who can view the workflow run.

The `::add-mask::` directives do not cover this: they mask the runner's log
stream, not a file the build writes. This contradicts the guarantee in
`docs/secrets.md`, that a secret never surfaces in Zuke's own output, a summary
included.

## Acceptance criteria

- One escaping helper, on a published entrypoint so `@zuke/console` can share it
  rather than growing a second copy, which neutralises both runner syntaxes
  while leaving ordinary output byte-identical and readable.
- Every GitHub-mode emission of a target name, failure message, actor or reason
  routed through it; Zuke's own commands keep working.
- `writeJobSummary` takes the run's redactor and applies it.
- Unit coverage per emission site plus an integration test asserting over the
  whole stream of a real build, so a future emission site is caught rather than
  trusted; and a test that a secret-valued summary note is masked in the summary
  file, which has no coverage today.

totollygeek:
Closes #546.

## Two findings on the path between a build and the runner

**Build output could forge workflow commands.** The runner acts on two syntaxes in a step's output: a line whose first non-blank characters are two colons, and the legacy bracket form, honoured anywhere in a line. The escapes existed but reached only the synthesized annotation and the group header. Everything else went out raw — the failure footer's detail line, which embeds a failed subprocess's stderr verbatim; the pass, wait and dry-run footers and the summary table, which embed a target name that for a fan-out carries an item key from repository or remote data; and the closing line's culprit.

Ordering made the first one worst: the detail is printed **before** the annotation, so a suspend directive in a tool's stderr disarmed the very annotation the escaping protects, and the group close after it.

**The job summary bypassed the redactor.** Every other sink is masked — the console through the reporter, the run record as it is persisted — so a summary note holding a resolved secret parameter read redacted on the terminal and in the record while the summary published it in the clear to anyone who could view the run. The mask directives do not cover this: they mask the runner's log stream, not a file the build writes. It contradicted the guarantee in the secrets guide.

## The shape of the fix

The command-body escape is the wrong tool for text printed as itself: encoding every newline folds a compiler dump into one unreadable line. So there is a second helper that alters only what would actually be interpreted and leaves everything else byte-identical, applied to the **value** rather than the finished line, so an embedded newline cannot smuggle a marker onto a line of its own. Both escapes live in one module, published from the render entrypoint every renderer already imports.

A sink-level guard was considered and rejected: at the sink, Zuke's own group command and a forged one are identical strings, so neutralising there would break the tool's own output.

## Sites the issue did not name

Writing the tests turned up three more, each verified with a probe before being accepted:

- **The console package** interpolated straight into its own warning and error commands with no escaping at all — and that body carries a subprocess's stderr too.
- **The forced-override line** is printed by the scheduler through the reporter rather than the report module, so wiring that module alone left it raw. Its actor and reason come from the shared state store.
- **The force command's echoed reason**, which in a scripted force is often interpolated from a pull request title or an issue body.

## Verification

Both original defects were reproduced before and after. A suspend directive in a failure message reaches the log as a live command without the fix and disarmed with it; a secret-valued summary note is written in the clear without it and redacted with it.

Every new test was mutation-checked, and three earlier versions of them were thrown away for failing that check. One asserted on a name carrying a marker with no newline, which is harmless because it lands mid-line. One asserted on the resumed run rather than the force, where the forced target settles before anything about it is printed. Both passed with the escaping removed, which is the only reason they were caught. The surviving tests each fail when the helper is neutered.

That last test also caught a real defect in the first attempt at the fix: it used the legacy host detector, whose values are host-specific names, so the comparison against the provider name could never be true and the escape never ran.

## Testing

Unit tests per escape, including that ordinary output is byte-identical and that a multi-line dump keeps its structure. Integration tests assert over the **whole stream** of a real build rather than one line, so an emission site added later is caught rather than trusted. Gate green at 23/23: 4448 tests, 98.5% lines, 97.4% branches.


totollygeek:
Superseded. #547 landed the same fix for the core reporting path while this was in review, and its `escapeLine` is stronger than the version here — it handles `\r`, `\r\n` and NEXT LINE (U+0085), which this branch's newline-only split missed.

What remained unique here were the sites #547 explicitly deferred: the console package's own warning, error and group commands, and the force command's echoed reason. Those are rebased onto #547 and carried forward, so nothing found here is lost.

PR #550, opened 2026-09-10
fix(console): escape the two workflow-command sites #547 left for later

totollygeek:
Refs #546. Follow-up to #547, which fixed the core reporting path and published the escapes "so that package can reuse them later".

Two sites still emit raw, and both carry text the job's own author does not control.

## The console package's own commands

`@zuke/console` interpolates straight into its `::warning::` and `::error::` commands and into its group header, with no escaping. The error body is `messageOf(options.error)` — for a failed command, that subprocess's stderr verbatim, which is the same untrusted source the core fix was written for.

These are command **bodies** rather than lines printed as themselves, so they take `escapeData` rather than `escapeLine`: a newline inside a command ends it, and what follows opens a new one.

## The force command's echoed reason

`zuke force` echoes the operator's reason back to the log. In a scripted force that reason is commonly interpolated from a pull request title or an issue body, so it is not the job author's own text. It is escaped where it is embedded, so it stays covered whichever way the message is later printed.

## On the overlap with #547

This branch originally carried its own version of the core fix as well. #547 landed first, and its `escapeLine` is the better implementation — it splits on `\r`, `\r\n` and `\n` and trims NEXT LINE (U+0085), where the version here split on `\n` alone and would have let a `\r`-separated marker through. That work is dropped rather than merged; only the two uncovered sites are carried forward, using #547's helpers.

## Verification

Every test here was confirmed to fail with its own escape removed, then to pass with it restored.

The force test asserts on the **force command's own output**, not on a later run. An earlier version watched the resumed run and passed with the escaping removed, because the forced target settles before the scheduler prints anything about it — the stream it watched never carried the value at all.

Gate green at 23/23: 4453 tests, 98.5% lines, 97.4% branches.

### DIFF
--- a/deno.lock
+++ b/deno.lock
@@ -481,8 +481,7 @@
       ]
     },
     "@xmldom/xmldom@0.8.13": {
-      "integrity": "sha512-KRYzxepc14G/CEpEGc3Yn+JKaAeT63smlDr+vjB8jRfgTBBI9wRj/nkQEO+ucV8p8I9bfKLWp37uHgFrbntPvw==",
-      "deprecated": true
+      "integrity": "sha512-KRYzxepc14G/CEpEGc3Yn+JKaAeT63smlDr+vjB8jRfgTBBI9wRj/nkQEO+ucV8p8I9bfKLWp37uHgFrbntPvw=="
     },
     "agent-base@7.1.4": {
       "integrity": "sha512-MnA+YT8fwfJPgBx3m60MNqakm30XOkyIoH1y6huTQvC0PwZG7ki8NacLBcrPbNoo8vEZy7Jpuk7+jMO+CUovTQ=="
@@ -1593,7 +1592,7 @@
       },
       "packages/console": {
         "dependencies": [
-          "jsr:@zuke/core@^1.25.0"
+          "jsr:@zuke/core@^1.51.0"
         ]
       },
       "packages/cspell": {
--- a/packages/console/deno.json
+++ b/packages/console/deno.json
@@ -7,7 +7,7 @@
     ".": "./mod.ts"
   },
   "imports": {
-    "@zuke/core": "jsr:@zuke/core@^1.25.0"
+    "@zuke/core": "jsr:@zuke/core@^1.51.0"
   },
   "publish": {
     "exclude": [
--- a/packages/console/src/console.ts
+++ b/packages/console/src/console.ts
@@ -27,6 +27,7 @@ import {
   box as renderBox,
   type BoxOptions,
   detectWidth,
+  escapeData,
   line as renderLine,
   type LineOptions,
   type Style,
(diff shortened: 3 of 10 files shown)


totollygeek:
**Blocked on a core release, not on a defect.**

`coreFloorCheck` fails because the console half uses `escapeData` from `@zuke/core/render`, which #547 added but core has not published yet — 1.50.0 is on JSR and does not carry it:

```
TS2305: Module 'https://jsr.io/@zuke/core/1.50.0/src/render.ts' has no exported member 'escapeData'
packages/console declares jsr:@zuke/core@^1.25.0
```

That is the gate doing its job: a wrapper may not use a core API before core ships it. The other 11 checks are green, and the force half is core-internal and unaffected.

To land, once core releases the version carrying the escapes:

1. raise `packages/console/deno.json` to that version, replacing the `^1.25.0` floor;
2. re-run `./zuke coreFloorCheck` to confirm it resolves;
3. merge.

Holding rather than splitting, so the two halves of #546's remaining gap land together.

totollygeek:
Added a third site in this package, found by the adversarial pass after the first two were fixed.

`packages/console/src/renderer.ts` draws its own target header, and under Actions that header is a `::group::` command with the target name interpolated raw — while core's equivalent in `report.ts` has escaped it all along. The name is a target's, which for a fan-out carries an item key from repository or remote data.

It matters more than the position suggests: the header is emitted **before** the body, so a suspend directive there disarms everything the target then prints, including core's own correctly escaped footers and its annotation. In one stream the reviewer showed the header forged and the footer two lines later neutralised.

Test added and confirmed to fail with the escape removed. Gate green at 23/23; still parked on the same core-release constraint as the rest of this PR.

totollygeek:
Finding `19tari8xra1r1` is confirmed and now fixed. It is right, and for a worse reason than it gives.

The escaping was inside `forceTarget`, gated on the ambient environment. But that message is **data**, not output: `runtools.ts` returns it inside the MCP `force_target` tool's JSON payload (`message: result.message`). Encoding it for a terminal the producer does not own meant an MCP client got percent-mangled text whenever the **server** happened to run under Actions — a client on a laptop receiving encoded output because of where the server sat. That is not merely a divergence in the escaping path; it is a correctness bug, and I introduced it.

The environment read was wrong on its own terms too. `forceTarget` already takes an injectable `readEnv` — the MCP tool threads one through — and `detectCiHost` accepts a reader for exactly this reason. Calling it bare ignored the seam the function already offered, which is the dependency-inversion rule in AGENTS.md guideline 13.

The fix moves the escape to the CLI printer, the layer that owns a terminal and already knows it is writing to one. Both sides are pinned:

- the existing integration test asserts `zuke force`'s own output carries no forgeable command;
- a new unit test asserts the returned message is unencoded **even when `readEnv` claims `GITHUB_ACTIONS=true`**, so it fails if the producer starts encoding again.

Gate green at 23/23 on the branch with master merged in.

Issue #559, opened 2026-09-10
chore(ai): raise the core floor to the release carrying job-summary redaction


totollygeek:
Follow-up to #553.

That fix redacts the job summary inside `appendJobSummary`, and `@zuke/ai`'s `writeStepSummary` now delegates to it. The redaction ships in the core release **after** `@zuke/ai`'s declared floor of `^1.51.0`, so a consumer pinned at the floor itself gets the delegation without the masking. No worse than the duplicated writer it replaces, and correct as soon as they move up — but the floor should say what the behaviour actually requires.

It could not be raised in the same change: a floor above the workspace's own core version stops the repository resolving at all.

```
Warning Workspace member '@zuke/core@1.51.0' was not used because it did not match '@zuke/core@^1.52.0'
error: Could not find version of '@zuke/core' that matches specified version constraint '^1.52.0'
```

### To do, once the release carrying the redaction is on JSR

1. Raise `packages/ai/deno.json` to that version.
2. `deno task lock`, and commit the lock in the same change.
3. `./zuke coreFloorCheck` to confirm it resolves.
4. Drop the residual paragraph from `writeStepSummary`'s JSDoc, which exists only to describe this gap.

Worth noting for whoever picks this up: `coreFloorCheck` cannot catch the underlying problem. It type-checks that the export exists at the declared floor, and `appendJobSummary` has existed since core 1.33.0 — what the floor needs to encode here is a *behaviour*, which no type check sees.

PR #560, opened 2026-09-10
fix(core): redact the job summary at the one writer that produces it

### LINKED ISSUE #553: bug: the AI package and zuke.ts write the job summary without the redactor
Follow-up to #546. #547 made the executor apply the run's redactor to the job summary it writes. Four other writers to the same file were not part of that change and still write unredacted.

## The writers

`writeStepSummary` in `packages/ai/src/report.ts` appends straight to `GITHUB_STEP_SUMMARY` with no redactor. Called from `reviewer.ts:569`, `fixer.ts:397` and `:411`, and `agent_fixer.ts:331`.

Separately, `zuke.ts:767` calls the exported `appendJobSummary` directly, also unredacted.

## Why it matters

The guarantee in the secrets guide is that a resolved `secret()` never surfaces in Zuke's own output. The reviewer and fixer markdown is built from a failed command's output and the model's response to it, which is exactly where a secret in an argv or an environment dump would appear — and the job summary is published to everyone who can view the workflow run. The mask directives do not cover it: they mask the runner's log stream, not a file the build writes. This is the same hole #547 closed for the executor's own section, in the other writers.

I have not demonstrated a secret reaching the file through this path; the gap is structural — no redactor is applied — and the fix should land regardless.

## A wrinkle worth reading before fixing

`writeStepSummary` carries this comment:

> Deliberately not delegating to `@zuke/core`'s `appendJobSummary`: this package declares an older core floor than the release that introduced it, and a consumer installing that floor from JSR would get a missing export. Revisit once this package's declared floor has moved past it.

So the duplication is deliberate and floor-driven, not an oversight, and a fix that simply delegates will fail `coreFloorCheck` exactly as #550 currently does. Either raise the floor first, or apply a redactor at the existing writer.

`ambientRedactor()` in `packages/core/src/ambient_redactor.ts` looks like the seam: a remediation or validation has no redactor on its context, and the executor already runs the plan inside `withAmbientRedactor`. That would need the same floor consideration.

## Acceptance criteria

- Every writer to the job-summary file applies the run's redactor, by whatever seam survives the floor constraint.
- A test that a secret-valued string in reviewer or fixer markdown is masked in the summary file, in the shape of the executor's existing one.
- If the duplication of `appendJobSummary` can be removed once the floor moves, say so in the code rather than leaving the comment to be rediscovered.

totollygeek:
Closes #553.

## The gap

The executor's own summary section was redacted; the four other writers were not. The AI reviewer and both fixers append their markdown through the AI package's own copy of the writer, and `zuke.ts` calls the exported one directly.

That markdown is built from a failed command's output and the model's response to it — exactly where a secret in an argv shows up — and the job summary is published to everyone who can view the run. The mask directives do not cover it: they mask the runner's log stream, not a file the build writes.

## Guarding the writer, not the callers

Redacting at each caller is another guard to forget, and a validation, a remediation and a lifecycle hook all have **no redactor on their context** to forget with. So the redaction lives in `appendJobSummary`, the one exported function that writes the file, reading the ambient redactor.

The AI package's duplicate writer is gone with it. Its comment said the duplication existed because the package's core floor predated the export — that was not true, `appendJobSummary` has been exported since core 1.33.0, below the floor even then.

## What the adversarial pass changed

The first commit claimed the executor installs the redactor "around the whole run, so every in-run caller is covered". **It installs it around the plan.** `Build.onStart` and `onFinish` run outside that, and they are the most natural place for a build to write its own summary section. Both published secrets in the clear, and nothing in the change tested either.

Guarding those two call sites would have repeated the mistake the fix exists to avoid, so the redactor is installed where hooks are **dispatched** — covering plugins and any hook added later. The JSDoc now states the boundary honestly, including what is *not* covered: a caller with no run in scope has no redactor to find.

The same pass caught the test probe passing a URL's `pathname` as a config path. That is percent-encoded and puts a slash before a Windows drive letter, so a repository path containing a space — or the Windows leg of the matrix — would not have found the file.

## One residual, stated rather than glossed

The redaction ships in the core release *after* `@zuke/ai`'s floor, so a consumer pinned at the floor itself gets the delegation without the masking. That is no worse than the duplicated writer it replaces, and correct as soon as they move up.

It could not be fixed here: a floor above the workspace's own core version stops the repository resolving at all. Filed as #559, with the reason `coreFloorCheck` cannot catch it — that gate type-checks the export, and what the floor needs to encode is a behaviour.

## Verification

Both halves reproduced before and after. A probe confirmed the ambient redactor reaches a target body, a validation (where the reviewer runs) and a remediation (where both fixers run) — that reachability is what the whole design rests on, so it was measured rather than assumed.

Every new test was mutation-checked: removing the redaction from the writer fails the core and remediation tests, and removing it from the hook dispatch fails the lifecycle test.

Gate green at 23/23: 4474 tests, 98.5% lines, 97.4% branches.


totollygeek:
Finding `3v8p5vxozzsrq` is fixed rather than contested — it is right, and the repo already had the helper for it.

The probe's config path was counted in `..` segments from the test file. That breaks silently if the test moves, and the failure would present as the writer regressing rather than the path being wrong, which is the worse half of the problem.

It now resolves through `repoRoot("deno.json").path`, which walks up to the config file. That is the existing helper this should have used in the first place — guideline 12's "reuse the helper" — and I confirmed it also works when the suite is run from inside `packages/ai` rather than the repository root, which the segment count only handled by coincidence.

For the record on why the path is passed at all: the probe script lives in a temp directory, and this module now delegates to `@zuke/core` instead of carrying its own copy of the writer, so it needs the import map a published consumer would get from the registry. The property under test is unchanged — no permissions, no throw, no write.

Rebased onto master (it had gone `BEHIND` after #557). Gate green at 23/23.

PR #570, opened 2026-09-11
fix(core): escape every line the runner parses, not only the renderer's

### LINKED ISSUE #567: fix: escape every line Zuke prints on an Actions runner, not only the renderer's
#547 and #550 closed the workflow-command injection the assessment named: the
pass/fail/wait/dry-run footers, the summary rows and their notes, the
forced-override line, the echoed command lines, the build-failed culprit, the
suspended-target list, and — in `@zuke/console` — the `::error::`/`::warning::`
body and both `::group::` headers. All of those are escaped at construction, and
`escapeLine` is the right tool for them.

They are all **renderer-composed** lines. Everything else Zuke prints is not
escaped, and the Actions runner parses every line of a step's output regardless
of which function inside Zuke composed it.

## What is still open

About forty `reporter.info` / `reporter.error` sites interpolate text this
process did not author, straight onto that stream:

| module | sites | what they interpolate |
| --- | --- | --- |
| `cancel.ts` | 18 | store-derived run id, actor, status, compensation names, thrown-error messages |
| `executor.ts` | 9 | remote-cache warnings (network-derived), service-stop errors (subprocess stderr), plugin lifecycle messages, parameter-validation messages, state-store open errors, a resumed run id |
| `reap.ts` | 4 | run ids and per-run error messages from the store |
| `resume.ts` | 3 | record-derived text |
| `execute_cancel.ts` | 3 | run id, compensation summary |
| `scheduler.ts` | 2 | the fan-out count line and the lock waiting notice |
| `execute_plan.ts`, `execute_state.ts` | 3 | the `--since` base ref, state-store warnings |

Two of them sit directly beside code that was fixed, and carry the same
untrusted class it was fixed for:

- **The lock waiting notice** — `scheduler.ts:632` passes `reporter.info` as the
  notifier, and the notice describes the current holder: its actor, run id,
  timestamp and run URL, all **written by another run into the shared state
  store**. Thirty lines earlier, `scheduler.ts:455` escapes the forced-override
  actor and reason from that same store, with a comment explaining why. This one
  does not.
- **The fan-out count line** — `scheduler.ts:719` interpolates the target name
  raw. Sub-targets are named after the item key, dispatch to the fan-out path is
  per-target, and `cloneTarget` copies the fan-out spec, so a stage that itself
  declares a fan-out reaches this line with the outer item key in its name.

## Why this needs a sink, not forty edits

Forty call sites cannot be kept escaped by hand, and the decision is already
duplicated four ways and has already drifted:

- the same two-line `safe` lambda appears three times, in `report.ts` and twice
  in `scheduler.ts`;
- about seven more sites inline the same ternary;
- `cli.ts` asks a **different** question — it branches on the detected CI host
  rather than on the resolved style;
- `execute_output.ts` keeps a private helper reading `GITHUB_ACTIONS` for a third
  form of the same predicate.

That is guideline 12's exact shape: a guard whose whole value is being applied
everywhere, with more than one implementation.

## Proposed shape

An escaping reporter decorator beside `redactingReporter`, applying `escapeLine`
to every line. `escapeLine` is a no-op on ordinary text and is idempotent — it
removes the `::` it would encode — so it composes safely with the escaping the
renderer already does.

The exception has to be explicit, because the run's reporter also carries the
workflow commands Zuke deliberately emits, and `::endgroup::` does not survive
being escaped. So the split is:

- **renderer-composed lines** go to an unescaped (still redacted) sink — the
  renderer owns its own escaping and already does it, per-site, with the
  reasoning #547 and #550 wrote. There are exactly seven such sites, five in
  `scheduler.ts` and one in `executor.ts`, plus the wait footer.
- **everything else** goes to the escaping sink, and is covered by construction
  rather than by remembering.

`cancel.ts` and `resume.ts` default their own reporters and emit no workflow
commands at all, so they wrap unconditionally with no exception to carve out.

The invariant to state in the JSDoc and pin with a test: **no line leaves core
unescaped unless the renderer composed it.**

The job summary is out of scope here — it is a file, not the log stream, and
#560 covers its redaction.

## Acceptance criteria

- One implementation of "should this line be escaped for the runner", replacing
  the three copied lambdas, the inline ternaries, `cli.ts`'s host check, and
  `execute_output.ts`'s private environment read. The last two fold onto the
  injectable `ciHost` reader.
- Every non-renderer line Zuke prints during a build, a cancel, a resume and a
  reap is escaped when the run is on Actions.
- The lock waiting notice and the fan-out count line are covered.
- Grouping still works: a target's section still opens and closes, proven by a
  test that would fail if the commands were escaped.
- Unit tests for the decorator: ordinary text unchanged, a leading `::` and a
  mid-line `##[` encoded, idempotent.
- An integration test running a build in Actions mode whose fan-out key and lock
  holder both carry a forged `::stop-commands::` line, asserting that no printed
  line opens a command except the ones the renderer authored — with the negative
  assertion shown able to fail.
- No public API removed or narrowed; `deno doc --lint` clean and the generated
  API artifacts regenerated if the surface changed.

Follows #547, #550. Closes out the residual of the assessment's log-injection
finding.

totollygeek:
Closes #567.

## The gap

The Actions runner parses every line a step writes, whichever function inside Zuke composed it. The escaping so far was applied at construction, per site: it covered the renderer's lines and left about forty reporter calls interpolating text this process did not author — a store-derived actor, a subprocess's stderr, a network warning, a run id. Two sat directly beside code that had been fixed and carried the same untrusted class it was fixed for.

Forty call sites cannot be kept escaped by remembering, and the decision had already drifted four ways.

## The shape

The escaping moves to the sink, with the exception made explicit rather than incidental: lines the renderer composed carry Zuke's own workflow commands, which do not survive being escaped, and the renderer escapes the untrusted values inside them at construction — where it still knows which part of a line is a command and which is data.

With the sink in place the per-site escaping became redundant, and removing it cascaded: two copied lambdas, an inline ternary, then the linter reporting `escapeLine` unused in that module at all, then a parameter that existed only to carry the style into that decision.

## What the adversarial pass changed, which is most of this PR

**A regression I introduced.** Under `--parallel` each target's framework output is buffered so its block flushes atomically, and the scheduler overrode only the escaped sink when handing that buffer down. The renderer's lines went to the live sink instead, so two folds opened before either closed — and the runner does not nest folds, so one target's swallowed the other's while every body line landed outside both. Nothing caught it: the grouping test runs sequentially, the one arrangement where the bug cannot appear. The buffer now records which sink each line came from and replays them in order.

**Twenty-five sites the fix did not reach.** `cancel`, `resume` and `reap` default their own reporters and never touch the run's output composition. A compensation throwing a hostile message put `::stop-commands::` straight on the runner, which disables command processing for the rest of the job — the bulk of what this issue was about, and the first commit missed it despite the issue saying so plainly.

**The CLI was worse in kind.** Its catch arms quote back the argument the invoker passed, which on Actions is often a `workflow_dispatch` input; one of them sits eleven lines below a success path that escapes the same value deliberately. It has a single sink now rather than forty-three console calls to keep straight.

**A trap in my own wrapper.** Written the obvious way, a module-level `const` reads the environment at import time — before a test, or any embedder that sets the variable itself, has set it — and the escaping then silently does not apply. Two probes passed and two still forged commands, which is the only reason it surfaced. The decision is made per line now.

**The contract sentence was wider than the code.** A subprocess's own output is relayed byte for byte to the real file descriptors through no reporter, deliberately — a build may run a tool whose commands are wanted. The guarantee is narrower than "no line leaves core unescaped", and the JSDoc now says the narrow thing: text Zuke itself interpolated into a line Zuke wrote cannot be read as a command.

## Verification

Every guard was removed in turn and each fails a distinct test: the sink, the renderer carve-out, the parallel buffering, and the standalone wrapper. Gate green at 23/23: 4489 tests, 98.5% lines, 97.4% branches.


totollygeek:
Superseded by #569, which landed the same work while this was in review — and took a better approach to the part that overlaps.

#569 keeps per-site escaping for the run's reporter, consolidated behind `escapeLineIf`, and applies the sink wrapper only to reporters that never carry workflow commands. This branch instead split `RunContext` into an escaped and an unescaped sink, which introduced a regression #569 does not have: under `--parallel`, the per-target buffer was overridden for one sink and not the other, so folds opened around the wrong lines and every body line landed outside them. I verified master has no such fault.

What did not overlap is `cli.ts`. I probed current master and two paths still let an argument reach the runner as a command:

```
resume=["::stop-commands::deadbeef","note: ##[error]forged\""]
unknown=["::stop-commands::deadbeef","note: ##[error]forged."]
```

#569 escapes exactly one CLI site (the force message); the other forty-two console calls are raw, including the catch arms that quote a thrown message straight back. That is carried forward in #571, rebuilt on master and using #569's own `escapeLineIf` rather than this branch's helpers.

Issue #575, opened 2026-09-11
bug: the zuke CLI package echoes its arguments unescaped on an Actions runner


totollygeek:
Follow-up to #551. The escaping sweep covered `@zuke/core`'s command surface. `@zuke/cli` is a second one, and it is the binary users actually run.

## The gap

`packages/cli/mod.ts` implements `setup`, `import` and `doc`, and writes through `SetupHost.log`, which is a bare `console.log` (`packages/cli/src/setup.ts:203`). Nothing in that package touches the escaping.

Under `GITHUB_ACTIONS=true`, an unknown command echoes its argument straight back:

```text
Unknown command: ::stop-commands::pwn
```

The first non-blank characters are `::`, so the runner executes it and stops processing commands for the rest of the job.

The raw-argv sites in that file are the unknown-command message (`mod.ts:435`), the scaffold destination from `--dir` (`:244`), the `--from` source (`:321`), and the catch-all `host.log(error.message)` (`:432`) — the same "catch arms are the sharp end" shape that #571 fixed one package over.

## Why it was not fixed with the rest

The sink lives in `packages/core/src/reporter.ts` as `cliReporter`, and it is **internal** — not exported from `mod.ts`. Using it from `@zuke/cli` needs two things that cannot land in the same change:

- a new public export from `@zuke/core`, and
- `@zuke/cli`'s core floor raised from `^1.40.0` to the release carrying it.

A floor above the workspace's own core version stops the repository resolving at all, which is the same constraint #559 is parked on:

```text
error: Could not find version of '@zuke/core' that matches specified version constraint
```

## Options, in the order I would consider them

1. **Wait for the release, then export and raise the floor.** Cleanest, and it keeps one implementation of the escaping. Blocked until core ships.
2. **A local copy in `@zuke/cli`.** Unblocked now, but it is a second implementation of a guard whose whole value is being applied everywhere — exactly what AGENTS.md guideline 12 names, and this escaping has already drifted twice while it had more than one home.
3. **Pass a reporter in.** `SetupHost` is already an injectable seam; the caller could supply an escaped sink rather than the package deciding. Worth weighing against 1, since it removes the floor question entirely.

## Acceptance criteria

- No write in `@zuke/cli` reaches the runner unescaped, by whichever route above.
- The existing source-level assertion in `packages/core/tests/escaping_sink_test.ts` extended to cover `packages/cli`, so a new direct console write in that package fails the suite rather than being noticed later.

PR #576, opened 2026-09-11
fix(core): route every command-surface write through the escaping sink

### LINKED ISSUE #551: bug: zuke runs show prints an untrusted actor raw, forging workflow commands under Actions
Follow-up to #546. #547 escaped the build's own reporting path; the run-inspection commands were not part of that change and still print untrusted values raw.

## Reproduction

On master (a46fd72), with a run whose actor carries a newline and a workflow command, then `zuke runs show <id>` under `GITHUB_ACTIONS=true`:

```
forged=["::stop-commands::TOKEN","::stop-commands::TOKEN (human)"]
```

Two lines the runner would parse as commands, from one hostile actor. `::stop-commands::` suspends command processing for the rest of the step, so anything Zuke emits afterwards — including its own annotations and `::add-mask::` directives — is inert.

The sites are in `packages/core/src/runs.ts`: the `actor` line, the `started` line for the initiator, the per-target forced-override lines, and the event rows, all plain interpolation. The forced-override `reason` reaches the same output through `why`.

## Why the actor is not merely operator-typed

`resolveActor` (`packages/core/src/state/record.ts:52`) takes an `extra` list of candidates, and `registry_server.ts:881` passes `this.#clientLabel`. That label is set at `registry_server.ts:553-554` from `negotiateInitialize(params, …)` — the MCP client's **self-reported name in its `initialize` request**.

So a client connecting to a registry MCP server chooses the string that is later stamped on the run and printed by `runs show`. This is the same untrusted-input class the original issue was about, not an operator typing into their own terminal.

## Same shape, not reproduced

`packages/core/src/lock.ts:44` formats a lock holder as `${holder.actor} (run …) since …`, which reaches the "waiting for lock" message. Holder actors come from the shared state store, so it looks like the same exposure; I did not reproduce it and it should be confirmed rather than assumed.

## Acceptance criteria

- Every value the run-inspection commands print that the build does not control — actor, initiator actor, forced-override actor and reason, event actor — routed through the existing `escapeLine`, which #547 published on `@zuke/core/render` for exactly this.
- The lock-holder line checked and escaped if it is affected.
- An integration test asserting over the whole stream of `runs show` for a run with a hostile actor, in the shape of the existing GitHub-output tests, so a future printer is caught rather than trusted.

totollygeek:
Closes #551.

## The gap

`zuke runs show` printed an actor read back from the shared store straight to the log, forging workflow commands on a runner. `zuke runs list` did too, which the issue did not mention.

That actor is not always operator-typed: the registry MCP server feeds `resolveActor` the client's own self-reported name from its `initialize` request, so it can be chosen by whoever connects.

The escaping sinks added in #569 and #571 did not reach it — `runs.ts` is a different module with its own console writes. Sweeping for the shape found four such modules, so the sink moved into the reporter module as one implementation and all of them use it. The tell that it now sits in the right place: `cli.ts` imports neither the escape helper nor the host detector any more.

## What the adversarial pass changed

**I introduced a regression, and the commit message denied it in so many words.** It claimed there was "nothing to corrupt". But the line escape encodes the legacy bracket marker *wherever* it appears, and the JSON commands went through the same sink — so a consumer piping `runs list --json` into a parser read a different actor on Actions than it read locally, silently, because the payload still parses.

Leaving JSON unescaped is not the answer either: the runner matches that marker mid-line, so raw JSON genuinely can forge a command. There was a third option I had not looked for. Escaping it *as JSON* costs nothing, because the escape for `#` is `#` to every parser — the bytes on the wire cannot carry the marker, and the value round-trips unchanged. The test asserts all three properties: safe on the wire, identical after parsing, identical to the non-Actions run.

**The graph command has its own host**, with its own console write, reached from the very function this change rewrote. It reports a path that on a runner can come from a matrix entry or a dispatch input.

**Two of the four modules had no test at all.** Reverting either the registry command or the MCP diagnostics to a bare console call left the whole suite green. That is answered at the source rather than behaviourally: what is promised is not that one message is escaped but that no module on this surface writes past the sink, and a behavioural test can only cover the lines it happens to trigger.

## Left out deliberately

`@zuke/cli` — the binary users actually run — echoes its arguments unescaped through its own host. Fixing it needs a new public export from core and a floor above an unreleased version, which cannot land in the same change. Filed as #575 with three options and the constraint spelled out.

## Verification

Every guard removed in turn fails a distinct test: the sink, the JSON encoding, the graph host, and the source-level assertion. Gate green at 23/23.

### DIFF
--- a/packages/core/src/cli.ts
+++ b/packages/core/src/cli.ts
@@ -7,8 +7,7 @@
  */
 
 import { type Build, discoverGroups, discoverTargets } from "./build.ts";
-import { detectCiHost } from "./host.ts";
-import { escapeLineIf } from "./render.ts";
+import { cliReporter } from "./reporter.ts";
 import { discoverCiFiles, syncCiFiles } from "./ci.ts";
 import { isEntryModule } from "./entry.ts";
 import { messageOf } from "./internal.ts";
@@ -75,32 +74,6 @@ import {
   type InstallOptions,
 } from "./completions_install.ts";
 import { describeBuildSurface } from "./describe.ts";
-
-/**
- * The CLI's own output, neutralised for a GitHub Actions runner.
- *
- * Every command here echoes arguments the invoker chose — a run id, a target
- * name, a force reason — and on Actions those commonly come from a
- * `workflow_dispatch` input or an issue body rather than from a person at a
- * terminal. The catch arms are the sharp end: they quote a thrown message
- * straight back, so they leak exactly what the success paths are careful about.
- *
- * A sink rather than a call-site guard because there are forty-odd writes here
- * and no reason for any of them to be exempt: unlike the run's reporter, this
- * module emits no workflow commands of its own, so there is no half to carve
- * out. The decision is made per line, not once at module load — a top-level
- * `const` would read the environment before an embedder or a test has set it,
- * and the escaping would then silently not apply.
- */
-const cli: Reporter = {
-  info: (line) => console.log(escapeLineIf(inActions(), line)),
-  error: (line) => console.error(escapeLineIf(inActions(), line)),
-};
-
-/** Whether this process is running on a GitHub Actions runner. */
-function inActions(): boolean {
-  return detectCiHost() === "github";
-}
 
 /** `completions` sub-action: print the script to stdout. */
 const PRINT_SUBCOMMAND = "print";
@@ -862,18 +835,18 @@ export async function syncCiConfig(
   const files = discoverCiFiles(build);
   if (files.length === 0) {
     if (!options.quietWhenEmpty) {
-      cli.info("No CI configuration is declared on this build.");
+      cliReporter.info("No CI configuration is declared on this build.");
     }
     return 0;
   }
   const results = await syncCiFiles(files, { check: options.check });
   const stale: string[] = [];
   for (const { path, status } of results) {
-    if (status === "written") cli.info(`Generated ${path}`);
+    if (status === "written") cliReporter.info(`Generated ${path}`);
     else if (status === "stale") stale.push(path);
   }
   if (stale.length > 0) {
-    cli.error(
+    cliReporter.error(
       `CI configuration is out of date: ${stale.join(", ")}.\n` +
         `Run \`zuke generate-ci\` and commit the result.`,
     );
@@ -941,19 +914,23 @@ async function installCompletionScript(
       params,
       options.installOptions,
     );
-    cli.info(`Installed ${result.shell} completion to ${result.scriptPath}`);
+    cliReporter.info(
+      `Installed ${result.shell} completion to ${result.scriptPath}`,
+    );
     if (result.rcPath === undefined) {
-      cli.info("Open a new shell (or restart fish) to load it.");
+      cliReporter.info("Open a new shell (or restart fish) to load it.");
     } else if (result.alreadySourced) {
-      cli.info(`${result.rcPath} already sources it — nothing to change.`);
+      cliReporter.info(
+        `${result.rcPath} already sources it — nothing to change.`,
+      );
     } else {
-      cli.info(
+      cliReporter.info(
         `Added a source line to ${result.rcPath}; open a new shell to use it.`,
       );
     }
     return 0;
   } catch (error) {
-    cli.error(messageOf(error));
+    cliReporter.error(messageOf(error));
     return 1;
   }
 }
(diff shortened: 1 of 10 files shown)


totollygeek:
Finding `1wnn9jxk54p7j` is right and is fixed. The assertion listed five modules by name, so it could only catch a write in a module someone had remembered to add — while the regression it exists for is a **new** write appearing somewhere unexpected.

It scans the tree now, and found two more the moment it did: the conformance helper defaulted its injectable log to the console (routed through the sink), and the graph HTML generator has a `console.warn` inside a template literal — browser JavaScript that runs in the viewer's page, not in this process. That one is named in the allowlist with the reason, rather than stripping template literals automatically and risking a real write being hidden.

---

Finding `ro9ftef3jvhs` I'd contest. It reports that JSON output "mutates payload bytes by replacing every `#` with \`\\u0023\`". The byte change is the fix, not a defect, and the payload is unchanged where it counts.

`\\u0023` **is** `#` to every JSON parser, so a consumer reads the identical string. The test asserts all three properties, and each fails if the encoding is removed: the marker is absent from the wire, the parsed value equals the original, and it equals what the same command prints when not on a runner.

The alternative is worse in both directions. Before this, `runs list --json` went through the line escape, which rewrites that marker **in place** — so a script piping into a parser genuinely did read a different actor on Actions than it read locally, silently, because the payload still parsed. Leaving JSON unescaped is not available either: the runner matches the marker mid-line, so raw JSON really can forge a command. Encoding it as JSON is the only option that is both safe on the wire and faithful to the value.

The one real cost, stated plainly: a consumer comparing **raw bytes** — a checksum or a golden-file diff of the output — sees a difference on Actions. That seemed clearly preferable to a silently different parsed value, which is what the previous behaviour gave.

totollygeek:
Findings `ro9ftef3jvhs` and `3509tdwe0dakc` are partly right, and are now addressed rather than argued.

The encoding was applied unconditionally, so the output changed shape everywhere — while the threat exists only where something is scanning it. That was over-broad, and "a consumer comparing exact bytes sees a difference" is a fair objection to it.

It is confined to a runner now:

- **off Actions** the JSON goes out byte for byte as it always has, so a consumer diffing it against a golden file or checksumming it sees no change at all;
- **on Actions** the bytes differ and the parsed value does not — `\\u0023` is `#` to every parser, so a script piping into `jq` reads the identical string.

Both properties are asserted and each fails without its half.

Where I'd still push back is on the framing of the high-severity version: the alternative was not "leave the payload alone". Before this change `runs list --json` went through the line escape, which rewrites the marker **in place**, so a script genuinely did read a different actor on Actions than it read locally, silently, because the payload still parsed. And leaving JSON unescaped is not available: the runner matches that marker mid-line, so raw JSON can forge a command. Encoding it as JSON was the only option that is both safe on the wire and faithful to the value — the correct criticism was of its scope, which is what is fixed here.

totollygeek:
Finding `l1mrqxi75civ` is fair and is fixed. Escaping every `#` was broader than the threat: the runner matches a specific three-character marker, so escaping the first character of each occurrence is enough. Every other hash in the payload is now written as itself, and the raw bytes differ only where they must.

Checked against the overlapping and adjacent forms — three and four hashes before the bracket, two markers in one value, a bracket preceded by a single hash — none leaves the marker on the wire, and each still parses back to exactly what it was.

Also worth recording against `ro9ftef3jvhs`, which is now marked fixed: the scope criticism was the right one, and narrowing it to a runner was the fix. The byte change itself was not avoidable — the previous behaviour rewrote that marker **in place** in the payload, so a script piping `--json` into a parser silently read a different value on Actions.

totollygeek:
Both findings addressed.

`35cp2oouh8vtc` is right and was a real hole. The scan's allowlist matched **file names**, so a nested module sharing a basename with an allowed one was exempt by accident. It is keyed by path from the source root now, and I checked it by planting `registry/reporter.ts` with a bare console write — the test catches it, where before it would have been silently allowed.

The template-literal half of that finding is why `graph_html.ts` is named in the allowlist with its reason rather than inferred: its `console.warn` is browser JavaScript that runs in the viewer's page, and stripping template literals automatically would risk hiding a real write.

`28vy4mfaomfrw` I'd contest on the facts, but it asks for something worth having. The replacement runs over the **whole serialised document**, so field position is not a factor — there is no per-field handling that could miss a nested value. What the finding is right about is that a reader cannot tell that from the call site.

So there is now a test over a payload with the marker in a key, in a deeply nested value, twice inside one string, in array elements, and in adjacent runs of three and four hashes. It asserts the marker is absent from the wire, that the document parses back **equal** to what went in, and that a hash which is not part of the marker is written as itself.

On the suggestion of "a transport that preserves the data model exactly on all runners": that is what this is. The data model is preserved exactly — `\\u0023` is `#` to every parser — and the alternative is not available, because the runner matches the marker mid-line, so raw JSON on a runner really can forge a command.

PR #592, opened 2026-09-13
feat(core): let a tool wrapper write to its tool's stdin

### LINKED ISSUE #591: feat(core): let a tool wrapper write to its tool's stdin
#566 gave `Command` a `stdin` so a credential would have somewhere to go other
than argv. A wrapper still cannot reach it.

`ToolSettings.run` calls a private `#execute`, which calls a private `#command`
to build the `Command`. A subclass sees none of them. Its extension points are
`defaultTool`, `buildArgs`, `onOutput` and `defaultResolution` — argv and
output, nothing about input.

So the API currently promises something it cannot deliver. `markSecret`'s own
JSDoc says:

> Prefer `Command.stdin` wherever the tool offers a stdin form: this masks the
> *rendering*, not the argv, so the value is still visible in the process table
> to anyone on the same host.

There is no wherever. Every wrapper that wants the safe form has to use the
unsafe one and mask it.

## What it blocks

`docker login --password-stdin`, which is the canonical case and the most
severe of the eight argv-credential findings — a registry password readable by
any local user through `/proc` or `ps`.

`@zuke/docker` already has the shape of the fix and it does not work:
`DockerLoginSettings.passwordStdin()` takes **no argument**. It appends the flag
and nothing pipes anything, so the child inherits Zuke's stdin — EOF on CI, a
hang on a terminal. The module's own example calls `.passwordStdin(token)` with
an argument the method does not accept, so the documented safe path has never
run.

## Proposed shape

A protected hook alongside the others, consulted by `#command`:

- returns the text to write, or `undefined` to leave stdin inherited;
- default returns `undefined`, so every existing wrapper is unchanged;
- the value never reaches `commandLine`, which `Command.stdin` already
  guarantees.

Named for what it is rather than for the flag, since it is not docker-specific:
`npm` and others take secrets on stdin too.

## Why a separate change

The wrapper packages resolve `@zuke/core` at the version published on JSR —
`coreFloorCheck` type-checks them against it. So the seam has to land and
release before `@zuke/docker` can use it. Splitting it out lets the other seven
argv-credential findings, which need only the already-published `markSecret` or
an environment route, land in parallel rather than queue behind a release.

## Acceptance criteria

- A `ToolSettings` subclass can supply stdin text without reaching into core
  internals.
- A wrapper that supplies none behaves exactly as before — stdin stays
  inherited, and the spawn is unchanged.
- The supplied text does not appear in `commandLine`, in a dry-run echo, or in
  a `CommandError` message.
- Covered by a unit test that exercises both arms through a real subprocess,
  since the property is about a real pipe rather than about argv.
- Documented, `deno doc --lint` clean, and the generated API artifacts
  regenerated.

totollygeek:
## What & why

#566 gave `Command` a `stdin` so a credential would have somewhere to go other
than argv. No wrapper could reach it.

`ToolSettings.run` calls a private `#execute`, which calls a private `#command`
to build the `Command`. A subclass sees `defaultTool`, `buildArgs`, `onOutput`
and `defaultResolution` — argv and output, nothing about input. So the API
documented a route nobody could take: `markSecret`'s own JSDoc says to prefer
`Command.stdin` wherever the tool offers a stdin form, and there was no
wherever.

This adds the missing extension point. It returns the text to write, or nothing
to leave stdin inherited; the base returns nothing, so every wrapper in the
workspace behaves exactly as it did. It is named for what the tool reads rather
than after the flag that motivated it, since docker is not the only tool that
takes a secret this way.

## What it unblocks

`docker login --password-stdin` — the most severe of the eight
argv-credential findings, a registry password readable by any local user
through `/proc` or `ps`.

Worth knowing what is there now: the `passwordStdin` chainer on
`DockerLoginSettings` takes **no argument**. It appends the flag and nothing
pipes anything, so the child inherits Zuke's stdin — EOF on CI, a hang on a
terminal. The module's own example calls it *with* a token, which the method
cannot accept. The documented safe path has never run. That fix needs this seam
released first, because `coreFloorCheck` type-checks each wrapper against the
core version published on JSR.

Splitting it out is deliberate: the other seven argv-credential findings need
only the already-published `markSecret` or an environment route, so they land in
parallel rather than queue behind a release.

## Related issues

Closes #591

## Checklist

- [x] The PR title is a
      (`type(scope): summary`).
      changed.
      type guards instead).
      one implementation.
- [x] A new package was wired into all six places (see
      if applicable.

## Notes for the reviewer

**Open/closed rather than a flag.** It is a hook beside the existing ones, not a
parameter that switches the run path between two behaviours — a wrapper states
what its tool reads and the base decides what to do about it.

**Two of the three tests drive a real subprocess**, because what is being proved
is that text arrives on a real pipe: the child reads it back, and it appears in
no rendering of the command — not the argv, not a failure message.

**The third test is deliberately not a subprocess test.** The default arm is
"stdin stays inherited", and inherited input under the test runner is whatever
the runner happened to have, so a spawn-based assertion would hold whether or
not the default were correct. It pins the contract instead — that the base
returns nothing — which is the thing a wrong default would break. A base
returning an empty string would compile, would read as harmless, and would
switch every wrapper in the workspace from inherited stdin to a pipe that closes
immediately.

**Each was mutation-tested**: ignoring the hook in the command builder,
returning an empty string from the base, and letting the text reach argv as
well. All three turn a test red. The last one matters most — a secret leaking
into the command line is the exact failure this exists to prevent, so it is
asserted rather than assumed.

https://claude.ai/code/session_01HNN7wc1BczPDB7zEaQ9noH

---
_Generated by [Claude Code](https://claude.ai/code/session_01HNN7wc1BczPDB7zEaQ9noH)_


totollygeek:
Contesting `y5t5emfkhxhq` — the stdin-hook documentation is not duplicated.

Counted on this head:

| symbol | `llms-full.txt` | `packages/core/README.md` |
| --- | --- | --- |
| `stdinInput` | 1 | 1 |
| `markSecret` | 2 | 2 |
| `onOutput` | 29 | 2 |

`stdinInput` appears exactly **once** per generated artifact, which is the
minimum possible for a documented symbol. `markSecret` appears twice in each
because the new hook's JSDoc cross-references it, and `onOutput` appears 29
times in the index because every wrapper package documents its own override.

The finding appears to read the same JSDoc text landing in both files as
duplication. Those are two different artifacts by design: `llms-full.txt` is the
repo-wide index of every package's surface, and `packages/core/README.md` is
that one package's `## API` block. Both are generated from `deno doc` by
`./zuke apiDocs`, and `apiDocsCheck` fails the gate if either drifts from the
source — so the text appearing in both is the generator working, not redundancy
introduced here. The same is true of every exported symbol in the workspace;
nothing about this hook is special.

No change made. Happy to be shown otherwise if the finding meant something
narrower than the two-file overlap.

---

PR #600, opened 2026-09-14
fix(docker): make the documented safe login path actually pipe the password

### LINKED ISSUE #599: bug: @zuke/docker's documented safe login path has never worked, and pipes nothing
`DockerLoginSettings` offers two ways to pass a registry password. One puts it in the process argv. The other is documented as the safe alternative and **does not work at all**.

## 1. `passwordStdin()` takes no password and pipes nothing

```ts
/** Read the password from STDIN (`--password-stdin`). */
passwordStdin(): this {
  this.#passwordStdin = true;
  return this;
}
```

All it does is append the flag:

```ts
if (this.#passwordStdin) argv.push("--password-stdin");
```

Nothing is ever written to the child's stdin. `docker login --password-stdin` then blocks reading a stream the wrapper never writes to, or fails outright. There is no way through this API to supply the password it promises to read.

## 2. The module's own example calls it with an argument it cannot accept

`registry.ts`'s `@module` block:

```ts
await DockerTasks.login((s) => s.registry("ghcr.io").username(user).passwordStdin(token));
```

`passwordStdin` is `(): this`. The documented call passes `token` to a zero-argument method — so the example never type-checked against the code it documents, and the one path the docs steer people to has never run.

## 3. The unsafe path is the only working one, and it is not even masked

```ts
if (this.#password !== undefined) argv.push("-p", this.#password);
```

`password()`'s own JSDoc says this "lands directly in the process argv, where it can leak through `ps`/process listings" and points at `passwordStdin` as "the safe choice in CI (and generally)". So the wrapper documents the hazard accurately, offers a remedy that cannot work, and leaves the hazardous route as the only one that does.

It also never registers the value with the run's redactor, so unlike the wrappers fixed in #594, even the *rendering* of the command is unmasked.

A child's argv is world-readable on a default Linux host — `ps`, `/proc/<pid>/cmdline` — to every other user and every process the build starts. This was rated the most severe instance of that class in the original assessment, and it is the last one outstanding: #594 fixed the other seven and deliberately left this one because it needed a hook that did not exist yet.

## Now unblocked

`@zuke/core@1.55.0` is published and carries the `stdinInput()` hook from #592 (`tooling.ts:295`, consumed in `#command` at `:457`), so a settings class can finally supply stdin. Confirmed against the published source, not assumed.

`DockerTasks.login` reaches it: `runSettings` → `run()` → `#execute` → `#command`, which calls `this.stdinInput()` and wires `command.stdin(input)`.

The package's floor is `^1.25.0` and has to rise to the release carrying that hook. That also brings `markSecret`, which arrived in 1.53.0.

## Acceptance criteria

- `passwordStdin(token)` takes the password, appends `--password-stdin`, and actually pipes the token to the child's stdin via the `stdinInput()` hook.
- The token never appears in argv, asserted on the whole argv rather than just the flag — a regression could pass it positionally or glued to a flag.
- The token is registered with the run's redactor, so renderings mask it.
- `password()` keeps existing (the wrapper mirrors the real CLI, which has `-p`), but registers its value with the redactor too, and its JSDoc stops pointing at a remedy that does not work.
- The `@module` example matches the real signature — it is the thing that has been wrong longest.
- `@zuke/docker`'s core floor is raised to the release that carries `stdinInput`, and `coreFloorCheck` passes.
- Unit tests for the argv, plus an integration test proving the password reaches the child's stdin rather than its command line.

totollygeek:
`DockerLoginSettings` offered two ways to pass a registry password. One put it in the process argv. The other was documented as the safe alternative and **did nothing**.

`passwordStdin` took no argument. It set a private boolean, and the only effect of that boolean was to append `--password-stdin` to the argv. Nothing was ever written to the child's standard input, so `docker login` was left reading a stream the wrapper never opened.

Worse, the module's own usage example called it **with a token it could not accept** — passing an argument to a zero-argument method. So the documented safe path had never type-checked against the code it documented, and had never run. Meanwhile `-p`, the route the very same JSDoc warns about as leaking through process listings, was the only one that worked. That JSDoc pointed at the broken method as "the safe choice in CI (and generally)".

A child's argv is world-readable on a default Linux host, through `ps` or `/proc/<pid>/cmdline`, to every other user on the machine and every process the build starts. This was the most severe instance of that class in the assessment and the last one outstanding: #594 fixed the other seven and deliberately left this one, because it needed the stdin hook from #592, which only reached JSR in `@zuke/core@1.55.0`.

## The fix

The setter now takes the password, registers it with the run's redactor, and returns it from the `stdinInput` override that core reads when it builds the command. The flag goes to argv; the token does not.

The `-p` flag **stays**. docker accepts it and the wrapper mirrors the real CLI, per guideline 7, so silently dropping a supported flag would be the wrong correction. It now registers its value too, so Zuke's renderings mask it, and its JSDoc says plainly that masking does not cover the process table — instead of pointing at a remedy that was broken. Overstating what masking buys would repeat the original mistake in a quieter form.

The module example is corrected to the signature it always described. It is the part that had been wrong longest, and the reason the gap was not obvious to anyone reading the docs.

### On the signature change

Requiring the token is a breaking signature change, and worth being explicit about: there is no working caller to break. The zero-argument form could not supply a password through this API at all, so every existing call either hung or failed. Requiring the token is what turns the flag into something that means anything.

## Tests

Unit coverage in the docker package, and an integration test driving a real build through the CLI, per AGENTS.md.

The integration test asserts that a **real child process actually read the token**, rather than asserting the flag is present — and that distinction is the whole point. Four mutations are checked, including one that restores the exact original defect, with the flag appended and nothing piped. The argv assertions all pass against that broken version; only the test that checks something genuinely arrived on standard input catches it. The other three cover the token leaking into argv, and the redactor registration being dropped from either route.

It stays hermetic per guideline 5: the child process is the running `deno` binary, so no docker daemon and no network are involved. Only the binary and the subcommand are swapped — the password setter and the stdin override under test are the real ones, inherited.

## Floor

The core floor moves from `^1.25.0` to `^1.55.0`, the release carrying the stdin hook, verified in the published tooling source rather than assumed. That release also brings the redactor registration. `coreFloorCheck` re-checked the docker package against the new floor and passes across all 57 packages. The `deno.lock` diff is the one line.

Gate is green — 4622 tests, 98.3% lines / 97.4% branches.

Closes #599

https://claude.ai/code/session_01HNN7wc1BczPDB7zEaQ9noH


totollygeek:
Two red runs on `f5b2abe` — both were this PR's **description**, not its code.

`prBodyLint` rejected the original body: two fenced code blocks and one sentence that reads as a call statement. That gate exists because release-please's parser chokes on parentheses and silently drops the commit from the release, and RELEASING.md says to keep snippets in the discussion rather than the body. My mistake, and squarely the thing the rule warns about.

The body is rewritten in prose and verified locally against the gate. The second red run was a GitHub re-run, which replays the original event payload — so it re-read the old body and produced a byte-identical error, still quoting a sentence that no longer exists. `ci.yml` triggers on `pull_request` with the default types, which exclude `edited`, so a body edit alone cannot clear it; only a fresh `synchronize` event can.

`a24f743` carries that event, and it is a real change rather than a nudge to CI. Re-reading the diff turned up a defect in my own JSDoc: the stdin setter documented the bug it had just fixed — *"until this release it took nothing and piped nothing"*. That is changelog text in reference documentation. It reaches the JSR page and the package README, where "this release" names nothing a reader can resolve, and it ages badly. It now states the contract — the password is required because the flag only tells docker to read standard input, and the wrapper is what must put something there — which stays true whatever the version. The history is in this PR and the changelog already.

The code under review is otherwise unchanged. Local gate on `a24f743` is green: 4622 tests, 98.3% lines / 97.4% branches, `prBodyLint` passing against the corrected body.

Worth recording for anyone reading back: `security` was **never** failing here. It succeeded on CI in 3.5s, no findings, no leaks. I reproduced a zizmor 403 against the advisories API locally and it also reproduced on the base commit, which made a convincing case for a base-branch break — it was wrong. That 403 is an artifact of my unauthenticated sandbox, and the real cause was only visible in the job log.

---

Issue #623, opened 2026-09-20
chore(docs): sweep the remaining `from "jsr:@zuke/…"` snippets to bare specifiers


totollygeek:
## Context

#622 moved the scaffold to bare specifiers: `zuke setup` now writes
`import { Build, run, target } from "@zuke/core";` and declares
`"@zuke/core": "jsr:@zuke/core@^1"` in the scaffolded `deno.json`, because
Deno's default lint set rejects an inline `jsr:` specifier under
`no-import-prefix` and the scaffold configures no `lint.rules`.

That PR updated the two agent skills, since they are what an agent reads
immediately after running setup: `skills/zuke-setup/SKILL.md`,
`skills/zuke-write-build/SKILL.md` and its `references/cheatsheet.md`. It
deliberately stopped there to stay reviewable.

## The remaining corpus

`grep -rE 'from "jsr:@zuke/'` still matches roughly 230 import statements
across ~90 files:

- `docs/` — ~30 files, `docs/installing-tools.md` (13), `docs/run-context.md`
  and `docs/authoring.md` (9 each) the largest;
- `packages/*/README.md` — ~55 files, mostly the two-line usage intro, plus
  `packages/core/README.md` (9);
- `examples/*/zuke.ts` — 5 runnable example builds (these also need an
  `imports` block in each example's config, so they are the only part of the
  sweep that is more than a text substitution);
- `README.md`.

Anyone pasting one of those into a scaffolded project reintroduces the lint
error the scaffold change removed, so they should agree.

## Shape of the work

- The substitution is mechanical and safe when anchored on `from "jsr:@zuke/`
  — it must NOT touch command lines, which stay as they are:
  `deno doc jsr:@zuke/deno`, `deno run -A jsr:@zuke/cli`,
  `deno install -A -g -n zuke jsr:@zuke/cli`.
- `build/snippets.ts` already rewrites `jsr:@zuke/` to the workspace name
  before type-checking, so `snippetsCheck` passes either way — the sweep does
  not change what the gate proves.
- `examples/` need a real `imports` block per example and must stay green under
  `examplesCheck`.
- Some package README content is generated: check whether the `## API` blocks
  `./zuke apiDocs` writes carry any of these specifiers, and fix the generator
  rather than the output if so.
- Worth deciding at the same time: whether `packages/*/README.md` should show
  the `deno.json` `imports` entry alongside the import line, since a bare
  specifier is not self-contained the way the inline one was.

## Acceptance criteria

[...]

PR #624, opened 2026-09-20
fix(cli): scaffold bare specifiers in deno.json import map

### LINKED ISSUE #622: bug: `zuke setup` scaffolds a project that fails `deno lint` (no-import-prefix)
## What happened

`zuke setup` scaffolds a `zuke.ts` whose first line is an inline JSR specifier:

    import { Build, run, target } from "jsr:@zuke/core@^1";

and a `deno.json` that declares a `lint` task but no `imports` map. The very
first thing the scaffold invites you to run — `deno task lint` — fails on the
file setup just wrote:

```
error[no-import-prefix]: Inline 'npm:', 'jsr:' or 'https:' dependency not allowed
 --> /tmp/demo/zuke.ts:1:36
  |
1 | import { Build, run, target } from "jsr:@zuke/core@^1";
  |                                    ^^^^^^^^^^^^^^^^^^^
  = hint: Add it as a dependency in a deno.json or package.json instead and reference it here via its bare specifier

Found 1 problem
```

Editors show it as a squiggle on line 1 of a brand-new project, which is the
first impression Zuke makes.

`no-import-prefix` is in Deno 2.8's *default* rule set — the set that applies
when a `deno.json` does not configure `lint.rules` explicitly, which is exactly
what setup writes. (This repo's own `examples/` escape it only because the root
`deno.json` pins `"tags": ["recommended"]`.)

## Reproduction

```sh
mkdir demo && cd demo
deno run -A jsr:@zuke/cli setup --yes --name Demo
deno lint      # or: deno task lint
```

## Expected

A freshly scaffolded project passes `deno fmt --check`, `deno lint` and
`deno check` with no edits.

## Scope

Same defect in `zuke import`, which generates its build through the same
scaffolder and emits `jsr:@zuke/core@^1` plus `jsr:@zuke/cmd@^1`.

## Proposed fix

Take the hint the linter gives, which is also the form
[`docs/versioning.md`](docs/versioning.md) and
[`docs/node-projects.md`](docs/node-projects.md) already document — declare the
dependency in `deno.json` and import it by bare specifier:

- scaffold `"imports": { "@zuke/core": "jsr:@zuke/core@^1" }` into `deno.json`
  (merged like the tasks are: only missing entries added, existing ones kept);
- have the starter build and the imported build use `@zuke/core` /
  `@zuke/cmd`;
- generalise the "already set up" check on `deno.json` so a file that has the
  `zuke` task but not the imports is still completed rather than skipped —
  otherwise re-running setup over an existing project leaves an unresolvable
  `zuke.ts`.

Not an option: scaffolding `lint.rules.tags: ["recommended"]` to silence the
rule. That hides the diagnostic instead of fixing the import, and it would
quietly narrow the lint rules of every project Zuke is set up in.

## Acceptance criteria

- [ ] `zuke setup` + `deno lint` is clean in a fresh directory.
- [ ] `zuke import` output is clean too, including the `@zuke/cmd` import.
- [ ] Re-running setup on a project that already has the `zuke` task adds the
      missing `imports` rather than skipping `deno.json`.
- [ ] An existing `@zuke/core` import entry (a pinned version) is never
      rewritten.
- [ ] Unit + integration coverage, plus an e2e test that runs a real
      `deno lint` over the scaffold so the regression cannot come back.
- [ ] `skills/zuke-setup/SKILL.md` documents the `imports` merge (with the
      plugin version bumped).

totollygeek:
## What & why

`zuke setup` scaffolded a project that failed its own linter on the first run. The starter `zuke.ts` opened with an inline JSR specifier, and the `deno.json` beside it declared a `lint` task but no import map — so `deno lint` rejected line 1, column 36 of the scaffold's only source file under `no-import-prefix`, "Inline npm:, jsr: or https: dependency not allowed". The full output is in #622.

`no-import-prefix` is in Deno's **default** rule set — the one that applies when `deno.json` configures no `lint.rules`, which is exactly what setup writes. So `deno task lint` failed on a brand-new project, and editors showed a squiggle there from the start. This repo's own `examples/` escape it only because the root `deno.json` pins the `recommended` tag explicitly.

The fix takes the hint the linter gives, which is also the form `docs/versioning.md` and `docs/node-projects.md` already document: declare the dependency in `deno.json`, import it by bare specifier. Scaffolding a `recommended`-only lint rule set to silence the rule was rejected — it hides the diagnostic instead of fixing the import, and would quietly narrow the lint rules of every project Zuke is set up in.

**Key changes**

- The starter build imports `@zuke/core` bare; `deno.json` gains an `imports` entry mapping it to `jsr:@zuke/core@^1`. The caret-major pin moves into the map rather than being lost, and `zukeDependency` is its single definition, shared with `zuke import`.
- `mergeDenoJson` merges `imports` the way it already merged tasks: additive per key, so a project that pinned `@zuke/core` to an exact version keeps that pin.
- `denoJsonState`, renamed from `zukeTaskState`, checks the `zuke` task *and* the imports, so re-running setup over a project that has the task completes the file instead of skipping it — skipping left the `zuke.ts` written in the same run with nothing to resolve.
- `generateBuild` returns the source plus the imports it needs, so `zuke import` declares `@zuke/cmd` only when it actually generates a command.

**Two regressions this introduced, found by adversarial review and fixed here**

Both come from the same new coupling — the build now resolves through `deno.json`, so every path that used to degrade gracefully became load-bearing.

- A `deno.json` that Deno accepts but a strict JSON parser does not. `deno.json` is JSONC; comments and trailing commas are valid, and `deno task` reads such a file happily. Setup called it unparseable, skipped it, and still printed its `Next` line and exited 0 — the build then died on its first import. Parsing JSONC is not the answer either, since writing the file back would discard the reader's comments. Setup now leaves it byte-for-byte alone, names the entry to add, and exits non-zero.
- A `deno.json` that delegates via `importMap`. Deno ignores that field the moment `imports` appears beside it, so writing an `imports` block silently disabled the project's entire module resolution — a working project stopped working, reported as success. Setup now keeps the delegation, reads the file `importMap` points at, and asks for the entry only when that file does not already declare it.

Both surface through the setup result's manual steps, printed under an `Incomplete` heading with a non-zero exit. A scaffold that claims success over a build that errors on its first import is worse than one that says plainly what is left to do.

**Tests**

- E2E, in `tests/e2e/setup_lint_e2e.ts`: runs the real `deno lint` and `deno fmt --check` over a setup scaffold, an import scaffold and a re-scaffolded legacy project, plus a real `deno run` against a delegating project before and after setup. Each was verified to fail when its fix is reverted — they are not vacuous. Lint and fmt parse without resolving modules, so the suite stays hermetic.
- Integration: pins the exit code and the untouched file for both incomplete-scaffold shapes.
- Unit: the merge, the state classification, the delegation guard, the remap detection, and both generators.

4706 tests, lines 98.3% and branches 97.4%.

**Not in scope**

Roughly 230 inline `jsr:@zuke` import specifiers remain in `docs/`, `examples/` and the package READMEs — tracked in #623. The two agent skills are updated here, since those are what an agent reads immediately after running setup. Plugin version 0.58.0 to 0.59.0.

## Related issues

Closes #622
Refs #623

## Checklist

- [x] This PR closes a tracking issue (`Closes #622` above).
- [x] The PR title is a
      (`type(scope): summary`).
      changed.
- [x] Public API changes were regenerated with `./zuke apiDocs` — no published
      entrypoint changed, and `apiDocsCheck` confirms no drift.
- [x] No `any`, no `as` casts or `!` non-null assertions in `src/` — narrowed
      with type guards instead.
- [x] No duplicated logic: the caret-major pin has one definition, and the
      `imports` and `tasks` merges share one helper.
      lambda rather than a behaviour-switching flag, narrow parameter types,
      and dependencies on the injectable seam rather than the global runtime.
- [ ] A new package was wired into all six places — not applicable.

https://claude.ai/code/session_01PQ8JRbSe6tqDuH9eUs5M1b


totollygeek:
Both findings verified against the real code and fixed in 1ac6d67.

**`19t0j2x3v3eef`** — confirmed, and worse than a maintainability concern: the delegated-import-map path never stopped asking. Reproduced:

```sh
mkdir d && cd d
printf '{\n  "importMap": "./import_map.json"\n}\n' > deno.json
printf '{\n  "imports": { "@zuke/core": "jsr:@zuke/core@^1" }\n}\n' > import_map.json
zuke setup --yes --name Demo
```

The project has already done exactly what the previous run asked, and setup still printed `Incomplete — … 1 step(s) need you` and exited 1 — because it only ever read `deno.json`. A correctly configured project could never get a clean run.

`setupDenoJson` now reads the file `importMap` points at and stays quiet once it declares every scaffold specifier. The conservative direction is preserved: anything unreadable, missing, or unparseable answers "not declared", so an unverifiable map still gets the step rather than a silent pass. A path that leaves the setup directory — absolute, a URL, or climbing out with `..` — is never opened at all, only reported unverified; there is a test asserting the file is not read.

On the "split can drift" framing specifically: `delegatesImportMap` is one exported predicate called from `mergeDenoJson`, `denoJsonState` and `setupDenoJson`, so the three cannot disagree about *whether* a document delegates. The real defect was narrower — nothing checked what the delegated file already contained — and that is what changed.

**`270hrspze91b`** — confirmed in substance, though not at `mod.ts:286`. Propagation through `commandSetup`/`commandImport` is complete: `ImportResult extends SetupResult`, `runImport` spreads the setup result, its early return carries both fields, and neither type is re-exported from `mod.ts`, so there is no external constructor to break. What *was* partial is the producing side — `setupDenoJson` fed `manualSteps`/`notes` while `setupMcpConfig` skipped an unparseable `.mcp.json` with nothing but a line of progress output. A run could be given `--mcp`, silently not register the server, and still exit 0. That now records a note. A note and not a manual step: the build runs either way, only an agent client will not find it.

Gate on this push: 4706 tests, lines 98.3% / branches 97.4%, plus `format`, `lint`, `spell`, `docLint`, `snippetsCheck`, `examplesCheck`, `apiDocsCheck`, `skillsCheck`, the sync checks and the e2e `integration` lane.

---

PR #645, opened 2026-09-21
feat(core): add `outdated --update` to bump the versions the lock resolves

### LINKED ISSUE #644: feat: `zuke outdated --update` to bump the versions the lock resolves
### Pre-flight checklist

- [x] I searched existing issues and did not find a duplicate.
- [x] This is a feature request, not a bug report or a usage question.

### Problem or motivation

`zuke outdated` answers *which* JSR packages the lock resolves behind their latest. It cannot do anything about them, and today it ends by telling you to go and edit the lock by hand:

> `N packages are behind. Delete these entries from the lock (or the lock file) and re-run to resolve them afresh …`

That instruction is correct, and that is the problem: the tool knows exactly which entries to remove and makes the human do it.

Worse, there is no other command that will. I checked each candidate against a real project rather than assuming — an inline `jsr:@std/encoding@^1.0.0` with the lock pinned at `1.0.5` while `1.0.11` is published:

| Command | Result |
| --- | --- |
| `deno outdated` | prints nothing — it reads manifests, and an inline specifier is in no manifest |
| `deno outdated --latest` | prints nothing, same reason |
| `deno outdated --update --latest` | lock still `1.0.5` |
| `deno install --frozen=false` | lock still `1.0.5` — a satisfying entry is left alone |
| remove the entry, then `deno install` | lock becomes `1.0.11` |

So the last row is the only thing that works, and it is precisely what `outdated` currently asks a person to perform by hand.

This matters most for the case `outdated` was written for. A build's specifiers are written inline — `jsr:@zuke/git@^1` in `zuke.ts` — rather than in an import map. The caret admits every 1.x, so nothing is *wrong*; the lock simply keeps resolving whatever it resolved the day the line was written, `--frozen` is satisfied, and the build sits months behind a wrapper it could be using.

### Proposed solution

`zuke outdated --update [<package>…]`, mirroring `deno outdated --update`, which is the vocabulary people already have.

With no names it updates every package the report lists as behind. With names it updates only those, so a risky dependency can be held back while the rest move.

The mechanic is the one proven above, automated:

1. Run the existing report to find what is behind.
2. Remove those packages' entries from the lock — both the `specifiers` mapping and the corresponding `jsr` integrity records.
3. Re-resolve, so the lock is written back complete and valid rather than left with holes for the next `--frozen` run to trip over.
4. Diff the resolved versions and report what actually moved.

Step 4 is the part worth insisting on, because **a bump can legitimately fail to move**. If the specifier is `jsr:@zuke/git@1.2.0`, or `@^1.2` against a new major, re-resolution returns the same version — the constraint, not the lock, is what pins it. Reporting "updated" there would be a lie, so a package that did not move is named, with the specifier that held it back, and the suggestion to widen the range in the source.

Shape:

```
@zuke/git   1.2.0  →  1.9.0
@zuke/gh    2.0.1  →  2.3.0
@std/yaml   1.0.5     held by `jsr:@std/yaml@1.0.5` — widen the range to move it
```

### Alternatives considered

**Delete the entries and stop, letting the next run re-resolve.** Fewer moving parts, but it leaves the lock incomplete: any `--frozen` run in between fails, and both this repo's launchers and its tasks pass `--frozen`. Writing a valid lock back is the whole job.

**Shell out to `deno outdated --update`.** Does not work, per the table above — it cannot see inline specifiers, which is the case that motivates the command.

**A separate `zuke update` command.** Discoverable on its own line, but it would have to re-derive the same report, and two names for one behaviour expands the surface for no new capability.

**Restrict it to `@zuke/*`.** Tempting, since the ask came from wanting Zuke's own packages current. Rejected because `outdated` reports every JSR package the lock resolves: a command that reports a package as behind and then declines to update it reads as a bug. The optional name arguments cover the narrow case without the asymmetry.

### Additional context

**Scope.** `--update` changes files, so it should refuse to do anything surprising: it touches only the lock, never a source file or an import map. Widening a specifier's range is the author's decision and is reported, not performed.

**Placement.** The rewrite is a different concern from the report, so it belongs in its own module beside `outdated.ts` rather than growing it. Note that `src/lock.ts` is about a target's cross-run lock and is unrelated — the new file should not be named in a way that invites that confusion.

**Testability.** The lock rewrite must be a pure string-to-string function so it is unit-testable without a filesystem or a network, matching how `lockedJsrSpecifiers`, `isBehind` and `formatOutdated` are already written. The re-resolution step is the one piece that shells out, and it should take the Deno executable from `denoExecutable()` so a project whose Deno came from the launcher still works.

**Acceptance criteria.**

- `zuke outdated --update` bumps every behind package whose specifier admits a newer version, and leaves a valid, complete lock.
- `zuke outdated --update <name>…` restricts the change to the named packages; an unknown name is an error naming it, not a silent no-op.
- A package whose constraint holds it back is reported as held, with the specifier responsible — never reported as updated.
- `--update` composes with `--exit-code` in a way that is documented and tested, since "something was behind" and "something is still behind after updating" are different questions.
- No source file or import map is modified.
- Unit tests for the pure lock rewrite, an integration test through the CLI, and the docs and cheatsheet updated in the same change.

totollygeek:
## What & why

`zuke outdated` told you which JSR packages the lock resolves behind their latest, and then ended by asking you to go and edit the lock by hand:

> `N packages are behind. Delete these entries from the lock (or the lock file) and re-run to resolve them afresh …`

That instruction was accurate, which is what made it worth fixing: the command already knew exactly which entries to remove and was making a person do it.

`outdated --update` does it. Names after the command narrow it, so a risky dependency can be held back while the rest move. It touches the lock and nothing else.

It prints one aligned line per package it moved — the name, the version the lock held, an arrow, and the version it holds now — then a count and a reminder to review the lock diff before committing it.

## Nothing else does this

Checked against a real project rather than assumed — an inline `jsr:@std/encoding@^1.0.0` with the lock pinned at `1.0.5` while `1.0.11` was published:

| Command | Result |
| --- | --- |
| `deno outdated` | prints nothing — it reads manifests, and an inline specifier is in no manifest |
| `deno outdated --update --latest` | lock still `1.0.5` |
| `deno install --frozen=false` | lock still `1.0.5` — a satisfying entry is left alone |
| remove the entry, then re-resolve | lock becomes `1.0.11` |

So the last row is the only thing that works, and it is exactly what the report was asking a person to perform.

## What the real run caught that the tests could not

The unit and integration tests inject the re-resolution seam, which is what makes them hermetic — and it also means they cannot see whether the real one is correct. Running the finished command against an actual project found that it was not, twice, and both fixes are the interesting part of this change.

**An entrypoint is required.** The first implementation shelled out to `deno install`, which resolves the *manifest's* dependencies. A build's specifiers are written inline in `zuke.ts`, so the dropped entries were simply never written back. `--entrypoint` is what makes resolution see the build's module graph; `Deno.mainModule` is that graph, and a `file://` URL is accepted as-is, which avoids converting one to a path.

**A registry refresh is required.** With the entrypoint fixed, re-resolution landed on `1.57.0` while `1.58.0` was published: Deno caches the registry's version listing, so it happily picks the newest version it already knew about. One release behind, silently — the exact failure this command exists to prevent. `--reload=jsr:` fixes it, and is the narrowest form that does: scoping further, to `--reload=jsr:@scope/name`, refreshes the module content but not the listing, and still lands short.

Both are recorded where the flags are, because both obvious spellings are wrong and the symptom of either is silence rather than an error.

## Honesty about what it cannot do

A bump can legitimately fail to move. If the specifier is `jsr:@std/yaml@1.0.5`, re-resolution returns the same version however often you ask — the specifier pins it, not the lock. Reporting that as updated would be a lie, so it is reported as *held*, naming the specifier responsible.

Its line names the version it is still on, the latest it is held back from, the specifier doing the holding, and what to do about it — widen the range in your source.

Widening that range edits the author's source, which this does not do on their behalf.

`--exit-code` answers a different question after an update than it does on the report. There it means "something is behind"; here, being behind is the expected state of anything a specifier pins, so failing on it would make the flag unusable in the scheduled pipeline the command is meant for. It fails on what is still actionable — a held package, or one that could not be checked — and a clean update exits 0. Both directions are tested.

## Safety

The lock is restored unchanged if re-resolution fails, and also if it returns without putting every dropped entry back. A lock with holes in it fails every `--frozen` run, which is every launcher and task in a Zuke project, so a failed update must not leave one behind. The second case is not theoretical: it is what the broken first implementation did, and the guard is what turned a silently broken checkout into a clear refusal.

That guard matches entries by package *name*, not by specifier key, because a resolver normalises the range it records — a dropped `jsr:@std/encoding@^1.0.0` comes back as `jsr:@std/encoding@1`. Keyed on the old spelling it would report a perfectly good update as lost and roll it back. Tested.

## Verification

Full gate green: 4798 tests, lines 98.3% and branches 97.3%, with every target passing. Beyond the suite, the finished command was run against a real project with a deliberately stale lock and no fakes anywhere, and moved `@std/encoding` from `1.0.5` to `1.0.11` in the actual file — which is how both resolver defects above were found.

The new guard was confirmed non-vacuous by removing it and watching its test fail. `security` could not run locally, since zizmor needs the GitHub advisories API and gets a 403 unauthenticated in this sandbox, so CI is the first place it runs; nothing here touches workflows.

## Related issues

Closes #644

## Checklist

- [x] This PR closes a tracking issue (`Closes #644` above).
- [x] The PR title is a
      (`type(scope): summary`).
- [x] Docs updated in the same PR: the CLI reference, the cheatsheet, and the
      report's own advice, which pointed at the manual route it replaces.
- [x] No dead code.
- [x] No duplicated logic: the rewrite reuses the lock reader the report
      already uses, rather than parsing the lock a second way.
- [x] SOLID shapes respected: the lock rewrite is pure, and the resolution is
      an injected seam so the command is testable without a network.
- [ ] A new package was wired into all six places — not applicable.

https://claude.ai/code/session_01PQ8JRbSe6tqDuH9eUs5M1b

---
_Generated by [Claude Code](https://claude.ai/code/session_01PQ8JRbSe6tqDuH9eUs5M1b)_


totollygeek:
Three checks are red on `9c7c582` — **Core version floors**, **Scan with Zuke (zuke/security)** and **E2E (ubuntu-latest)** — and none of them is this PR's.

All three die in the launcher's Deno bootstrap, before any build code runs:

```
zuke: Deno not found — installing it now...
curl: (22) The requested URL returned error: 504
zuke: Deno download failed (attempt 1/4); retrying in 3s...
curl: (22) The requested URL returned error: 504
zuke: Deno download failed (attempt 2/4); retrying in 6s...
curl: (22) The requested URL returned error: 504
zuke: Deno download failed (attempt 3/4); retrying in 9s...
curl: (22) The requested URL returned error: 504
zuke: failed to download Deno after 4 attempts.
```

An HTTP 504 from the download endpoint, on all four retries, in the jobs whose `Set up Deno` step is skipped because they bootstrap through `./zuke` instead.

**Why it is not this PR's**, on the evidence rather than by assertion:

- **`Build & verify with Zuke` passed on this exact commit.** That job is the full `ci` gate — format, lint, spell, type-check, the whole suite, coverage, and every sync check. The change is fine; what failed is a download.
- **The re-run reproduced it identically**, same four retries with the same backoff, in the same place.
- **It is partial, and split by platform.** On that re-run, `E2E (macos-latest)` and `E2E (windows-latest)` both went green on this same commit; only the Ubuntu one failed again. A code defect would not be choosy about the runner image.
- **`Tests (macos-latest)` passed**, and master's recent CI runs are all green.
- The diff touches `packages/core/src/outdated_update.ts`, `cli.ts`, their tests and the docs. Nothing in it participates in fetching a Deno tarball.

**No fix to port.** This is an upstream availability failure, not a bug with a patch sitting somewhere. Widening the launcher's retry budget to ride out a CDN outage would be an unrelated change smuggled into this PR, and it would paper over exactly the signal a maintainer wants to see when the endpoint is down.

The one re-run is spent and it reproduced, so I am standing down rather than re-running again into the same 504. I will keep watching and re-run these three once the endpoint recovers; `Core version floors` has not had its re-run yet, because its workflow still has a job in flight and GitHub refuses a partial re-run until that finishes.

---

